Issue #417, opened 2014-01-04
jenkins has many complaints about PHP code


cgolubi1:
Current output of a jenkins run on buttonmen-dev master includes:
- 2198 "checkstyle" warnings
- 160 "PMD" warnings
- 8 "duplicate code" warnings
- 77% PHP code test coverage
- 606 "API documentation" errors (i think these are all duplicates with checkstyle complaints about the same)

Obviously, we can't really enforce coding standards using this tool under these conditions.  Look at these and figure out which ones are real errors (and fix those) and which ones are a coding standard we like but Jenkins doesn't (and fix jenkins).

PR #516, opened 2014-02-01
UI invokes responder.php as /api/responder

### LINKED ISSUE #228: responder.php should be reachable at /api/responder
Since responder.php is the public API interface, it shouldn't be tied to being written in PHP, because that would mean that if we ever wanted to reimplement it some other way, we'd need to get our own UI and also any alternate UIs and people's custom tools to change their locations.  Have it answer to /api/responder, an implementation-neutral name, instead, either by moving the file or by setting up a ScriptAlias in apache.

cgolubi1:
- Fixes #228 (though this turned out to be trivial because apache does all the hard work in its default PHP configuration, and redirects /api/responder to /api/responder.php)
- While i was at it, moved the guts of responder and dummy_responder to their own class files and gave them camelcase class names, taking care of 4/6 of the remaining checkstyle issues
- While i was at _that_, double-checked that PHP unit tests still pass in vagrant and jenkins if i remove those vestigial requires from BMInterface.php and BMDie.php, and that i can still play a game.  With that, we have zero checkstyle warnings, making it easy for us to be alarmed if we see any again in the future.
- So those things partially address #417
- Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/128/

### DIFF
--- a/src/engine/BMDie.php
+++ b/src/engine/BMDie.php
@@ -1,7 +1,5 @@
 <?php
 
-require_once('BMSkill.php');
-
 /*
  * BMDie: the fundamental unit of game mechanics
  *
--- a/src/engine/BMInterface.php
+++ b/src/engine/BMInterface.php
@@ -1,8 +1,5 @@
 <?php
 
-require_once 'BMButton.php';
-require_once 'BMGame.php';
-
 /**
  * BMInterface: interface between GUI and BMGame
  *


blackshadowshade:
So, I've finally got BM up and running again on my tablet. This won't interest anyone else, but I'm recording how I did it here for posterity.

My tablet is running lighttpd, so it needs an explicit rewrite rule in its lighttpd.conf file:

url.rewrite-once = ( "^(.*)/$" => "$1/" )
url.rewrite-if-not-file = ( "^([^?]*)(\?.*)?$" => "$1.php$2" )

The only problem with this setup is that it disables my phpmyadmin link, so I'll need to find some way of disabling the rewrite rule for that.

blackshadowshade:
Okay, getting phpmyadmin not redirected is trivial. The rewrite rules become:

    url.rewrite-once = ( "^(.*)/$" => "$1/" )
    url.rewrite-if-not-file = (
      "phpmyadmin" => "phpmyadmin",
      "^([^?]*)(\?.*)?$" => "$1.php$2"
    )

PR #1222, opened 2014-08-17
Hid link to show all logs when all logs were already shown

### LINKED ISSUE #784: when all log entries are already being shown, suppress option to show more
This is obviously not super important.  But, near the beginning of a game, there may be fewer than 10 each of action log and chat log entries in existence.  It'd be nice to *not* show the user the "View full activity and chat history" option if they're already seeing everything.

AdmiralJota:
Resolves #784;

http://jenkins.buttonweavers.com:8080/job/buttonmen-AdmiralJota/422/

### DIFF
--- a/src/engine/BMInterface.php
+++ b/src/engine/BMInterface.php
@@ -2861,7 +2861,13 @@ class BMInterface {
 
     // Build the various different WHERE, ORDER BY and LIMIT clauses for the
     // different action and chat log SELECT queries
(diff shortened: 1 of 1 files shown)


blackshadowshade:
So, I believe that the last issue mentioned here is resolved in #1571.

That means that I should be able to basically copy the code from here into a new pull request, excepting the updated test files, and have this work.

Issue #779, opened 2014-04-08
Play with a randomly-generated button


jimmosk:
The previous site had an option to create a game where the players had randomly-generated buttons, with different skill-densities ranging from no skills (though perhaps one Swing die?) up to what I think was termed the CRAZY level (most dice having one or more skills).  
Can someone recall what those levels of skill-density were? And more importantly, I'd officially like to request this functionality be added.
[edited title to make clear I didn't mean simply selecting an existing button randomly]

irilyth:
Hey, so, yay! What's next on this ticket, come up with some other flavors of RandomBM? (Given "vanilla", should the others in fact be named after flavors too? Chocolate, Strawberry, CookieDough, Salmon, Sriracha... Ok, perhaps not. :^)

I'd be happy to help with figuring out a full set of flavors; or to just suggest a perhaps-obvious idea for the next flavor: Soldier (four regular dice plus one X swing).

irilyth:
Huh, here is another random idea: We could make the RandomBM buttons based on other buttonsets. (Possibly even part of those buttonsets? Maybe that's not desirable, not sure.) So we could have RandomBM Vampyres, RandomBM Brom, RandomBM Foglio, etc, with the algorithm for creating them designed to give you results very much like the other buttons in those sets.

And then maybe other other things too, but that might be an interesting start. And I actually kind of like the idea of "give me a random button that looks like the other buttons in this set" for a wide range of sets.

Other thoughts?

yawetag:
Didn't the old site have random buttons based on several sets? Or am I
mis-remembering?


Andrew Senger
asenger@gmail.com

irilyth:
> Didn't the old site have random buttons based on several sets? Or am I mis-remembering?

It had different types, and they were sometimes labeled as being related to sets, but only a few. I happened to have written it down once, and I put a list in a comment early in this ticket. (Short summary: Soldiers, Fantasy/Foglio, and Anime. I think things like Brom, Brawl, L5R, Button Brains, Diceland, and Samurai, would also make good additions.)

blackshadowshade:
For information, adding a new RandomBM type is close to trivial---and would be a great place for an up-and-coming back-end dev to jump into adding new button skills!

Pinging @nkmcalli, @ManvilleFog.  ;)

nkmcalli:
Soon - almost there...

jimmosk:
[...]

Issue #1176, opened 2014-08-10
Add flavor text for all buttons


AdmiralJota:
Before #1174, there was apparently no way to put flavor text for buttons into the DB. Now that there is a way (for values of "now" equal to "after that pull request has been merged"), we need to add it for all buttons that need it.

Is there any way we can set things up so that a non-programmer can do this without having to write bunches of SQL?

ericeslinger:
I could write some sql to put the flavor text in- would you want a separate flavor text update file, or for me to change the data.button.sql file?

cgolubi1:
Thanks for offering!

The way we normally do this is you make the identical change in two places:
* in data.button.sql for new site loads (e.g. test sites)
* in an updates file whose name begins with the ticket number, so in this case that would be something like: deploy/database/updates/1176_flavor_text.sql, which is intended to be applied to existing installs (like the dev and prod sites).

The two approaches should result in the same output database format/contents (one of the tests i do on pulls with database changes is to verify this).

ericeslinger:
Yep, I'm working on this. The main goal first is getting my dev environment
set up again, I let it get stale.

ManvilleFog:
My assumption was always that the flavor text for each button would have it's own text file within its own directory in the ``` src/ui ``` directory.  I don't think flavor text should be included within the database itself.  For one thing - keeping it out of the database removes any potential problems from dealing with special characters.  

I would start, perhaps, by creating a ``` src/ui/text/flavor ``` directory, and place within a file called ``` avis.txt ``` containing: 
```
Avis is an expert chainsaw dueler and ice sculptor, and she likes to beat people up.
```

cgolubi1:
I'm strongly in favor of putting flavor text in the database.  It's structured data (i.e. the flavor text is a structured part of the data about the button).  :>)  It can't have anything in it that's harder to encode than the chat text we already encode.  And there's already a field for it in the database.

Let's not reengineer this one from scratch, folks.

blackshadowshade:
So, I've just done a query on those buttons in the database for which we still have no flavour text. I've added the list at the end of this post. I figure that someone almost certainly has flavour text for some of these.

Tagging @yawetag in case you have any more time for this.

Zeppo
Eiko
Wu Lan
Carson (promo)
Gordo
Me am ORK!
MAX(p)
Apples
Green Apple
Abe Caine
Amara Wintersword
Beatnik Turtle
Captain Bingo
Oni
Spite
Super Germ
[...]

PR #1474, opened 2014-12-21
Implemented vanilla RandomBM

blackshadowshade:
Addresses #779 partially, with vanilla recipes (5 dice, no swing, no skills).

Should work fine with Echo and Zero.

Requires database update 00779_RandomBM.sql.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/564/


deploy/database/data.button.sql
```diff
@@ -1084,3 +1087,9 @@ INSERT INTO button (id, name, recipe, btn_special, tourn_legal, set_id) VALUES
 (547, 'Zomulgustar', 't(4) p(5/23)! t(9) t(13) rdD(1) rsz(1) r^(1,1) rBqn(Z)?', 0, 0, (SELECT id FROM buttonset WHERE name="Classic Fanatics")),
 (548, 'Zophiel',         'k(1) (6) z(8,8) H(12) (Y)?',                 0, 0, (SELECT id FROM buttonset WHERE name="Classic Fanatics")),
 (549, 'Zotmeister',      'd(1) d(1) d(2,2) d(8,8) rd(1) rd(26,26)',    0, 0, (SELECT id FROM buttonset WHERE name="Classic Fanatics"));
+
+#############################################
+#####   S P E C I A L   B U T T O N S   #####
+#############################################
+INSERT INTO button (id, name, recipe, btn_special, tourn_legal, set_id) VALUES
+(10001, 'RandomBMVanilla', '', 1, 0, (SELECT id FROM buttonset WHERE name="Special"));
```
cgolubi1: This file should end with a newline.

src/engine/BMBtnSkillEcho.php
```diff
@@ -9,41 +9,56 @@
...
-                // thus choose a default recipe
+        if ('' == $button->recipe) {
+            $newRecipe = $oppbutton->recipe;
+
+            if ('' == $newRecipe &&
+                (
+                    ('Echo' == $oppbutton->name) ||
+                    ('Zero' == $oppbutton->name)
+                )
+               ) {
```
cgolubi1: Why do Echo and Zero need to be hardcoded now, when they didn't used to be?  Can't you look for the presence of BMBtnSkillEcho in the opponent button somehow?

src/engine/BMBtnSkillRandomBMVanilla.php
```diff
@@ -0,0 +1,33 @@
...
+        }
+
+        $button = $args['button'];
+        $dieRecipeArray = array_fill(0, 5, NULL);
+        $validDieSizeArray = parent::$die_sizes_soldiers;
+        foreach ($dieRecipeArray as &$dieRecipe) {
+            $dieRecipe = '(' .
+                         $validDieSizeArray[bm_rand(0, count($validDieSizeArray) - 1)] .
+                         ')';
+        }
```
cgolubi1: IMO, it'd be good to sort the dice from smallest to largest in the recipe.  I think DHS did this, i bet people will ask for it, and it should be quite easy to do here.

src/engine/BMDie.php
```diff
@@ -6,7 +6,7 @@
  */
 
 /**
- * This class contains all the logic to do with dice, including instantiating, activating,
+ * This class contains all the logic to zdo with dice, including instantiating, activating,
```
cgolubi1: Seems like an introduced typo?  Unless i'm missing something...

cgolubi1:
Okay, what all do i need to test here?
* First, i should actually try out the new functionality interactively: Avis vs RandomBM, RandomBM vs RandomBM, Echo vs RandomBM, open games involving RandomBM, continuation game involving RandomBM, finished game behavior, multiround behavior, RandomBM vs auxiliary, RandomBM vs reserve
* Second, we don't have a lot of regression tests around pre-start-of-game functionality, so i'll need to try out a bunch of existing functionality involving open games, random button selection, and Echo and Zero (since that got refactored)
* Third, there should be no effect at all of this code on existing games, so i'll want to do the usual playing of a lot of RandomAI games looking for internal errors, and recording of RandomAI games and playing them back

cgolubi1:
Misc comments from looking at the site:
* I don't like having "Any button", "Random button", "RandomBM Vanilla:" show up in order at the top.  That's too confusing.  My fix would be to make RandomBM the last buttonset, rather than the first buttonset
* If we're going to implement RandomBM as a normal button in a normal buttonset, IMO the set should be named "RandomBM", not "Special".  Again, the purpose here is to make it clear to people who are browsing around what they are looking at.
* I'm pretty sure *button*-level flavor text is implemented, so why not insert RandomBMVanilla into the database with some flavor text describing the random selection?  (e.g. "This button gets a different random recipe in each game, with a vanilla random formula (5 dice, no swing dice, no skills)"o
* I'm reserving judgment about the decision to implement RandomBM as a different recipe for each type of random formula.  It has some benefits and some downsides.  I'm provisionally happy to see it play out and find out what our dev testers think of it.
* In open games, RandomBMVanilla correctly does not yet a recipe until the game starts.  That's as it should be.
* If you do a continuation game with "same" or "reverse" buttons, of course you don't get the same RandomBMVanilla, but a new one.  That seems fine to me --- i just wonder if there's a simple way to make it clear to the player that that will happen.
* All of my "new functionality" tests ("First" step from above) worked just fine.

irilyth:
A few random thoughts about this:

I like the idea of a RandomBM buttonset. Making it the last set sounds fine to me.

One of the things I wished the old site could do is save and reuse RandomBM recipes. I know we're not ready to actually implement that, but design considerations that would make it *possible* to implement that later would, IMHO, be awesome.

One side effect of that is that a "same" or "reverse" game *could* use the same recipe, which I think would actually be better. Or a nice additional option, or whatever. Making it clear meanwhile what will happen now, seems good to me.

irilyth:
Forgot one thing:

YAY RANDOMBM THANKS YOU GUYS :^) :^)

That is all. :^)

deploy/database/data.button.sql
blackshadowshade: We've come across this problem a number of times: that the SQL files aren't checked for conformity with our file requirements. I presume that this is just a matter of updating the path that is checked on Jenkins?

src/engine/BMBtnSkillEcho.php
blackshadowshade: I've taken a look at this, and no, there's no easy way to check for the presence of BMBtnSkillEcho, since Zero has BMBtnSkillZero. Yes, BMBtnSkillZero is implemented as a descendent of BMBtnSkillEcho, but we currently don't actually instantiate each skill when testing whether a button or die has a skill; instead, we simply search an array of strings with the names of each skill.

Of course, I could build something that does search, but I'm not convinced that we need this at the moment. If the issue arises again elsewhere (or even here), I'll build something then.

As for why we need to hardcode, that's a different issue altogether. Previously, we just dealt with the case where a button had an empty recipe, which was always Echo or Zero. Now, of course, we have RandomBMs coming in with empty recipes.

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/587/

I am unaware that descriptions for button specials are supported by the front-end at all. If they actually are, I'd appreciate being pointed in the direction of the front-end code that does this, so that I can work out what the back-end needs to do.

cgolubi1:
> I am unaware that descriptions for button specials are supported by the front-end at all. If they actually are, I'd appreciate being pointed in the direction of the front-end code that does this, so that I can work out what the back-end needs to do.

*Not* button special descriptions.  Button *flavor text*.  If i do:
```
mysql> update button set flavor_text='testing' where name='Eiko';
Query OK, 1 row affected (0.01 sec)
Rows matched: 1  Changed: 1  Warnings: 0
```

and browse to http://localhost:8080/ui/buttons.html?button=Eiko, i see the text i added.

blackshadowshade:
Ah, interesting about the button flavour text. Sorry, I've obviously totally misunderstood what you were asking for, and didn't realise that we had support for button flavour text either.  :)

Sounds like it should be easy enough to do. Any suggestions on how we eventually store the flavour text? By editing each appropriate row in data.button.sql, by adding an additional set of SQL statements, or some other method?

(i.e., do I edit the row for RandomBMVanilla, or add a new statement altogether?)

blackshadowshade:
Linking to #1176.

cgolubi1:
Oh, hmm, i guess the way we're currently doing data.button.sql, the inserts don't have a place for flavor text.  Eventually eventually, i think we should be adding all buttons with flavor text --- DHS had flavor text for all of them.  In the short-term, just add a separate set of UPDATE statements below the INSERTs in data.button.sql, IMO.

blackshadowshade:
Okay, flavour text updated.

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/588/

cgolubi1:
Verified fixes:
* RandomBM is now the last buttonset
* ...and is named "RandomBM"
* ...and RandomBMVanilla has flavor text
* the dice in RandomBMVanilla are now sorted

I did a small number of regression tests, and they turned out okay.  I'm going to go ahead and merge this, and let our dev testers have at it.

cgolubi1:
For the record, i'm very confused about why the newline audit passed on these files --- it *does* check stuff in deploy/database/ --- see http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/569/console for an example.

Or is the newline audit only checking whether the file overall uses DOS line endings, and not actually checking that the file *ends* with a newline.

PR #1479, opened 2014-12-29
Replaced many BMDie properties with BMFlags

### LINKED ISSUE #1266: API should lie about player 2's waitingOnAction value when player 1 is choosing auxiliary
The API goal for choose_auxiliary should be that the API response player 1 sees is identical before and after player 2 chooses to keep the auxiliary die, to avoid leaking information about player 2's choice until player 1 has made their own choice.

It's actually surprisingly close, but we do leak one avoidable tell: once p2 has decided to keep the auxiliary die, p1 sees:
```php
$data['playerDataArray'][1]['waitingOnAction'] = FALSE;
```

In this situation, the API should lie to p1 and always report the same thing about whether the game is waiting for p2's action.

### LINKED ISSUE #1393: clean up unused BMDie properties
The BMDie class contains a number of properties, such as `$inactive` and `$unavailable`, which are not actually used in the code, because they were placed in anticipation of future development, which then took another path.

Remove unused properties from BMDie, and stop suppressing the `PMD.TooManyFields` warning.

blackshadowshade:
Fixes #1393. Fixes #1266 (lies about waitingOnAction during auxiliary selection).

This touches auxiliary, reserve, chance, focus, ornery, and morphing, so this will need quite a bit of testing.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/568/


test/src/api/responderTest.php
```diff
@@ -4559,10 +4559,6 @@ public function test_interface_game_009() {
             $gameId, 'add', 5);
         $_SESSION = $this->mock_test_user_login('responder003');
 
-        // expected changes
-        // #1266 - the API should lie about this to avoid information leaks
-        $expData['playerDataArray'][1]['waitingOnAction'] = FALSE;
```
cgolubi1: While you're changing here, maybe replace the comment with one noting that there should be no expected changes, so later regressions will jump out at us more?

cgolubi1:
If the existing responder tests are any evidence, it looks like this *should* be a noop from the API level, so it's a good candidate for testing by simply throwing a large number of regression-test games at the code (with some hooks to modify the capitalization of Dizzy, Disabled, etc).  So i'll try that, and also a bunch of random novel games to look for internal errors.

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/583/

cgolubi1:
Oh, i literally just meant to add a new comment to the test saying nothing should change here.  But this is okay.

cgolubi1:
Sigh.  So, we're going to have to rethink this lying about waitingOnAction for chooseAuxiliaryDice.  The intention was to lie about whether the game is waiting on action *by your opponent*.  The game can't lie about whether it's waiting on action *by you*, because if it does, the UI will still show you that it's waiting for you to decide whether to use auxiliary dice.  Then you wind up in this situation, where you click the button to add an auxiliary dice and the game gives you an error because you already clicked it:

choose_auxiliary_dice_api_problem

So, that's a bug, and my testing tripped over it.

cgolubi1:
I've told my tester to exclude buttons with auxiliary dice for now, and am running a bunch of games to look for other internal error/broken game state type problems, and then later i'll look for actual regressions.  Sorry i've been slow on testing this.

When you have a fix for the auxiliary bug, i'll bring those buttons back in.

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/597/ (if it passes)

So, I believe that the bug is fixed, but I can't check to see if Jenkins passes, since I only have short bursts of Internet access at the moment.

cgolubi1:
Okay, the auxiliary waitingOnAction business looks right to me now.

cgolubi1:
Ah... so, something i found in regression testing: there's a new responderTest game, 024, which was added during the trip twin die bug business, which isn't in this pull, and which fails when run on this pull:
```php
$ /usr/local/bin/create_buttonmen_databases && sudo -u vagrant php /etc/php5/deploy-includes/phpunit.phar --bootstrap /usr/local/etc/buttonmen_phpunit.php --group fulltest_deps /buttonmen/test/src/api/responderTest.php
buttonmen already exists
buttonmen_test already exists - recreating it
Creating buttonmen_test
Populating buttonmen_test
PHPUnit 4.4.2 by Sebastian Bergmann.

..F

Time: 1.37 seconds, Memory: 14.25Mb

There was 1 failure:

1) responderTest::test_interface_game_024
--- Expected
+++ Actual
@@ @@
                     'properties' => Array (
+                        0 => 'AddAuxiliary'
                     )
                     'recipe' => '+s(X)'
                     'description' => 'Auxiliary Shadow X Swing Die'
                 )
             )
         )
         1 => Array (...)
     )
     'gameActionLog' => Array ()
     'gameChatLog' => Array ()
     'gameChatEditable' => false
 )

/buttonmen/test/src/api/responderTest.php:708
/buttonmen/test/src/api/responderTest.php:8660

FAILURES!
Tests: 3, Assertions: 164, Failures: 1.
```

This should be a super easy fix --- can you go ahead and merge upstream and fix it?

cgolubi1:
Beyond that, i ran 150 random games with no regressions, by applying the following API changes:
 * dizzy => Dizzy wherever it appears in die properies
 * in `CHOOSE_AUXILIARY_DICE`:
     * if the game claims not to be waiting for the opponent's action, change that because the API now lies
     * if the game claims not to be waiting for the player's action, insert the `AddAuxiliary` flag

I'll collect a bunch more regression test games overnight, and run them in the morning, but i'm optimistic.

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/605/

cgolubi1:
Okay, looks good --- regression-tested 1000 games with the API changes noted above, and the only failure i saw was in responderTest itself (it expects games not to go more than 99 moves, and one of my test games got in a trip-vs-shadow loop).

Issue #614, opened 2014-02-27
Statistics for buttons


irilyth:
It'd be cool if there were a way to see statistics for buttons, e.g. win/loss records vs other buttons. Even fancier stuff could be nice too, like a way to see the W/L/T numbers for the games played between two buttons, or even to actually see the games themselves (if they're public or yours), etc.

cgolubi1:
Designating this as my BotM for February.  Next steps are to populate the DHS data on the dev and prod sites, and start working on API methods to access the data.

blackshadowshade:
So, just noting that there is some progress on generating stats from the new site in Issue #2208.

blackshadowshade:
My first proof-of-concept attempt to generate stats (based on my local branch 614_r_stats) produces a HTML page which can be found here: http://stats.dev.buttonweavers.com/ui/stats/button_stats.html

Comments are welcome.

Things I've thought of already:
- adding new columns, e.g., is the button TL, is the button a Fanatic, recipe
- allowing sorting on columns or other dynamic behaviour (which I would consider doing using a Shiny server)
- converting names to links

irilyth:
Is the plan to generate stats into HTML pages, rather than into data in the databse?

blackshadowshade:
I think that a mixture of the two approaches might be appropriate. Some stats pages that are commonly consulted are probably better suited to be static HTML pages, while customised stats pages (e.g., those pertaining to a particular user) may need to be dynamically generated from supporting summary data stored in the database.

We know from the old site just how much server load statistics pages can generate, and don't want to end up copping out and adding the standard disclaimer "please don't do this more than 10 times a day", especially if static HTML pages are enough for most purposes.

dwvanstone:
Thank you! This is wonderful!

Are the buttons (if any) that have played 0 games in this document too?

It would be great to have the Button Set name as another column, especially if it becomes sortable. 

There are some buttons at the top I don't recognize from the production site: Skinny & Gordo.

blackshadowshade:
No, buttons that haven't played any games (apart from mirror battles) aren't included at the moment.

Game 432 has Skinny, and game 416 has Gordo. This must have been before we disabled those buttons where we didn't support the full recipe.

irilyth:
Here are some things I'd like to do with button vs button stats.

[...]

PR #1480, opened 2014-12-29
Adding button stats tables

blackshadowshade:
This is a working pull request made for convenience so that everyone can see where we're up to.

Addresses #614.

Requires database updates 00614_stats_tables_01.sql and 00614_stats_tables_02.sql


blackshadowshade:
Now that a database update script also exists for AMP stacks, I'll need to update that too.

blackshadowshade:
Right, I've updated the script to add stats initialisation now.

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/620/ (assuming it passes)

deploy/database/drop_databases.sql
```diff
@@ -1,2 +0,0 @@
-DROP DATABASE IF EXISTS buttonmen;
-DROP DATABASE IF EXISTS buttonmen_test;
```
cgolubi1: Why does this pull remove the drop_databases.sql script?  Is that intentional?  Do you need to merge from upstream?

cgolubi1:
Other than that, this looks straightforward, and the only test i need to do is that the regular and update SQL files yield the same result as each other.

deploy/database/drop_databases.sql
blackshadowshade: If you look at the function recreate_database() in database_initialise.php, you'll see that the drop actually happens there. Thus, we don't need drop_databases.sql any more for the AMP stacks.

You also don't seem to use it in your initialise_schema script, so it's orphaned, as far as I can tell.

cgolubi1: It's not orphaned --- i never used it, and you introduced it recently for use by AMP, which is why i was surprised that you removed it.  But, anyway, i have no objection.

cgolubi1:
Looks fine.

PR #1486, opened 2015-01-02
Increased max chat message length from 500 characters to 2000 characters

### LINKED ISSUE #367: Chat allows 500 characters
I'm okay with it allowing only 500 characters. However, there should be an indication on how many characters are remaining as someone is typing.

blackshadowshade:
Fixes #367.

Requires database update 00367_chat_length.sql.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/576/

### DIFF
--- a/deploy/database/schema.game.sql
+++ b/deploy/database/schema.game.sql
@@ -85,7 +85,7 @@ CREATE TABLE game_chat_log (
     chat_time          TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
     chatting_player    SMALLINT UNSIGNED NOT NULL,
     -- A chat message is limited to 2000 Unicode characters (see
-    -- GAME_CHAT_MAX_LENGTH in BMInterface and Game.js). Since a Unicode
+    -- GAME_CHAT_MAX_LENGTH in ApiSpec.php and Game.js). Since a Unicode
     -- character can be up to four bytes, this requires a varchar(8000).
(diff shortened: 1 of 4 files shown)


deploy/database/schema.game.sql
```diff
@@ -84,10 +84,10 @@ CREATE TABLE game_chat_log (
     game_id            MEDIUMINT UNSIGNED NOT NULL,
     chat_time          TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
     chatting_player    SMALLINT UNSIGNED NOT NULL,
-    -- A chat message is limited to 500 Unicode characters (see
+    -- A chat message is limited to 2000 Unicode characters (see
     -- GAME_CHAT_MAX_LENGTH in BMInterface and Game.js). Since a Unicode
```
cgolubi1: If GAME_CHAT_MAX_LENGTH was actually moved from BMInterface to ApiSpec, which it seems like based on what you changed in this pull, this comment should be updated to say so.

cgolubi1:
This seems simple.  I'll try it out by updating a database with an open game with some chat in it, and modifying an existing message to be more than 500 chars.  I'd like to see that comment updated per the above, but don't see anything else complicated here.

blackshadowshade:
Comment updated.

cgolubi1:
Looks good.

Issue #205, opened 2013-11-26
Offer the option to reject a game


irilyth:
When another player challenges you to a game, you should be able to reject it. (Ideally with an accompanying message explaining why.)

blackshadowshade:
Just an update on this: I've got a branch now with BMInterface functions and database accessors, and this shouldn't change any current behaviour. Once I've tested a little, I'll issue a pull request for this.

cgolubi1:
This is my best bet for BFotM (whether or not i also get a TDotM pick).  Letting people reject games is both the best way to square the uses of the site between people with different attitudes towards cheese buttons, and the currently-active step in changing the pre-start-of-game logic, which i think is a prereq for things like tournaments and different kinds of open games.

cgolubi1:
After discussion with shadowshade, apparently there's no API work left to do on this, so never mind.

blackshadowshade:
So, I've been thinking about this, and this is what I reckon would be easiest:

1. On the Overview page, a new section right up the top called New Games, containing the following information:
- Game number
- Game description  (this would begin to address issue #1508)
- Your Button
- Opponent's Button
- Opponent
- Some sort of accept/decline selector (should this be two buttons?)

2. Ensure that the Next Game link at the top ignores proposed but unaccepted games.

3. Make sure that there is some sort of acceptable placeholder page if someone manually tries to view a proposed game.

What do you think?

blackshadowshade:
Oh yes, there would also be a new user preference to autoaccept games, default TRUE.

irilyth:
I don't have strong feelings about this, but an alternative: Have them show up in the overview just like games in progress, but on the *game* page, display information about the game including the option to accept or reject it.

One thing I like about that is that you can put all kinds of information there, including the full button recipes, potentially even let people chat about the game before they accept or reject it (which would let you say "who are you and why did you challenge me" and await a reply, or "no thanks I have too many games now" so the other person knows why you rejected it, etc), and so on, with no worries about "but it has to be compact and unobtrusive on the overview page".

blackshadowshade:
Did you miss the word placeholder?  :p 

Again, the idea is to make this change as minimal as possible, to ensure that I can actually get it working.

blackshadowshade:
I have nothing against the placeholder evolving over time, but initially, I think that it should be extremely simple, perhaps something like: Game 20000 hasn't started yet.

cgolubi1:
[...]

PR #1494, opened 2015-01-07
Implemented BMInterface functions required for rejecting games

blackshadowshade:
Addresses #205.

This change should be invisible to players.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/591/


src/engine/BMInterface.php
```diff
@@ -605,6 +613,38 @@ protected function retrieve_button_ids($playerIdArray, $buttonNameArray) {
         return $buttonIdArray;
     }
 
+    public function save_join_game_decision($playerId, $gameId, $decision) {
+        if (!is_bool($decision)) {
+            throw new InvalidArgumentException('decision must be boolean');
+        }
```
cgolubi1: Editorial comment: for the most part, the other public BMInterface() functions take text arguments like 'add' or 'decline'.  It seems weird that this one should take a boolean, and, given the little API inconsistencies we already have with arguments which are passed as booleans (there are a couple of outstanding technical debt items about this), i'm lukewarm on booleans as being a simple interface.  I think we should expect that the API function for whether or not to join a game will pass a string, and that apiResponder.php will pass that string on to BMInterface as a string.

cgolubi1:
Other than that, in terms of testing this:
* It looks like all existing games on dev and prod have `has_player_accepted=1` for both players.  However, this is definitely a pull for which i'll want to populate a system with a number of games in different states of play (open, open/random button, in-progress in a few places, finished), then update it, and basically visually look for regressions in how those games look.
* I might as well throw some random games and random/regression games at the system just in case, but realistically RandomAI only tests one path between start and a playable game, so i think it's unlikely that i'll catch much with those, and am not going to test heavily that way.  If i find anything, it's going to be by trying creating games a bunch of different ways and looking at them.

blackshadowshade:
So, I've changed the input parameter from a boolean to a string, as requested.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/607/

blackshadowshade:
Sorry, added the change to the QUnit test for copyright year so that I can get a clean Jenkins run.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/609/

cgolubi1:
I don't see any problems in manual testing.  Merging.

PR #1496, opened 2015-01-07
Created PHP file to initialise both databases on the local AMP stack

### LINKED ISSUE #924: Simplify database initialisation under an AMP stack
Currently, if a developer is using a WAMP/MAMP stack and wants to recreate the database from the original SQL files, it is necessary to run multiple files in the correct order, one after the other. This is becoming more tedious and painstaking with each extra SQL file that is added.

It should be possible to initialise the database in some other way that is simpler and less prone to error. One suggestion from Chaos is to create a PHP wrapper that basically runs each of the SQL files in order, similar to ./deploy/pagodabox/init_test_db.php.

It may also be useful to create a simple one-button webpage that enables one-click database initialisation.

blackshadowshade:
Fixes #924.

To use the file deploy/amp/database_initialise.php on an AMP stack, it is necessary to add an extra alias to the httpd.conf. I added the following line to the bottom of the file:

Alias /amp "/Applications/MAMP/htdocs/buttonmen/deploy/amp"

Then, I call the file by going to http://localhost:8178/amp/database_initialise.php

Chaos, I'd like your opinion about whether we leave the DROP statements in the schemas in the format they currently have. They are obviously not doing what we want them to do, because of the CONSTRAINTs in the tables.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/592/


blackshadowshade:
Also, Fog, do you see any utility in separating out the initialisation of the test and the play databases?

blackshadowshade:
So, to be clear, I'd like ManvilleFog (or someone else running an AMP stack on Windows) to test this for me, so that I know that it also works for someone else.

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/593/

cgolubi1:
> Chaos, I'd like your opinion about whether we leave the DROP statements in the schemas in the format they currently have. They are obviously not doing what we want them to do, because of the CONSTRAINTs in the tables.

So... i hadn't thought about this in awhile.  What's the use case for the drops?  In practice, whenever i need to drop/recreate a database, i drop/recreate *the database*.  There's no point in worrying about "did we get rid of this table or rename that view?" --- if what you want is to drop everything, you can drop everything as a one-liner by dropping the database.  (And, in vagrant-land, /usr/local/bin/create_buttonmen_databases does the right thing in that case.)

Is there a reason a similar approach wouldn't work for AMP?  If it would work, then let's ditch the approach of dropping tables by hand altogether, and AMP can just implement a way to drop the databases.

blackshadowshade:
I couldn't find an easy way to drop the databases in AMP that didn't resort to doing things manually.

The idea here is to end up with one-click database recreation, of course.

blackshadowshade:
Notes to myself, so that I can try simplifying all the DROP logic:

It's apparently possible to create databases using PDO:
http://stackoverflow.com/questions/2583707/can-i-create-a-database-using-pdo-in-php

It's probably also possible to drop a database using PDO using a standard MySQL command.

If I add these steps to the beginning of the script, then we don't need to DROP everything manually.

blackshadowshade:
I'm taking responsibility for this pull request until I have a satisfactory way of dropping the database.

blackshadowshade:
So, Chaos, assuming that this Jenkins run passes, and that you're happy with the changes, I think that this is ready to be merged, since it works on my local MAMP.

Note that this is *in spite* of the fact that Fog and I can't get it running on our WAMP stacks yet, since our Apache-fu is obviously not high enough. After this pull request is merged, we can put in a bit more of an effort into getting our Apache setup right so that we can actually see and run the PHP file.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/612/

cgolubi1:
Sure, that's fine.  I don't think we need to merge it during the freeze (though it's likely harmless), but i don't see any issue with it post-February release (assuming Jenkins passes).

Issue #1130, opened 2014-07-23
Remove PHPDoc warnings


blackshadowshade:
Currently, our Jenkins runs show huge numbers of PHPDoc warnings. The documentation in the code should match the format needed to stop these warnings arising.

blackshadowshade:
As an quick comment here, @cgolubi1, is it easy to exclude single files or directories from the phpdoc run? If so, I'll be compiling a list as soon as I'm back to having reliable Internet access.

blackshadowshade:
I'm going to pick this as my BotM for March. Chaos, I'm looking for help on how to exclude single files or directories from the PHPDoc run. I'll be aiming to reduce the PHPDoc warnings this month too.

blackshadowshade:
So, the PHPDoc help says that it's possible to add the flag "-i" or "--ignore" to the command-line call of phpdoc:

--ignore (-i)         Comma-separated list of file(s) and directories (relative to the source-code directory) that will be ignored. Wildcards * and ? are supported (multiple values allowed)

That should do nicely. Where do I need to add this in the config files for Jenkins?

blackshadowshade:
This is currently blocked by #1942.

PR #1498, opened 2015-01-11
More work on PHPDoc

blackshadowshade:
Partially addresses #1130.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/595/

### DIFF
--- a/src/engine/BMSkillBerserk.php
+++ b/src/engine/BMSkillBerserk.php
@@ -9,6 +9,12 @@
  * This class contains code specific to the berserk die skill
  */
 class BMSkillBerserk extends BMSkill {
+    /**
+     * An array containing the names of functions run by
+     * BMCanHaveSkill->run_hooks()
+     *
+     * @var array
+     */
     public static $hooked_methods = array('attack_list', 'capture');
 
     public static function attack_list($args) {
@@ -66,6 +72,11 @@ class BMSkillBerserk extends BMSkill {
         $game->activeDieArrayArray = $activeDieArrayArray;
     }
 
+    /**
+     * Description of skill
+     *
+     * @return string
+     */
     protected static function get_description() {
         return 'These dice cannot participate in Skill Attacks; ' .
                'instead they can make a Berserk Attack. These work exactly ' .
@@ -77,6 +88,14 @@ class BMSkillBerserk extends BMSkill {
                'characteristics it may have had.';
     }
 
+    /**
+     * Descriptions of interactions between this skill and other skills
+     *
+     * An array, indexed by other skill name, whose values are descriptions of
+     * interactions between the relevant skills
+     *
+     * @return array
+     */
     protected static function get_interaction_descriptions() {
         return array(
             'Mighty' => 'Dice with both Berserk and Mighty skills will first ' .
@@ -91,6 +110,11 @@ class BMSkillBerserk extends BMSkill {
         );
     }
 
+    /**
+     * Does this skill prevent the determination of whether a player can win?
+     *
+     * @return bool
+     */
     public static function prevents_win_determination() {
         return TRUE;
     }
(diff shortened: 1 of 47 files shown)


src/engine/BMSkillBerserk.php
```diff
@@ -91,6 +125,11 @@ protected static function get_interaction_descriptions() {
         );
     }
 
+    /**
+     * Does this skill prevent the determination of whether a player can win?
```
cgolubi1: So... to me, comments like this one don't seem very useful.  I get that copy/paste is the easiest way to shut up phpdoc, but maybe that means we simply shouldn't run phpdoc?  In particular, this isn't useful because:
* A relevant piece of metadata about this function is that it overrides the base function in BMSkill (that you actually *could* say and copy/paste among all the children, so it doesn't require a lot more work)
* You're commenting a function in which BMSkillBerserk overrides the base `prevents_win_determination()` value, and the code of the function shows that the Berserk skill does prevent win determination, but the new comment doesn't say a word about *why*.

My takeaway is, if we're not going to use comments to explain design decisions, why write comments?  And if we're not going to use the phpdoc check as an excuse to write comments that explain design decisions, i'd rather disable the phpdoc check.

In particular, obviously, `BMSkillBerserk::prevents_win_determination()` is a simple function and it's pretty easy for anyone with a passing familiarity with buttonmen to figure out why it might return true for berserk dice.  But if i were commenting this, i'd say something more like:
```php
/**
 * Override the base function from BMSkill because dice change size when they participate
 * in Berserk attacks, which prevents determination of whether a player can still win the round.
```

Or, y'know, whatever.  Anyway, i can also just blindly approve this pull on the theory that it isn't very important, and i haven't read down very far, so there may be more informative comments below, or it may be that having the `@return` statements is helpful enough that we should keep the phpdoc check to force us to write those and just intentionally not worry about the function description strings.

But, basically, i'd like your two cents on what the purpose of adding these comments is.

cgolubi1:
Restating what i was getting at above for clarity: as far as i can tell from my read so far, very few of these comments improve my understanding of how/why the codebase works.  Is that something i should take into account in deciding whether to approve this pull?  If not, why not?  (This may very easily be the case --- i just want to understand the rationale if so.)

blackshadowshade:
My intent at the moment is to get the PHPDoc warnings down to zero. That will mean that we either have placeholders everywhere where we need a comment, or that we have excluded files that don't need commenting.

Once we're at that stage, we can start actually adding useful comments for code that needs it.

I'm working off a statement that I read recently that said that the easiest way to enforce a standard of any sort (including commenting) is to implement it in existing code, since most new code is based on old code. Since we don't even have documentation placeholders, there's nothing signalling where documentation is necessary for new code.

cgolubi1:
Okay, i buy that.  I'll give this a skim and merge it.

test/src/engine/BMGameTest.php
```diff
@@ -8827,12 +8827,10 @@ public function test_mood_swing_rounds() {
         $this->assertInstanceOf('BMDieSwing', $button1->dieArray[4]);
         $this->assertTrue($button1->dieArray[4]->needsSwingValue);
         $this->assertTrue($button1->dieArray[4]->has_skill('Mood'));
-        $this->assertEquals(array('pre_roll', 'add_skill'),
+        $this->assertEquals(array('pre_roll'),
```
cgolubi1: Huh, why did this test result change?  I thought this was just a doc pull.

blackshadowshade: I also removed the die hook add_skill (and associated functions) from Mood because it wasn't doing anything.

cgolubi1:
Ran some regression tests, and they look fine, so accepting this.

PR #1499, opened 2015-01-16
warrior, time & space fanatics, and redone niles

ManvilleFog:
Jenkins likes this
http://jenkins.buttonweavers.com:8080/job/buttonmen-ManvilleFog/96/console

Before accepting the fix to Niles, please take a look at him compared to other soldiers. Also compare both the current and the fixed image to the source image from James Ernest's cafepress page. 

http://dev.buttonweavers.com/ui/buttons.html?set=Soldiers
http://www.cafepress.com/+button_men_niles,470765850

I'm not sure that Niles should be made lighter.  The current Niles seems to match the tone of the rest of the soldiers.  I don't think I did anything differently with Niles when I updated the other soldiers images - they all have a higher contrast and look darker.


ManvilleFog:
niles
niles
_niles

For quick comparison: newly submitted Niles,  old Niles,  currently merged Niles.  
I think that the new Niles will be too light, but the current Niles' coat is more grey than blue.

cgolubi1:
Hmm... so, it probably won't surprise you to hear that i like the new Niles a lot, presumably hence all those caveats.  :>)

I think the important features of Niles are the purple background, the blue coat, and being brighter than the other Soldiers --- that is, i think it's the right implementation for several of the other Soldiers to be fairly dark, and for Niles to be sort of incongruously wandering around with his flower and his pastels.

Anyway, i'll need to give the new Fanatics a closer look and test them and make sure they load in games and stuff, so you have a little while to convince me i'm wrong, but my immediate reaction is "yay".

ManvilleFog:
There's a little loss of detail around the cuffs, face, and hair, too.  But I'm not too unhappy with the new Niles.

cgolubi1:
There's a mismatch between Caligari's art and recipe --- the second die should be 6-sided, not 12-sided.

cgolubi1:
GoldenOtter's auxiliary +z(8) is not shown on the art.

cgolubi1:
professorbooty's `fp(X)` is mistyped on the art as `fp(6)`

cgolubi1:
Those three are all i found.  Once they're fixed, this can be merged.

ManvilleFog:
Fixed. grend had +z(4) instead  of +z(8). GoldenOtter should be fine, unless the recipe is wrong in the database. 

Thanks for finding these. I thought I fixed all the goofs before I submitted this, but I didn't do well double checking my work. :disappointed:    I'm wondering if this is an old-age thing, because my accuracy was previously something I could take pride in.

cgolubi1:
Thanks.

PR #1500, opened 2015-01-16
Implemented RandomBMFixed and RandomBMMixed

blackshadowshade:
Addresses #779.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/598/

Requires database update 00779_02_RandomBMFixedMixed.sql


blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/600/

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/601/

blackshadowshade:
Ok, this should be ready to go.

deploy/database/data.button.sql
```diff
@@ -1128,3 +1130,11 @@ INSERT INTO button (id, name, recipe, btn_special, tourn_legal, set_id) VALUES
...
 SET flavor_text='This button gets a different random recipe in each game, with a vanilla random formula (5 dice, no swing dice, no skills)'
 WHERE name='RandomBMVanilla';
+
+UPDATE button
+SET flavor_text='This button gets a different random recipe in each game, with a fixed random formula (5 dice, no swing dice, two of them having a single skill chosen independently from c, f, and d)'
+WHERE name='RandomBMFixed';
+
+UPDATE button
+SET flavor_text='This button gets a different random recipe in each game, with a fixed random formula (5 dice, no swing dice, three skills chosen from all existing skills, with each skill dealt out twice randomly and independently over all dice)'
+WHERE name='RandomBMMixed';
```
cgolubi1: Newline, please.  Sorry.

deploy/database/updates/00779_02_RandomBMFixedMixed.sql
```diff
@@ -0,0 +1,11 @@
...
+(10002, 'RandomBMFixed', '', 1, 0, (SELECT id FROM buttonset WHERE name="RandomBM")),
+(10003, 'RandomBMMixed', '', 1, 0, (SELECT id FROM buttonset WHERE name="RandomBM"));
+
+UPDATE button
+SET flavor_text='This button gets a different random recipe in each game, with a fixed random formula (5 dice, no swing dice, two of them having a single skill chosen independently from c, f, and d)'
+WHERE name='RandomBMFixed';
+
+UPDATE button
+SET flavor_text='This button gets a different random recipe in each game, with a fixed random formula (5 dice, no swing dice, three skills chosen from all existing skills, with each skill dealt out twice randomly and independently over all dice)'
+WHERE name='RandomBMMixed';
```
cgolubi1: Here too.

cgolubi1:
This looks good at a glance.  Glad to see you using the random vals thing in testing.  We should probably add a responderTest which uses RandomBM at some point, but i don't think that's a blocker for this pull.  In the meantime, i'll just test this by playing a few games with all three random types, and see how it goes.

blackshadowshade:
Done.

cgolubi1:
Did some hand-testing, and it seems to work as advertised.  One thing i'm not sure about: i played a "mixed" game in which one of the skills i got was reserve.  That had the effect that the button only had 3 regular dice to start with, and thus was at a serious disadvantage.  So i'm torn between (a) maybe reserve shouldn't be one of the "mixed" skills vs. (b) these buttons have totally arbitrary win probabilities anyway, so who cares?

cgolubi1:
Hm... the implementation of auxiliary as a randomly-selected skill also has some pretty weird effects:
* Two of your dice get the Auxiliary skill
* Your opponent gets offered the same two Auxiliary dice
* The UI will only allow you to pick one of them (the second one you're offered) to actually play
* Obviously, you lose the other one when you pick the one you pick.  So you wind up in a situation in which the player who got the Aux skill has 4 dice and 2 skills (some of which may have been lost on the die you had no option to select), and the other player has 6 dice and 3 (or even 4) skills.

The unevenness is one thing, but the fact that the UI won't let you do something (specify the first aux die) which you can *see* you might want to do, and it's possible the API would let you do, makes me think we should remove Auxiliary from consideration as one of the RandomBMMixed skills, and open a separate issue to readd it once we sort out how it should work.  What do you think?

Extremely unscientifically, it looks to me like on the order of 1/4-1/2 of games wind up containing any given skill (since each player gets 3 skills and there are two players).  So any given skill whose behavior is super frustrating, will be seen in what feels to players like "most" of the games.

irilyth:
Would it make sense to just leave out any skills that have die-selection effects?

cgolubi1:
I had a weird thing happen just now (i'm running tests with RandomAI, and hacked it to exclude Auxiliary dice from random games), in which i was in choose_reserve and called specify reserve with die 0, and the API gave me "Invalid reserve choice".  I'm betting this is just a python client issue, though, because when i went to replicate it via the web UI, it worked.  So i bet the python client just encodes 0 as NULL or something unhelpful.

cgolubi1:
Hmm, another thing: when konstant is chosen as a skill, the button gets two konstant dice.  It looks to me like the backend isn't handling it correctly if you try to do a konstant attack using the second konstant die.  Tell me what you think: unless i am missing something, this attack should be legal:
konstant_rejected

I heard through the rumor mill that DHS disallowed buttons with multiple konstant dice because the calculation of possible combinations was too hard.  So that may be something we want to keep in mind.

cgolubi1:
Okay, indeed, it actually *did* allow a legal attack with both konstant dice, though mind you i'm only using die 1 for its negating power in this attack:
konstant_succeeded

cgolubi1:
Also, it's very easy to use RandomBMMixed to find new internal errors, but ridiculously difficult to then actually *find* the internal error, since i have no idea what i'm even looking at.  Here's the state immediately *before* the internal error:
right_before_internal_error

When i try to load this game, i get:
```
[Tue Jan 20 14:04:22 2015] [error] [client 10.0.2.2] PHP Notice:  Undefined property: BMDie::$value in /var/www/engine/BMDie.php on line 739
```

But i don't know what RandomAI did to cause the internal error.  If i had to take a guess, it would be that responder004 made an attack with the Dm(10), but in reality i have no idea.  More later.

blackshadowshade:
I'm happy to disallow auxiliary and random, on the grounds that we really should have 5 starting dice.

blackshadowshade:
I don't even know what behaviour should be exhibited by a Dm(10).  XD

cgolubi1:
Yeah, let's ditch aux and reserve for now and see what happens.  The game state in my game with the internal error is truly weird --- i didn't show you the whole screen, but the other thing is that i only have five lines of action_log (in the screen before the internal error).  That doesn't make any sense --- there should be 10.  Is this RandomAI's fault, or the backend's?  When i look in the DB, i see way more ornery rerolls than makes sense at a glance, so maybe that accounts for it (and thus ornery is somehow implicated in whatever other weirdness i'm seeing).  Okay.  i have to put this down right now and go to work.  More later.

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/603/

cgolubi1:
Another internal error: after the die `kh(6)` makes an attack, `load_game()` throws an internal error:
```
[Wed Jan 21 01:11:01 2015] [error] [client 10.0.2.2] Caught exception in BMInterface::load_game: Invalid die value: 6 is not between 1 and 4 for die kh(4)
```

I'm going to be honest: i think there are a lot of internal errors standing between us and the ability to put two random skills on a die without anything weird happening.  I'm not willing to tackle that without some better tooling (to turn RandomAI games into responder tests more quickly than i currently can).  Given the number of backlogged pulls, is that the best use of time?  What alternatives do we have?  Let's discuss.

cgolubi1:
I tested RandomBMFixed for awhile, and the only issue i found was #1515.  IMO, that's not a blocker for accepting this pull --- it's a bit confusing, but the game doesn't actually get into an illegal state at any point.

cgolubi1:
I've now played a large number of RandomBMMixed games without internal errors, with the following excluded skills list:
```php
$excludedSkillArray = array(
    'Auxiliary', 'Reserve', // game-level skills
    'Berserk', 'Doppelganger', 'Mad', 'Mighty', 'Mood', 'Morphing', 'Radioactive', 'Weak', // recipe-changing skills
    'Fire', // skills that add an extra step to attacks
);
```

(Implicitly, that means the following skills are included:
```
Chance
Focus
Konstant
Maximum
Null
Ornery
Poison
Queer
Shadow
Slow
Speed
Stealth
Stinger
Trip
Value
```

That's my recommendation for a first cut at this.  Then you can try adding more skills in later pulls, and we can debug any issues we encounter.

blackshadowshade:
Okay, that list of excluded skills has been expanded as suggested.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/614/ (assuming it passes)

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/617/ (assuming it passes)

cgolubi1:
Okay, looks good.

PR #1520, opened 2015-02-11
Implemented warrior skill

### LINKED ISSUE #264: Warrior Die skill needed
Warrior: `
These are on the left side of the button as with reserve dice, but are framed in a diamond rather than a circle.
- - - - - - -
Warrior Dice are used for the Shadow Warrior Buttonman from the Sailor Moon set. They are represented with a '`' before a die. Warrior Dice are worth no points, cannot be attacked, and ALWAYS roll the highest number possible. Warrior dice can ONLY attack in a skill attack with NON warrior dice. When this happens the die looses the Warrior skill, and can attack and be captured normally.

 Clarifications: 
•`B (Warrior Berserk): Warrior Berserk dice can not skill attack. This combination is useless. 
•`w (Warrior Slow): Slow is redundant. 
•`g (Warrior Stinger): Warrior Stinger can not be used for the Stinger ability until it looses Warrior. 

 Inconsistencies: •`f (Warrior Focus): Focus dice can NOT be used with dice that are not counted for initative. 
NOTE: ALWAYS roll the highest number possible no longer applies when they lose their warrior ability.

blackshadowshade:
Fixes #264.

This should fully enable the warrior skill. However, it will probably need extra unit testing, for example, responder tests for valid and invalid attacks.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/624/ (assuming it passes)


cgolubi1:
There's a bunch of changes to attack logic here, but this is a case where i feel reasonably comfortable leaving that regression testing to RandomAI.  So i think what i need to do is:
* Look over the code more closely to see if i have any questions or objections
* Run a large number of regression test games, comparing the behavior of the current upstream to the behavior with this logic
* Run a large number of RandomAI tests with Warrior dice, and look for internal errors
* Add a responder test containing Shadow Warriors, with a few types of attack, as shadowshade suggests:
    * Attempt to bring in multiple warrior dice at one time (and get rejected)
    * Attempt to attack a warrior die which hasn't been brought in (and get rejected)
    * Attempt to bring in one warrior die on a skill attack (and succeed)
    * Attempt to attack a warrior die which has been brought in (and succeed)
    * Choose to pass where it would be legal to bring in a new warrior die (and succeed)
* Play a game or two by hand to make sure nothing is too strange about the UI experience

blackshadowshade:
If possible, it would be nice to check out the list of possible attack types to see that that is also being passed correctly to the front end.

cgolubi1:
> If possible, it would be nice to check out the list of possible attack types to see that that is also being passed correctly to the front end.

Any responderTest will do that implicitly --- it's part of the API.  But i could also add a UI test based on the responderTest, if that would add value.  (I'm not sure offhand whether it would.  I'll double-check.)

src/engine/BMGame.php
```diff
@@ -59,46 +59,46 @@
...
+    protected $gameId;                // game ID number in the database
+    protected $playerIdArray;         // array of player IDs
+    protected $nPlayers;              // number of players in the game
+    protected $roundNumber;           // current round number
+    protected $turnNumberInRound;     // current turn number in current round
+    protected $activePlayerIdx;       // index of the active player in playerIdxArray
+    protected $playerWithInitiativeIdx; // index of the player who won initiative
+    protected $buttonArray;           // buttons for all players
+    protected $activeDieArrayArray;   // active dice for all players
+    protected $attack;                // array('attackerPlayerIdx',
```
cgolubi1: Why the change of all these variables from private to protected as part of this pull?  I don't see a problem with the change, but i don't understand what it had to do with Warrior.

src/engine/BMGame.php
```diff
@@ -1979,36 +1979,66 @@ public function valid_attack_types() {
...
+            foreach ($this->activeDieArrayArray[$attackerIdx] as $activeDie) {
+                if ($activeDie->has_skill('Warrior')) {
+                    $doWarriorDiceExist = TRUE;
+                    break;
+                }
+            }
+
+            if ($doWarriorDiceExist) {
+                if ($this->does_valid_attack_exist($attackerIdx, $defenderIdx, 'Skill', TRUE)) {
+                    $validAttackTypeArray[$attackType] = $attackType;
```
cgolubi1: I found a bug and tracked it down to this line of code --- you don't want `$attackType` here, you want `Skill`.  Because you use but don't define `$attackType`, it takes whatever the most recent value was in the iteration above.  In a game with no skills other than Warrior, that may well be Skill.  In a game like this one, the API is telling me that i have a valid Speed attack (and *isn't* telling me i have a valid Skill attack --- it lists my options as Default, Pass, and Speed), when it should obviously be the other way around:

api_erroneous_speed_attack

cgolubi1:
Another comment: as a user-friendliness thing, i don't like that the items appear in the order:
```
Default Attack
Pass
Skill Attack

Surrender
```

I think the order should instead be:
```
Default Attack
Skill Attack
Pass

Surrender
```

The UI could do this, but by design the UI defers to the attack ordering laid out by the backend.  So i think the backend should offer the optional skill attack as an option before it offers Pass.

cgolubi1:
Another UI thing: if you're in a situation in which Pass is a valid option for you, and you don't select any dice, leave "Default Attack" selected, and click "Beat People UP!", it will find a valid attack with no dice, namely you'll pass.  Is that the behavior we want?

cgolubi1:
I ran a reasonable number of regression test and warrior test games, and didn't notice anything other than the above.

src/engine/BMGame.php
blackshadowshade: I have an outstanding note to myself to change all private variables and methods to protected unless there's a good reason to have them as private. I'm doing this as I touch various files, so it's likely that you'll see this sort of change happening as different files are changed.

As for the rationale, I know that we've discussed the pros and cons of using private vs protected as the default level. With all the stuff that I've been doing with inheritance of skills, it's proven to be somewhat frustrating to have private as the default level, especially with the way that PHP doesn't really enforce encapsulation correctly. Thus, I'm changing the default level to protected as I encounter instances.

Of course, since the change doesn't actually change behaviour in the absence of inheritance, it currently looks like the change is purely superficial, but I anticipate gains in the future.

src/engine/BMGame.php
blackshadowshade: You're absolutely right about the bug fix. I'll put that in.

blackshadowshade:
I'm happy to sort Pass always after all other attack types other than Surrender.

Regarding whether the Default attack should allow a default pass, that's something that should be up for discussion. I think it makes sense to allow it, but I could be persuaded otherwise.

irilyth:
I don't like Default Attack including Pass; too much risk of a mis-click messing up a game, IMHO.

cgolubi1:
Hey, question: why are dice rolled (i.e. why is `bm_rand()` called) for warrior dice?  They always start with their max value, so it seems incorrect to roll a die and throw away what you rolled.

irilyth:
Without looking at the code, I'd guess that you roll any die that enters play, even if you know what it's going to roll, e.g. Maximum dice. Do they do the same thing? They seem substantially similar in this way.

blackshadowshade:
If I remember correctly, both Maximum and Warrior implement their "die is always at maximum" as a post-roll hook, i.e., the value is set AFTER the roll to the maximum. So, yes, the roll is currently being discarded.

blackshadowshade:
Right, I've fixed the bug, and forced the Pass attack to be sorted to the end. I've left default attack untouched for the moment, to be tested on dev.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/632/

blackshadowshade:
Okay, try this. 

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/636/ (if it passes)

cgolubi1:
Hmm, no: now you've taken the addition of "Pass" out of the if block, so Pass will always be added even if there are valid attacks.

If you're not around to verify, i'll double-check this and make sure i didn't miss something, but i think this won't work.

blackshadowshade:
Oops! (and yes, Jenkins caught the problem)

Shows that I shouldn't go around fixing things when I only have five minutes to look at them before I head off to church.

We'll see if Jenkins passes this time.  :)

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/637/

blackshadowshade:
And finally, Jenkins passes!

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/638/

cgolubi1:
And... a merge conflict with the skill pull.  Maybe i should have looped back to this one first; sorry!

blackshadowshade:
Not an issue, I'll deal with this right away.

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/640/

cgolubi1:
This looks good.  Did various RandomAI tests just to make sure there were no merge regressions, and i'm pretty sure if the optional attack type thing weren't fixed, it would have blown up within 100 games.  Merging, and i'll have another pull with a responder test for Warrior soon.

PR #1527, opened 2015-02-13
Removed repeated skills and added missing skill to skill ordering

### LINKED ISSUE #1526: BMSkill->skill_order_comparator() has repeated and missing skills
Reported initially in #802.

What has happened here is a bad manual deconflict during a merge in commit de16393b1ca60090072da3426762ede366ff30d6 on 23 October.

Konstant and Morphing thus occur twice in the list.

In addition, I note that Fire is not on the list, which is a little surprising.

I'm not yet certain which positions were intended for konstant and morphing, although the earlier positions in the array are currently used, and pass the unit tests in BMSkillTest.php.

blackshadowshade:
Fixes #1526.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/627/

### DIFF
--- a/src/engine/BMSkill.php
+++ b/src/engine/BMSkill.php
@@ -216,6 +216,7 @@ class BMSkill {
                      'BMSkillShadow',
                      'BMSkillSlow',
                      'BMSkillStinger',
+		     'BMSkillFire',
                      'BMSkillStealth',
                      'BMSkillOrnery',
                      'BMSkillMood',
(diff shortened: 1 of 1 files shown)


cgolubi1:
I'd like to play some regression test games on this, to make sure the reordering is actually a noop for live play purposes.  That should be sufficient.

cgolubi1:
Looks good.

PR #1531, opened 2015-02-18
Added logic to disallow trip attacks against too-large maximum dice

### LINKED ISSUE #1522: Trip Attack against a too-large Maximum die should be invalid.
If I have a t(4), I shouldn't be able to make a Trip Attack against an M(8), because the M(8) will always reroll 8, and I can't roll an 8. If it's an sM(8), the game could get into a non-terminating state.

See http://dev.buttonweavers.com/ui/game.html?game=1446: "irilyth performed Trip attack using [tn(4):2] against [sM(8):8]; Attacker tn(4) rerolled 2 => 2; Defender sM(8) rerolled 8 => 8, was not captured"

(The Null skill on my t(4) is irrelevant, I believe.)

blackshadowshade:
Fixes #1522.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/633/

### DIFF
--- a/src/engine/BMAttackTrip.php
+++ b/src/engine/BMAttackTrip.php
@@ -45,6 +45,11 @@ class BMAttackTrip extends BMAttack {
             return FALSE;
         }
 
+        if ($this->is_disabled_by_maximum($attackers, $defenders)) {
+            // validation message set within $this->is_disabled_by_maximum()
+            return FALSE;
+        }
+
         $attacker = $attackers[0];
         $defender = $defenders[0];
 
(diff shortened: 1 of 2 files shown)


cgolubi1:
Seems reasonable.  I'll have RandomAI play a bunch of trip games to look for unexpected regressions, and then make a responder test case for this.  We could use a responder test case with RandomBMMixed anyway.

cgolubi1:
This looks good.  Merging it, and i'll have a separate pull with the new responder test shortly

PR #1532, opened 2015-02-20
Sped up checking whether skill attacks are valid

### LINKED ISSUE #1404: looking up skill attacks can be unreasonably slow
During radioactive testing, i came across the following pathological case --- this game takes 5 seconds to load (on either firefox or safari, without firebug enabled):
slow_skill_lookup

It turns out that the full five seconds happens in one function, `BMAttackSkill::find_attack()`.  That's big enough that it might be worth an efficiency refactor of that function.

blackshadowshade:
Fixes #1404.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/635/

### DIFF
--- a/src/engine/BMAttackSkill.php
+++ b/src/engine/BMAttackSkill.php
@@ -56,17 +56,29 @@ class BMAttackSkill extends BMAttack {
 
     protected function sort_distance(array &$values, $targetValue) {
         // centre values around $targetValue
-        array_walk($values, function (&$val, $key, $centre) {$val = $val - $centre;}, $targetValue);
+        array_walk(
+            $values,
+            function (&$val, $key, $centre) {
+                $val = $val - $centre;
+            },
+            $targetValue
+        );
 
         // sort according to absolute distance from centre
         usort($values, array($this, 'absolute_distance_cmp'));
 
         // remove centration
-        array_walk($values, function (&$val, $key, $centre) {$val = $val + $centre;}, $targetValue);
+        array_walk(
+            $values,
+            function (&$val, $key, $centre) {
+                $val = $val + $centre;
+            },
+            $targetValue
+        );
     }
 
-    protected function absolute_distance_cmp($a, $b) {
-        return abs($a) > abs($b);
+    protected function absolute_distance_cmp($aVal, $bVal) {
+        return abs($aVal) > abs($bVal);
     }
 
     public function validate_attack($game, array $attackers, array $defenders, $helpValue = NULL) {
--- a/src/engine/BMAttackSkill.php
+++ b/src/engine/BMAttackSkill.php
@@ -20,10 +20,10 @@ class BMAttackSkill extends BMAttack {
     // Fire still makes life more complex than it might be.
     //
     // "Premature optimization is the root of all evil." -- Knuth
-    protected $hit_table = NULL;
+    protected $hitTable = NULL;
 
     protected function generate_hit_table() {
-        $this->hit_table = new BMUtilityHitTable($this->validDice);
+        $this->hitTable = new BMUtilityHitTable($this->validDice);
     }
 
     public function find_attack($game) {
@@ -34,41 +34,39 @@ class BMAttackSkill extends BMAttack {
         }
 
         $this->generate_hit_table();
-        $hits = $this->hit_table->list_hits();
-        sort($hits);
+        $hits = $this->hitTable->list_hits();
 
-        // Should perhaps start around the defending die's value and
-        // work outward, but probably not worth the extra overhead to
-        // do so. We half-ass it by starting in the middle.
+        foreach ($targets as $t) {
+            // Sort the multidimensional array by distance from target
+            $this->sort_distance($hits, $t->value);
 
-        // PHP, why do you have no integer division operator?
-        $idx1 = (int)(count($hits) / 2);
-        $idx2 = $idx1 - 1;
-        $idx2 += count($hits) % 2; // Start in the exact middle of an odd count
-        while ($idx2 >= 0) {
-            foreach ($targets as $t) {
+            foreach ($hits as $hit) {
                 $def = array($t);
-                $combos = $this->hit_table->find_hit($hits[$idx1]);
-                foreach ($combos as $att) {
-                    if ($this->validate_attack($game, $att, $def)) {
-                        return TRUE;
-                    }
-                }
-                if ($idx1 == $idx2) {
-                    continue;
-                }
-                $combos = $this->hit_table->find_hit($hits[$idx2]);
+                $combos = $this->hitTable->find_hit($hit);
                 foreach ($combos as $att) {
                     if ($this->validate_attack($game, $att, $def)) {
                         return TRUE;
                     }
                 }
             }
-            $idx2--;
-            $idx1++;
         }
 
         return FALSE;
+    }
+
+    protected function sort_distance(array &$values, $targetValue) {
+        // centre values around $targetValue
+        array_walk($values, function (&$val, $key, $centre) {$val = $val - $centre;}, $targetValue);
+
+        // sort according to absolute distance from centre
+        usort($values, array($this, 'absolute_distance_cmp'));
+
+        // remove centration
+        array_walk($values, function (&$val, $key, $centre) {$val = $val + $centre;}, $targetValue);
+    }
+
+    protected function absolute_distance_cmp($a, $b) {
+        return abs($a) > abs($b);
     }
 
     public function validate_attack($game, array $attackers, array $defenders, $helpValue = NULL) {
@@ -118,7 +116,7 @@ class BMAttackSkill extends BMAttack {
     }
 
     protected function is_direct_attack_valid($attackers, $dval) {
-        $combos = $this->hit_table->find_hit($dval);
+        $combos = $this->hitTable->find_hit($dval);
         if ($combos) {
             foreach ($combos as $c) {
                 if (count($c) == count($attackers) &&
@@ -148,7 +146,7 @@ class BMAttackSkill extends BMAttack {
             // james: This logic assumes that firing effectively reduces the defence value.
             //        This assumption fails in the case that part of the skill sum comes
             //        from a konstant die that is being subtracted from the total.
-            $combos = $this->hit_table->find_hit($dval - $i);
+            $combos = $this->hitTable->find_hit($dval - $i);
             if ($combos) {
                 foreach ($combos as $c) {
                     if (count($c) == count($attackers) &&
--- a/test/src/engine/TestDummyBMAttSkillTesting.php
+++ b/test/src/engine/TestDummyBMAttSkillTesting.php
@@ -2,11 +2,11 @@
 
 class TestDummyBMAttSkillTesting extends BMAttackSkill {
     public function reset() {
-        $this->hit_table = NULL;
+        $this->hitTable = NULL;
         $this->validDice = array();
     }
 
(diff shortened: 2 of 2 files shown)


blackshadowshade:
In this fix, I basically implemented the suggestion in Julian's code comment:

> // Should perhaps start around the defending die's value and
> // work outward, but probably not worth the extra overhead to
> // do so. We half-ass it by starting in the middle.

Thus, we now start checking skill attack possibilities from each defender's value, and then work outwards from that.

cgolubi1:
Will this really help all that much?  I'm dubious.

I'll be honest: i don't really understand any of three things:
1. how the skill-checking code works right now
2. why the fix mentioned in Julian's code comment will help all that much (seems like an optimization which comes into play only when the defender's value is far from "the middle", whatever that means, and when some valid attack exists)
3. how the code change in this pull is the same thing as the fix mentioned in Julian's code comment

All else equal, i'd prefer not to regression-test this blind, to which end i created #1534.  (But i want that task to get done even if i *do* end up regression-testing and approving this blind, because this won't be the last time i ever have to look at the skill attack code.)

Anyway, assuming i do go ahead and test this without understanding the things above, what i'll do is:
* Regression test a *lot* of RandomAI games (more like 1000 than like 100) to make sure the new algorithm gives the same results
* Try to time that big corpus of games against the old and new code, to see if there is any appreciable difference in speed
* Play some games involving Aysez with big swing dice (what's caused the actual live sites to time out at 30 seconds of PHP execution time), and see if there's any timing difference there

blackshadowshade:
I would suggest that you play the games involving Aysez manually first to convince yourself that there is a difference. I'm seeing game load times come down from over 10 seconds to under a second.

As for #1534, I'm happy to do that. Basically, what we're doing is taking attacker die combinations that can possibly hit a certain target number, and then testing to see if the combined attacker/defender die combination is a valid attack. The obvious target number to check first is the defender value, but this wasn't what was happening in the old code---instead, it was starting at the middle of the hit table, which is in the middle of the range of possible target numbers. For Aysez, if the defender had a value that was high or low, it was a LONG way away from the middle of the range, since the range is so huge.

cgolubi1:
I think the phrase "check first" is what has me confused.  If you check attacker combinations that can hit the defender's die value first, what do you check second?  Put another way, why would you ever check any target value *other* than the defender's die value?

blackshadowshade:
The obvious answer here is when the defender's die value is not part of the hit table. This happens when there is no direct combination that adds up to the correct value. Here, we still check for validity because of the possibility of help that Fire dice gives us.

blackshadowshade:
BTW, the current method of searching is not necessarily sensible when dealing with stinger dice---here, an exhaustive n vs 1 search without a hit table may well be quicker, since the current method is likely to analyse the same combination of attackers multiple times when there is no valid skill attack. However, the chance of having no valid skill attack is low when there are many dice on both sides, which is the test case when I experience very long loading times.

cgolubi1:
Well.  The Aysez games i tried seemed pretty zippy, and i ran 1000 RandomAI games with no regressions, and another 150 or so against the new codebase directly.  So let's try this.

jl8e:
Some explanation of why hit tables, and how the skill code should work and is working:

The degenerate case of skill attacks is where you have a lot of potential attacks, and very few (or none) of them hit. Golo vs Nelde (Y=1), perhaps. We’re going to have to check every combination of Golo’s dice, for every value of each die, against each of Nelde’s dice. Worst case (10, 12, 20, 20), that’s 62 1-die checks, 120 + 200 + 200 + 240 + 240 + 400 2-die checks, 2400 + 2400 + 4000 + 4800 3-die, and 48000 4-die. (Times five, since nelde has five dice.)

This can be made a lot faster if the skill attack code understands how the various skills behave, but that’s a nightmare of special cases, and should be avoided unless there’s no other choice.

The idea of hit tables is to make the bad cases faster by making the general cases a bit slower. We build a complete list of the values that the attacker’s dice can make, and each value has an attached list of which combinations of dice can make it. At the very least, that cuts the Golo vs. nelde case by a factor of five, and IIRC, there are some other optimizations in there to reduce it further. (And it’s almost certainly tractable to future optimization as well.)

The way the skill attack code should work is:

- build the hit table
- check each of the defender’s die values by looking it up in the table. If it’s there, and the attack validates, we’re good
- if we have no direct hits, check to see if there’s any assist values available (Fire, or any future skill that works similarly)
- if there is assistance, search the entire hit table, using validate_attack to see if any of the possible attacks are possible due to assistance

That last step is slow, and only necessary rarely.

The problem with the current code is twofold:

- validate_attack is rebuilding the hit table every time it is called. This is wrong, and insanely expensive.
- Because I forgot to call validate_attack in step two, it was creating bad results when stealth got implemented, and so the middle two steps got cut out, leaving only the Fire code to do the checking. It works, but it’s doing a brute-focre search in all situations, when the search should be needed only rarely. (Especially since there are very few buttons with both Stinger and Fire.)

The Aysez problem was particularly bad because Aysez’s big dice and single stinger meant that the hit table was full of results much higher than the opponent’s dice, so starting at the middle of the hit table required a lot of searching before numbers low enough were found.

This fix avoids that particular problem by sorting the hit table’s result list by distance from the value of die we’re checking, which means that, when there is a hit, we get it immediately. If there isn’t a hit, we’re still going to have to search the entire hit table, calling validate_attack (and thus rebuilding the hit table) for every entry. It’s treating a symptom, not the underlying problem.

cgolubi1:
On Mon, 23 Feb, 2015 at 11:54:16 -0800, Julian wrote:

> The idea of hit tables is to make the bad cases faster by making the general cases a bit slower. We build a complete list of the values that the attacker’s dice can make, and each value has an attached list of which combinations of dice can make it. At the very least, that cuts the Golo vs. nelde case by a factor of five, and IIRC, there are some other optimizations in there to reduce it further. (And it’s almost certainly tractable to future optimization as well.)

So, here's what i don't get: *why* make a hit table which is "a complete
list of all the values the attacker's dice can make... except for the
values the attacker's dice can make which the hit table doesn't cover"?

That is, if your hit table generation covers the idea that a konstant
die can contribute either N or -N, and your hit table generation covers
the idea that a stinger die can contribute any of 1, 2, 3, ... N, why
doesn't hit table generation cover the idea that a fire die which isn't
part of the set can contribute 1 ... N-1?

I'm asking because, if hit tables covered all scenarios, then you
wouldn't need to do this futzing around searching for defenders, right?
You'd just make the hit tables, then look up the defender's actual value,
find either an entry or no entry, and be done.

jl8e:
Because Fire can’t always contribute its full value, and how much it can contribute depends on the individual dice attacking.

cgolubi1:
On Mon, 23 Feb, 2015 at 12:59:22 -0800, Julian wrote:

> Because Fire can’t always contribute its full value, and how much it can contribute depends on the individual dice attacking.

That's not really an answer.  Fire dice obey the same rules at hit table
construction time as they do later.  Why can't you take those rules into
account when constructing the hit table?

jl8e:
Possibly I could have, but there are complications:

- Instead of querying the dice “What values can you contribute to this attack”, the question becomes “what values can you contribute to this attack with these other attackers” 
- Since fire dice are not considered participating in the attack, validation from the hit table becomes more complex. ATM, the table stores that a number can be made with various lists of dice, and those lists can just be handed straight to the validation function. There’s no provision for marking dice in a list as being used, but not actually attacking, which matters when the validation function has to ask a Stealth die whether this is a legit attack. (Also may matter when you’re firing with more than one other die.)

Neither of these are insurmountable, and a lot of the reason is that that’s the way I decided to do it, and part of that decision may well have been that I was still fuzzy on how Fire would work, because it’s weird, while stinger and konstant are simple.

I’m also unsure that, even if it’s doable, it’s worth doing, because most of the speedup gained can be gained with a lot less work, just by looking up the actual values.

And I’m starting to suspect that the result of doing it would be slower than the original design ought to be.

ATM, hit tables are constructed as follows:

- For each die, get a list of its values
- add each value to the table. If the value already exists, add the die to the list of ways the value can be constructed
- for each value that was in the table, add each of the die’s values to that value, and make a new table entry for the new value (or add the new combo to a preexisting entry)

The problem with just including Fire contributions in this is that the amount a Fire die can contribute can change as you add more dice into the combo, thus adding more headroom. It’d have to build the table from non-fire values, then go back and add new entries based on fire contributions to each listed combo.

And at that point, we’re effectively doing the same thing we’re doing now: building a hit table, then going through all its entries to handle Fire contributions. But, because the table has to be precomputed in its entirety, we can’t stop as soon as we have the first hit like we do now.

(It’s entirely possible I’m missing an approach that would work; this stuff is hard.)

blackshadowshade:
I agree with Julian here regarding not including Fire attacks in the hit table code.

I've been thinking about passing validate_attack() an additional parameter, basically a boolean saying whether it can trust the current hit table, and thus doesn't have to generate its own. This would be set to TRUE only when validate_attack() is being called from find_attack(), and then find_attack() would take responsibility for building the hit table for each combination of attackers. Would that address your concerns?

jl8e:
I’m pretty sure the only time validate_attack can’t trust the hit table is when it’s being called from elsewhere; the hit table can’t change.

Unless the skill object persists between API calls, an extra parameter isn’t necessary; just check if the table exists, and generate it if it doesn’t.

blackshadowshade:
I tried exactly that, but I got failures throughout our entire test battery. Somehow, the hit table is persisting even when it is invalid.

blackshadowshade:
For information, the load time for the beginning of round 1, Aysez (U=30) vs nelde (Y=1) takes 6.5 secs on the dev site from reload to full rendering, vs 3 secs for the beginning of round 1 for GorgorBey vs dexx.

blackshadowshade:
In other news, http://dev.buttonweavers.com/ui/game.html?game=1535 took around 20 seconds to load, so Julian is right that there will still be bad cases and that the root problem is not yet fixed.

jl8e:
>I tried exactly that, but I got failures throughout our entire test battery. Somehow, the hit table is persisting even when it is invalid.

That’s because of the use of the singleton pattern for attacks. It was a bad idea (as singleton often is), and I should’ve seen it at the time. Because it’s a singleton, the skill attack object persists, including its internal state, instead of a new one being allocated. We don’t need (or, indeed, want) an attack’s internal state to persist. In normal usage, this isn’t a problem, but in the tests you’re trying to run multiple attacks without the interpreter state being destroyed and set up again in between, and the persistent state screws things up.

Fortunately, de-singletoning attacks looks to be pretty easy, as there’s very little code that actually gets attack objects:

Raven:engine jl8e$ grep get_instance *.php |more
BMAttack.php:    public static function get_instance($type = NULL) {
BMAttack.php:                return $cname::get_instance();
BMAttackDefault.php:            $attack = BMAttack::get_instance($attackType);
BMGame.php:        $attack = BMAttack::get_instance($this->attack['attackType']);
BMGame.php:            $attack = BMAttack::get_instance($this->attack['attackType']);
BMGame.php:            $attack = BMAttack::get_instance($attackType);
BMInterface.php:            $attack = BMAttack::get_instance($attackType);

blackshadowshade:
Yeah, I remember the lengthy discussion around Singleton at the time.

I'll give this conversion a go.

jl8e:
I think it’s just fixing get_instance to return a new object each time. Having briefly inspected the code, I don’t think we’re ever using the singleton pattern.

And then it probably breaks a bunch of tests that are there to make sure the singleton pattern is working.

PR #1551, opened 2015-02-27
Warrior dice now always contribute their maximum value

### LINKED ISSUE #1541: Stinger warrior dice must contribute full value in a skill attack
Apparently, stinger warrior dice must contribute their full value in a skill attack, not only a partial value.

This needs to be fixed, both in the implementation and the description (as well as in the interaction description).

blackshadowshade:
Fixes #1541.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/648/


blackshadowshade:
Wait up on this, it isn't working quite right.

blackshadowshade:
Okay, this should be good now.

Note that I don't think this needs to be approved and pushed this pull freeze, since there are no stinger/konstant dice that also have warrior that are playable.

Ideally, we should have a responder test with a stinger warrior die to show that it can't contribute a partial amount, and a konstant warrior die to show that it can't contribute a negative value.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/649/ (if it passes)

cgolubi1:
Given the reordering, i want to play some regression-test games with Warrior, and some without.  Ideally, the with-warrior games will include test behavior which changes as a result of this fix (and that'll be a good hint as to where i can easily create a responder test).

cgolubi1:
Okay, status:
* played 500 regression test games using randomly-selected buttons, and those were all fine
* played 100 regression test games using randomly-selected buttons containing Warrior dice, and those were all fine
* comparing 35 regression test games using a hacked RandomBMMixed which allows Warrior as a skill.  I found 8 games with changed behavior among those, and i'm inspecting them.

In the process, i found a bug.  I wasn't expecting it, so it took me awhile to figure out.  The bug is: Warrior Focus dice can be turned down during Focus turndown, which is obviously not legit.

Here's the regression-test game which clued me into the bug, included mostly because it's amusing:

warrior_focus_bug

The regression is that, during the game data load shown in this screencap, `validAttackTypeArray` previously contained `[ "Power" ]`.  However, with 1541_stinger_warrior, it contains `[ "Power", "Skill" ]`.

What's the skill attack it finds?  Some digging led me to:
```
[Sat Mar 21 19:13:20 2015] [error] [client 10.0.2.2] In BMAttackSkill::find_attack(1), referer: http://localhost:8080/ui/game.html?game=1
[Sat Mar 21 19:13:20 2015] [error] [client 10.0.2.2] Found an unassisted attack hitting 16, referer: http://localhost:8080/ui/game.html?game=1
[Sat Mar 21 19:13:20 2015] [error] [client 10.0.2.2] Attack uses attacker: (8), referer: http://localhost:8080/ui/game.html?game=1
[Sat Mar 21 19:13:20 2015] [error] [client 10.0.2.2] Attack uses attacker: f`(12), referer: http://localhost:8080/ui/game.html?game=1
```

In other words, the new code says that a warrior die always uses its max value in an attack.  The Focus Warrior 12 showing 1 has a max value of 12, so it contributes 12 to a skill attack, and 4 + 12 = 16.  So the code thinks there's a valid skill attack.

But of course there is not --- the Focus Warrior 12 has a value of 1.

Summary in next comment.

cgolubi1:
To summarize:
* It's a bug that focus warrior dice can be turned down.  I'll file that as a separate issue.
* I don't like this implementation of the fix for #1541, because *if* there's ever a bug or other intentional behavior such that for any reason a Warrior die is showing something other than its maximum value, this implementation means that `find_attack()` will return something nonsensical, as it did in this case.  If you're going to pass along an item in addition to `attack_values`, why not pass along the actual showing value of the die?  (Maybe that's not even necessary because maybe the actual showing value can be inferred from `attack_values`, but, if it can't, why not pass it along?  Why bother getting maxValue involved?)

cgolubi1:
All of the rest of the regressions were exactly what you'd hope for --- games where Warrior Stinger or Warrior Konstant dice had previously been able to make attacks using negative or partial values, and now can't.  So mostly this fix seems fine, but i want to hear shadowshade's input on the issue above.

blackshadowshade:
Sorry about the late comment, I've been a little ill.

I'm happy to change the call from $die->maxValue to $die->value, and I'll address the focus warrior issue separately.

blackshadowshade:
So, that change has been made.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/657/ (if it passes)

cgolubi1:
Yep, in the test game i had, Skill is no longer an attack option.

cgolubi1:
I think this is all set now.

cgolubi1:
I think this is all set now.

PR #1566, opened 2015-03-18
Corrected description of berserk

### LINKED ISSUE #1563: Berserk Attack removes Mood Swing?
The description of Berserk says

> Once a Berserk Die performs a Berserk Attack, it is replaced with a non-berserk die with half the number of sides it previously had, rounding up. It also loses any Swing/Mood Swing/Mad Swing characteristics it may have had.

In http://www.buttonweavers.com/ui/game.html?game=4726, I made a Berserk Attack, and this didn't happen:

> irilyth performed Berserk attack using [B(X=20)?:11] against [(6):3,(Y=10)?:8]; Defender (6) was captured; Defender (Y=10)? was captured; Attacker B(X=20)? changed size from 20 to 10 sides, recipe changed from B(X=20)? to (X=10)?, rerolled 11 => 9

And then later, when I attacked with the same die, it did in fact change size.

Should a Berserk Attack remove Mood Swing? I vaguely recall discussing it at one point, so maybe we concluded that it shouldn't; but if so, we should fix the help text. (And if we did not conclude that, then we should fix the bug. :^)

blackshadowshade:
Fixes #1563.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/653/ (if it passes)

### DIFF
--- a/src/engine/BMSkillBerserk.php
+++ b/src/engine/BMSkillBerserk.php
@@ -99,8 +99,7 @@ class BMSkillBerserk extends BMSkill {
                'of dice which add up exactly to its value. Once a Berserk ' .
                'Die performs a Berserk Attack, it is replaced with a ' .
                'non-berserk die with half the number of sides it previously ' .
-               'had, rounding up. It also loses any Swing/Mood Swing/Mad Swing ' .
-               'characteristics it may have had.';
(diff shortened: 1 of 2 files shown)


cgolubi1:
Looks fine.

Issue #1440, opened 2014-11-28
Unsuccessful trip attack targeting radioactive die is logged incorrectly


blackshadowshade:
In #1432, there is a screen capture showing that the action logging is incorrect for an unsuccessful trip attack targeting a radioactive die.

blackshadowshade:
Chaos, when you have the time, it would be useful for me to have a responder test that addresses this issue and #1452 before I attempt to fix the underlying problem. It's quite possible that all that is necessary is to modify an existing test where trip attacks occur.

cgolubi1:
Oh, sure, i'd entirely lost track of this issue.  I'll take a look.

cgolubi1:
As demonstrated by https://github.com/buttonmen-dev/buttonmen/pull/1567#issuecomment-84604254, the fix in that pull does in fact address this issue too.

blackshadowshade:
Good, closing this issue then.

Issue #1432, opened 2014-11-24
Should a *failed* Trip attack on a Radioactive die cause the Trip die to decay?


jimmosk:
Sorry to add another picky point to the recent rewrite of the Radioactive skill's description, but as it currently reads, 'If a radioactive die is either the attacking die or the target die in an attack with a single attacking die and a single target die, the attacking die splits, or "decays", into two as-close-to-equal-sized-as-possible dice that add up to its original size.' That means that a Trip die targeting a Radioactive die with a Trip attack should decay regardless of whether it succeeds or not.
Is that what actually happens? Should it?
Or should that text be changed from "in an attack" to "in a successful attack"?

And of course, if we eventually have dice that are both Trip and Radioactive, should they decay when they make an unsuccessful Trip attack?  [Edited to add: Well if I'd've looked more carefully... we _do_ have one such die, in NoopMan. See below...]

PR #1567, opened 2015-03-18
Fixed logging of attacking trip dice that split

### LINKED ISSUE #1452: Logging for konstant trip dice performing trip attacks is incorrect
Konstant trip dice performing attacks against radioactive dice behave correctly, but the splitting doesn't log correctly.

capture

blackshadowshade:
Fixes #1452.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/654/

### DIFF
--- a/src/engine/BMDieTwin.php
+++ b/src/engine/BMDieTwin.php
@@ -169,7 +169,7 @@ class BMDieTwin extends BMDie {
     }
 
     public function split() {
-        $oldRecipe = $this->get_recipe();
+        $oldRecipe = $this->get_recipe(TRUE);
         unset($this->value);
         $newdie = clone $this;
 
--- a/src/engine/BMGameAction.php
+++ b/src/engine/BMGameAction.php
@@ -173,9 +173,13 @@ class BMGameAction {
             $message .= $this->messageAttacker($preAttackDice, $midAttackDice);
             $message .= '; ' . $messageDefender;
 
-            // now deal with morphing after trip
-            if (isset($postAttackDice['attacker'][0]['hasJustMorphed']) &&
-                ($postAttackDice['attacker'][0]['hasJustMorphed'])) {
+            $splittingAfterTrip = (count($midAttackDice['attacker']) !=
+                                   count($postAttackDice['attacker']));
+            $morphingAfterTrip = (isset($postAttackDice['attacker'][0]['hasJustMorphed']) &&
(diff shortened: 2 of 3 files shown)


blackshadowshade:
This may also fix #1440, but will need to be verified via testing.

cgolubi1:
What am i doing here?
* Playing back those 500 games i captured from buttonmen-master this morning, to look for any regressions
* Playing a bunch of games involving twin dice, to look for internal errors

cgolubi1:
So... here's another thing i found because of a regression, and am not sure what its state is.  Here's the situation --- BlackOmega's trip morphing 6 unsuccessfully Trip attacks grend's radioactive die... and splits.

I believe this demonstrates that this fix works in the morphing case --- the message changed:
```
-            'message' => 'responder004 performed Trip attack using [tm(6):4] against [%(4):3]; Attacker tm(6) rerolled 4 => 2; Defender %(4) rerolled 3 => 3, was not captured'
+            'message' => 'responder004 performed Trip attack using [tm(6):4] against [%(4):3]; Attacker tm(6) rerolled 4 => 2; Defender %(4) rerolled 3 => 3, was not captured; Attacker tm(6) showing 2 split into: tm(3) showing 3, and tm(3) showing 2'
```

and, to the extent that that's what actually happened, the message is an improvement.

However, the underlying behavior is a bug --- the `tm(6)` should not have split.  Is this a known issue or a new one?

unsuccessful_trip_split

cgolubi1:
Okay, so, in fact, this was discussed at length in #1432.  I dunno, the current behavior still feels wrong to me, but it's what we decided.  So it's not a bug.  Anyway, given that, the game above is a successful example of the fix on this ticket improving the situation.

irilyth:
> However, the underlying behavior is a bug --- the `tm(6)` should not have split.

Why not? Radioactive says "If a radioactive die is either the attacking die or the target die in an attack with a single attacking die and a single target die, the attacking die splits", which we've previously concluded means that the Trip die in a failed Trip Attack against a Radioactive die does split (because it's attacking, and Radioactive doesn't say anything about capture).

irilyth:
Ah, crossing e-mail there, ignore my last comment.

cgolubi1:
I ran 100 games including some combination of trip, radioactive, and twin dice.  No internal errors, and the action logs i spot-checked looked reasonable.  Merging this.

PR #1571, opened 2015-03-21
Chat now displays correct number of log entries, even when invisible entries are present

### LINKED ISSUE #1540: Internal error when calling submitChat (dev game #1536)
I had entered this text in the chat box for dev game 1536 (http://dev.buttonweavers.com/ui/game.html?game=1536):

> Yeah, ok, so:
> 
> > irilyth performed Skill attack using [o(Z=4):1,g`(2):2] against [(5/10=5):2]; Defender (5/10=5) was captured; Attacker o(Z=4) rerolled 1 => 1; Attacker g`(2) recipe changed from g`(2) to g(2), rerolled 2 => 1
> 
> This shouldn't be legal, because the Warrior Stinger 2 doesn't have Stinger while it has Warrior. (Or, at least, that's how the old site did it.)

I tried to edit this to say this:

> Yeah, ok, so:
> 
> > irilyth performed Skill attack using [o(Z=4):1,g`(2):2] against [(5/10=5):2]; Defender (5/10=5) was captured; Attacker o(Z=4) rerolled 1 => 1; Attacker g`(2) recipe changed from g`(2) to g(2), rerolled 2 => 1
> 
> This shouldn't be legal, because the Warrior Stinger 2 doesn't have Stinger while it has Warrior. (Or, at least, that's how the old site did it.)
> 
> (And that's how we wrote it up here: "While they are out of play, they are completely out of play: They aren't part of your starting dice, they don't count for initiative, they can't be attacked, none of their other skills can be used, etc." Of course, I wrote that text, so. :^)
> 
> (That text also contradicts a clarification from the old site: The new text implies that Berserk Warrior *is* usable, because Berserk doesn't take effect until the die loses Warrior.)

When I click "Change game message", I get "Internal error when calling submitChat", and the chat entry doesn't change.

I tried adding one line "Test adding one line.", and got the same result.

I tried deleting the chat text and replacing it with "Test deleting everything.", and got the same result.

I speculate that this has something to do with having used a Warrior die in the attack.

blackshadowshade:
Fixes #1540.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/656/ (if it passes)

### DIFF
--- a/src/engine/BMInterface.php
+++ b/src/engine/BMInterface.php
@@ -2899,9 +2899,6 @@ class BMInterface {
                      'game_state,action_type,acting_player,message ' .
                      'FROM game_action_log ' .
                      'WHERE game_id = :game_id ORDER BY id DESC';
-            if (!is_null($logEntryLimit)) {
-                $query = $query . ' LIMIT ' . $logEntryLimit;
-            }
 
             $statement = self::$conn->prepare($query);
             $statement->execute(array(':game_id' => $game->gameId));
@@ -2931,6 +2928,11 @@ class BMInterface {
                 }
             }
 
+            if (!is_null($logEntryLimit) &&
+                (count($logEntries) > $logEntryLimit)) {
+                $logEntries = array_slice($logEntries, 0, $logEntryLimit);
+            }
+            
             return $logEntries;
         } catch (Exception $e) {
             error_log(
--- a/test/src/api/responderTest.php
+++ b/test/src/api/responderTest.php
@@ -9722,7 +9722,6 @@ class responderTest extends PHPUnit_Framework_TestCase {
             array(array(1, 1)),
             array(array(0, array('value' => 2, 'sides' => 5, 'recipe' => '(5/10)')))
         );
-//        error_log(var_export($expData, TRUE));
         $expData['playerDataArray'][1]['optRequestArray'] = array(1 => array(1, 15));
         array_unshift($expData['gameActionLog'], array('timestamp' => 'TIMESTAMP', 'player' => 'responder003', 'message' => 'responder003 performed Skill attack using [o(Z=4):1,g`(2):2] against [(5/10=5):2]; Defender (5/10=5) was captured; Attacker o(Z=4) rerolled 1 => 1; Attacker g`(2) recipe changed from g`(2) to g(2), rerolled 2 => 1'));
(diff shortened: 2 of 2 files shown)


cgolubi1:
It occurs to me that this responder test conflicts with #1551, i.e. when i merge that one (assuming it works), this responder test will start failing.  I imagine we can deal with that problem, just making a note to myself.

blackshadowshade:
So, yes, you are correct that the merged responder tests fail. I'll deal with that.

blackshadowshade:
So, this should now be deconflicted.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/667/ (if it passes)

cgolubi1:
What i want to do here:
* Play back some games without ornery dice both before and after this change --- i'm concerned about a slowdown from loading all the action log entries, and i'll be difficult to get any information about speed, but i can try.  Also obviously verify that there are no changes in these games' action log sizes.
* Play a game with ornery dice by hand and verify that i always get 10 action log entries instead of some other number
* Play back some games with ornery dice --- these games should change, and should always load 10 action log entries rather than fewer.

Okay, so that first one: there are no regressions, but unfortunately the results of the performance comparison were not very good.
* Before:
```
Time: 16.78 minutes, Memory: 282.00Mb

OK, but incomplete, skipped, or risky tests!
Tests: 485, Assertions: 125999, Incomplete: 5.
385.12user 116.10system 16:49.46elapsed 49%CPU (0avgtext+0avgdata 959344maxresident)k
76528inputs+0outputs (720major+109824minor)pagefaults 0swaps
```

* After:
```
Time: 20.88 minutes, Memory: 282.25Mb

OK, but incomplete, skipped, or risky tests!
Tests: 486, Assertions: 126063, Incomplete: 5.
435.84user 123.01system 20:55.80elapsed 44%CPU (0avgtext+0avgdata 979968maxresident)k
80512inputs+0outputs (720major+112202minor)pagefaults 0swaps
```

It's only one data point, but, given this, we want to sit on this pull a bit more --- performance on prod is pretty bad, so exacerbating it this much seems unwise.

blackshadowshade:
One simple compromise might be to load some multiple of the originally requested number of entries (e.g. 30% more), instead of loading all entries. We'd need some reasonably good estimate of what this multiple should be, though.

cgolubi1:
Now that there have been some optimizations to action log loading, i'm going to return to this, and see how slow the tests are now.

cgolubi1:
Looks good.  I'm not able to identify a big slowdown any more, and the behavior in the face of ornery dice is definitely better.  Merging.

PR #1569, opened 2015-03-20
Fixed handling of asymmetric twin swing/non-swing dice in BMGame->getJsonData()

blackshadowshade:
When digging for a solution to #1550, I noticed that the code in BMGame->getJsonData() assumed that the first subdie of a twin swing die would be a swing die. This pull request fixes the code for the case where the first subdie is not a swing die.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/655/

PR #1576, opened 2015-03-25
Fixed bug that disallowed queer dice with non-positive integer values

### LINKED ISSUE #1574: Capturing a die split by radioactive twice seems to be broken
See #1569 for details.

blackshadowshade:
Fixes #1574.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/659/

### DIFF
--- a/src/engine/BMSkillQueer.php
+++ b/src/engine/BMSkillQueer.php
@@ -61,9 +61,7 @@ class BMSkillQueer extends BMSkill {
         if (FALSE ===
             filter_var(
                 $args['value'],
-                FILTER_VALIDATE_INT,
-                array("options"=>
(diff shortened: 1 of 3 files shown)


cgolubi1:
The explanation is reasonable.  I'm just going to test this one with the messy responder test i pasted into #1569, and call that good enough if it passes.

cgolubi1:
Actually, i lied --- i have all these games sitting around, and all it costs is time to play them back, and sometimes it turns up an unexpected regression.  So i'll run some of those games back as well, just to be on the safe side.

cgolubi1:
The safe side is safe.  I don't see any problems here.  Merging.

PR #1578, opened 2015-03-25
Mood now triggers after a split

### LINKED ISSUE #1570: Mood doesn't trigger on Berserk Mood Swing dice
When a Berserk Mood Swing die (Finngall has one of these) makes a Berserk Attack, the die halves (due to Beserk), but then does not change size when it rerolls (due to Mood Swing). See http://dev.buttonweavers.com/ui/game.html?game=1559 for an example of this in action.

blackshadowshade:
Fixes #1570.

This will need testing to see how it interacts with radioactive.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/662/

### DIFF
--- a/src/engine/BMSkillMood.php
+++ b/src/engine/BMSkillMood.php
@@ -28,7 +28,7 @@ class BMSkillMood extends BMSkill {
             return FALSE;
         }
 
-        if (empty($args['die']->value)) {
+        if (empty($args['die']->value) && !$args['die']->has_flag('HasJustSplit')) {
             return FALSE;
         }
 
--- a/test/src/engine/BMGameTest.php
+++ b/test/src/engine/BMGameTest.php
@@ -8581,6 +8581,66 @@ class BMGameTest extends PHPUnit_Framework_TestCase {
     }
 
     /**
+     * The following unit tests ensure that berserk mood dice apply mood after berserk attacks
+     *
+     * @coversNothing
+     */
+    public function test_berserk_mood() {
+        // load buttons
+        $button1 = new BMButton;
+        $button1->load('(1) (1) B(X)?', 'Test1');
+
+        $button2 = new BMButton;
+        $button2->load('(1,1) (1,1)', 'Test2');
+
+        // load game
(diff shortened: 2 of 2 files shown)


cgolubi1:
This is one of those deceptively simple pulls that should actually get a decent amount of regression testing.  I think RandomAI can do most of it, though.

What i need to do:
* Play a mood vs. radioactive game by hand, and make sure the behavior looks like what i'd expect
* There should be no behavior change at all for any games unless they contain both Mood and Radioactive dice.  So i should be able to replay a lot of games without both Mood and Radioactive, and see no regressions.
* Replay some games containing both Mood and Radioactive, and see what the changes look like
* Play some new RandomAI games with Mood and Radioactive and look for internal errors.

cgolubi1:
Oh, wait, my plan is a mess: this pull has nothing to do with radioactive.  Mood is one of the skills which is *lost* on a radioactive attack.  The expectation is that this pull changes how Mood interacts with *Berserk*, and shadowshade just suggested that i might want to look for Mood/Radioactive regressions (which there really shouldn't be any of, because Mood is lost on a Radioactive attack).  So, new plan, similar to the first, but better:
* Play a game with a Berserk Mood die by hand, and make sure the behavior looks like what i'd expect
* There should be no behavior change at all for any games unless they contain Berserk Mood dice. So i should be able to replay a lot of games without either Finngall or Jeffboyardy, and see no regressions
* Replay some games containing Finngall or Jeffboyardy, and see what the changes look like
* Play some new RandomAI games with Finngall, Jeffboyardy, and West Virginia, and look for internal errors

cgolubi1:
As shown in this screencap, the message on a berserk mood attack could be more informative, i.e. it could say that the die first split and then changed size:
berserk_mood_ambiguous_message

I don't think that's a blocker --- i'd be inclined to file it as a separate issue rather than hold up this pull over it.

cgolubi1:
Hmm, on second thought, i'm not even inclined to file an issue about the action log --- i think it's okay as is.  Anyway, i'm done testing this, and it's ready to merge once the freeze is lifted.

PR #1583, opened 2015-03-28
Implemented Time and Space

### LINKED ISSUE #263: Time and Space die skill needed
Time and Space: ^
Time and Space die appear within a triangle or pyramid.
- - - - - - - - - -
Time and Space Dice were created for the Looney Labs Button Man, Tirade. They are represented by a "^" before a die. If a Time and Space Die is rerolled after it participates in an attack and rolls odd, then the player will take another turn. If multiple Time and Space dice are rerolled and show odd, only one extra turn is given per reroll.

 Clarifications: •^o (Time & Space Ornery): Will only get an extra turn if the Ornery Time and Space Die attacks. If it is rerolled because of being Ornery, no extra turns will be given. 

 Inconsistencies: •^J (Time & Space Jolt): Will not give more than 1 extra turn if rerolled to an odd number.

blackshadowshade:
Fixes #263.

I'm open to discussion on including or excluding spaces from the name of the skill, i.e., "TimeAndSpace" or "Time and Space".

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/672/


src/engine/BMGameAction.php
```diff
@@ -571,6 +571,17 @@ protected function friendly_message_determine_initiative() {
         return $message;
     }
 
+    protected function friendly_message_play_another_turn() {
+        $message = $this->outputPlayerIdNames[$this->actingPlayerId] .
+                   ' gets another turn';
+
+        if ('TimeAndSpace' == $this->params['cause']) {
+            $message .= ' because a TimeAndSpace die rolled odd';
```
cgolubi1: IMO, even if the internal skill name is TimeAndSpace (which i'm fine with) the message should say "Time and Space"

cgolubi1:
Do we like the behavior shown in this image?  c2 got a bonus attack via TimeAndSpace as the last attack of the round, so the messages show up one after another.  Even if it's not ideal, it's not a big problem --- i'm on the fence about whether it's in fact the behavior we want (the alternative being, suppress the TimeAndSpace message if the player isn't actually going to get the turn because the round ended), so i figured i'd open the question on the ticket:
time_and_space_before_end_of_round

cgolubi1:
I did by-hand testing including T&S rolling odd numbers, T&S rolling even numbers, rolling more than one T&S die, capturing a T&S die, rolling T&S at the end of a round (see above), rolling T&S at the end of a game, and rolling T&S and getting another attack which i then had to pass.

Now i'm going to play back some misc games to look for regressions, play some T&S games looking for internal errors, and play some RandomBMMixed games looking for internal errors involving T&S.

cgolubi1:
Okay, all of that passed.  I'm inclined to go ahead and merge this, so people can test it.

Issue #1580, opened 2015-03-27
Speed up database access speed


blackshadowshade:
We have a number of indicators that game loading on the live site is slowing down as the number of games increases. Chaos and I would like to start a discussion on how to speed up database access speed. One consideration is that the memory available may be a currently-limiting factor: on the live site, it is 1706184 kB, on the T1 micro instances, it is 609452 kB.

Calling @nkmcalli, @irilyth, and other people with experience in DB optimisation.

cgolubi1:
So, i did some poking around while waiting for a test database to slowly do some things, and raised the following half-baked questions for myself based on looking at our mysql global status ("show global status"):
* Would using mysql large_pages help us?
* Is it a problem that Innodb_os_log_written is nearly as big as Innodb_data_written (i.e. are we using too many transactions unnecessarily)?
* Is Qcache_hits low relative to Queries because our query cache is too small?
* Select_full_join is nonzero, which the MySQL manual says is a sign our indexes are insufficient.  Select_scan is also very large, which similarly seems like bad design

I looked into that last item in more detail, and found the missing indices for game_action_log and game_chat_log --- i think the game_action_log index issue is actually responsible for a substantial amount of our pain on the prod site.

If we decide to look further: http://stackoverflow.com/a/637357 brought me around to the possibility that selecting from views in programmatic access is always a bad idea --- the issue is that the view gets executed first, and the where clauses you put on your select apply to that result.  So if the view brings in an entire big table, it doesn't matter if you're doing a clever select on the outside, you had to bring in the entire table first.  This is in a nutshell why loading the "Buttons" tab is so slow, because it loads the entire buttons table 70-odd times, once per buttonset.  To remedy this, we should look into which tables are the biggest (and thus most expensive to load unnecessarily) and prioritize writing selects against the base tables rather than loading from views.

blackshadowshade:
So, obvious candidates for indexing include:
- button id
- button name
- player name
- game id

blackshadowshade:
So, another idea that Chaos had was to export the die flag information to JSON only when it matters. I'll put up a pull request for that now.

jimmosk:
Is this the cause of the dramatic slowage that I've been experiencing on the site in the past several weeks? It's getting worse; right now it can take two or three minutes for the dev site (and about as long for the www site) to respond to my clicking anything.

irilyth:
[...]

PR #1584, opened 2015-03-28
BMDie now only reports on die flag information when it is non-trivial

blackshadowshade:
Addresses part of #1580.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/673/


cgolubi1:
This is a good test case for just playing back some games and calling it good --- should make no difference.  In principle, i should do some testing to make sure that games recorded before the update still work, but that's not easy to do even semi-automatically, so i think i'm going to just assume it works because it doesn't look like there should be any problems --- nothing in BMGameAction became more strict, after all.

cgolubi1:
Seems fine.  Thanks.

PR #1594, opened 2015-04-11
Time and Space now triggers correctly for twin dice

### LINKED ISSUE #1593: Time & Space (1,Y=2) die rolled odd but didn't give another turn
In http://dev.buttonweavers.com/ui/game.html?game=1599:

    irilyth performed Skill attack using [gn(X=20):10,`^(1,Y=2):3] against [gn(X=4):4]; Defender gn(X=4) was captured; Attacker gn(X=20) rerolled 10 => 19; Attacker `^(1,Y=2) recipe changed from `^(1,Y=2) to ^(1,Y=2), rerolled 3 => 3

But I didn't get another turn. Screenshot:

bm

blackshadowshade:
Fixes #1593. Also addresses the action logging wording left over from comments on #1583.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/678/

### DIFF
--- a/test/src/api/responderTest.php
+++ b/test/src/api/responderTest.php
@@ -246,6 +246,12 @@ class responderTest extends PHPUnit_Framework_TestCase {
                     'Shadow' => 'Dice with both Shadow and Stinger skills can singly attack with any value from the min to the max of the die (making a shadow attack against a die whose value is greater than or equal to their own, or a skill attack against a die whose value is lower than or equal to their own)',
                 ),
             ),
+            'TimeAndSpace' => array(
+                'code' => '^',
+                'description' => 'If a Time and Space Die is rerolled after it participates in an attack and rolls odd, then the player will take another turn. If multiple Time and Space dice are rerolled and show odd, only one extra turn is given per reroll.',
+                'interacts' => array(
+                ),
+            ),
             'Trip' => array(
                 'code' => 't',
                 'description' => 'These dice can also make Trip Attacks. To make a Trip Attack, choose any one opposing die as the Target. Roll both the Trip Die and the Target, then compare the numbers they show. If the Trip Die now shows an equal or greater number than the Target, the Target is captured. Otherwise, the attack merely has the effect of re-rolling both dice. A Trip Attack is illegal if it has no chance of capturing (this is possible in the case of a Trip-1 attacking a Twin Die). IMPORTANT: Trip Dice do not count for determining who goes first.',
@@ -9519,7 +9525,7 @@ class responderTest extends PHPUnit_Framework_TestCase {
     }
 
     /**
-     * @group fulltest_deps
+     * @depends test_request_savePlayerInfo
      */
     public function test_interface_game_030() {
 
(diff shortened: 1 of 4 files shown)


test/src/api/responderTest.php
```diff
@@ -9741,4 +9747,127 @@ public function test_interface_game_030() {
...
+        // responder003's idle ornery dice rerolled at end of turn: Ho(W=4)? changed size from 4 to 12 sides, recipe changed from Ho(W=4)? to Ho(W=12)?, rerolled 2 => 8
+
+        $this->verify_api_submitTurn(
+            array(1, 2, 3, 8),
+            'responder003 performed Power attack using [^B(T=3,T=3):4] against [(20):4]; Defender (20) was captured; Attacker ^B(T=3,T=3) rerolled 4 => 3. responder003 gets another turn because a Time and Space die rolled odd. responder003\'s idle ornery dice rerolled at end of turn: Ho(W=4)? changed size from 4 to 12 sides, recipe changed from Ho(W=4)? to Ho(W=12)?, rerolled 2 => 8. ',
+            $retval, array(array(0, 3), array(1, 0)),
+            $gameId, 1, 'Power', 0, 1, '');
+
+        $this->update_expected_data_after_normal_attack(
+            $expData, 1, array('Power', 'Skill'),
```
cgolubi1: Hmm, i'm suspicious --- the expected data update shorthand probably obscures this, and i should have stuck the word BUG in a few more places (for both of which pieces of unclarity i apologise), but i think the first argument is "what should the value of `$activePlayerIdx` be after this attack?"  And it should still be 0.  So i suspect we got the game action log message, but the fix didn't actually work.  Let me try this in a sandbox and get you a screencap either way.

cgolubi1:
Huh, nope, your fix does work.  I wonder what i'm looking at...

cgolubi1:
Ran some regression test games, and they look fine too.  Merging.

test/src/api/responderTest.php
blackshadowshade: Ah, I must have forgotten that. I've obviously manually changed these attributes afterwards (see lines 9864-9866) to get the test to pass. I'm happy for you to update the test to match the correct syntax.

jimmosk:
Actually, I think I'm still encountering this bug. In game 1614 this happened:

(2015-04-27 20:49:22)	jimmosk performed Power attack using [^B(T=3,T=3):3] against [p(4):3]; Defender p(4) was captured; Attacker ^B(T=3,T=3) rerolled 3 => 3.  
But then my opponent got to go.

Except, wait, at that point I was left with no legal attacks. So maybe it did give me an extra turn, and the actual bug was that it didn't display that I had gotten an extra turn nor that I had autopassed... I can't tell.

irilyth:
Looking at the log, I think your opponent got a turn before you you passed, and that you didn't get an extra turn. (Also, if you do get an extra turn, but then have to pass, I'm pretty sure that gets logged -- I think I've seen it.)

I think this fix has been merged to buttonmen-dev:master, but may not have been deployed to the site yet.

cgolubi1:
Ah, you're right.  I'll do a deployment to dev tonight.

Issue #1129, opened 2014-07-23
Fire overshooting


blackshadowshade:
Since the current interpretation of Fire allows increasing the attacker die value for a Power attack past the value of the target die:
- the guidance text provided by the UI needs to be corrected
- the player should be given the option of reducing the value of the Fire dice even when the current Power attack is valid without help

PR #1595, opened 2015-04-11
Changed fire message to add proviso about Power attacks

blackshadowshade:
Addresses part of #1129.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/679/ (if it passes)


blackshadowshade:
Actually, I'm going to close this because I think I've found a better way to address #1595.

Issue #1143, opened 2014-07-30
Handling attacks with post-selection decisions


irilyth:
Some attacks require you to make some additional decisions after you select the dice that will participate in the attack, and before the attack is actually completed. (Fire and Turbo are two obvious examples.) Figure out how best to handle these, from a UI and/or backend perspective.

irilyth:
Updating labels: I think this isn't a "question" any more; and, based on https://github.com/buttonmen-dev/buttonmen/issues/1143#issuecomment-50697848, it sounds like this is purely a UI thing.

cgolubi1:
So, as Shadowshade suggests, it could be done as purely a UI thing, but it'd be a somewhat complicated UI thing in that case, since the UI would need to make one API call, check its result, and make another API call based on what it got from the first call.  It might not be too bad, and you might even be able to use some of the existing plumbing for form submission calls, but it could be a slight pain to debug.

dwvanstone:
Here's an issue that (even with years and years of playing) I keep running into again with Fire dice. I suspect it's a newbie issue as well.

When I want to use the Fire die in a skill (or power) attack, I select the Fire die as part of the attack.

The response I get is "There is no valid attack corresponding to the dice selected."


I see Jota suggested a solution to this issue with a "checkbox for 'Use Fire dice to assist?'" option in issue https://github.com/buttonmen-dev/buttonmen/issues/1164. That would be a fine solution.

Another would be adding to the Attack menu:

  Power Attack
  Power Attack using Fire dice
  Skill Attack
  Skill Attack using Fire dice
  ...

irilyth:
Yep, I think implementing the kind of interface that Jota and I were talking about here would (a) be less confusing for newbies (and experienced players); (b) obsolete a whole bunch of other issues with the Fire turndown page (by eliminating it); and (c) be a nice UI for Turbo Swing too, a skill that shows up on a bunch of (not yet available) buttons.

I don't know that anyone's opposed to it, I think it's just that no one's had the right combination of time and expertise to do it.

blackshadowshade:
I'm tossing up how to implement the front-end proposed by @AdmiralJota, at least for Turbo Swing. At some stage, the UI needs to know whether a die has Turbo or not, and the question for me is whether we pass this information explicitly, or whether we parse the die recipe for an exclamation mark.

Also, Jota, do you happen to have any code samples / CSS lying around for your proposed front-end?

blackshadowshade:
[...]

PR #1596, opened 2015-04-12
Allow fire overshooting even when the original Power attack is valid

### LINKED ISSUE #1129: Fire overshooting

blackshadowshade:
Fixes #1129.

This definitely needs a corresponding responder test. Let me know if you don't like anything about the solution that I've proposed here.

Known issue: the upper limit in the fire turndown message currently isn't restricted by the maximum amount that the Fire dice can be turned down, since this information is not easily available.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/680/ (if it passes)


blackshadowshade:
This fails Jenkins, but I don't know why, so I'll wait until another pull request has been accepted before trying again.

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/693/ (if it passes)

blackshadowshade:
This is still failing Jenkins during PHPUnit, and I don't know why.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/693/console

cgolubi1:
What do you expect to happen next in terms of debugging why this failed Jenkins?  I don't have any idea why either.

blackshadowshade:
I'm going to wait until a few more pull requests have been merged, then do a new merge and try Jenkins again.

blackshadowshade:
Ah, finally! JSHint is failing! I can finally fix this pull request.  :)

blackshadowshade:
And hooray, a pass on Jenkins!

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/708/

src/ui/js/Game.js
```diff
@@ -866,12 +868,24 @@ Game.actionAdjustFireDiceActive = function() {
...
+
+  fireMessage += '.';
+
   Game.page.append($('<div>', {
-    'text': 'Turn down Fire dice by a total of ' +
-            (defenderSum - attackerSum) +
-            ' to make up the difference between the sum of your attacking' +
-            ' dice (' + attackerSum + ') and the defending die value (' +
-            defenderSum + ').',
+    'html': fireMessage,
```
cgolubi1: I don't believe there's any reason to change this div from 'html' to 'text' - it's still plain text even if there's now more of it, so let's treat it that way to be on the safe side.

cgolubi1:
Hmm... so, overloading validAttackTypeArray seems weird to me, because you already picked the attack type, so it's not so much "valid" as "in progress".  I can't point to a specific thing it'll break, though; i just think it's wrong.  I'm not going to reject it just because i think it's wrong.

cgolubi1:
Anyway, testing:
* Play some games with Fire by hand to see what the UI looks like
* Regression-test a bunch of games without Fire and make sure nothing changes
* Regression-test a bunch of games with Fire, and expect them to fail on message changes due to power attacks or because more attacks became valid, and not otherwise
* Autoplay a bunch of games with Fire and look for internal errors

src/ui/js/Game.js
blackshadowshade: Ah, that's a relic from when I was playing around with adding extra <br>s to the message. Sure, I can change that back.

cgolubi1:
This doesn't seem to work --- in the situation in which i make a power attack and the attacker's value is greater than the defender's, there are two problems:

Problem 1: i get dumped into the fire turndown window, but the original power attack was not registered as the first part of a fire attack.  So there's no action log entry for the first half of the attack, and thus the UI shows me "loaded data for game" rather than a reasonable message about what i did:
power_no_turndown_needed_bad_message

Problem 2: the UI sanity-checking for fire turndown assumes that if you are there, you need to turndown at least one side.  So if i'm in this situation, there's actually no way to complete the attack *without* turning down a fire die, because the UI rejects my claim that i am doing a "turndown" action if i don't actually turn a die down (and, of course, if i cancel the attack it cancels the attack):
power_no_turndown_needed_impossible_to_continue

cgolubi1:
I have to be honest --- even if you fix these bugs, my prediction is that it will be pretty annoying if, whenever you're playing a game with fire dice, you unexpectedly get dumped into the fire turndown screen after a power attack, and have to spend time saying "yes, yes, i actually am satisfied with this attack the way i submitted it the first time".

Basically, i think having Fire turndown fully work with Power attacks will make Fire dice more annoying to play.  And, no, i can't prove that Fire is only supposed to work with Skill.  I just think Fire would be better if it only worked with Skill, due to game ease-of-use considerations.

As a concrete comment, you definitely need to change the fire turndown text in some fashion, to make it clear to the player that if they just want to do their power attack and don't want to turndown anything, they should select "turndown" and click "submit".  The bug is that it doesn't *work* to select turndown and click submit.  But even once that bug is fixed, i think it's not very intuitive, and people won't be expecting to be thrown into the fire turndown screen in this situation, so we'd better make what to do when you're there intuitive.

irilyth:
> I have to be honest --- even if you fix these bugs, my prediction is that it will be pretty annoying if, whenever you're playing a game with fire dice, you unexpectedly get dumped into the fire turndown screen after a power attack, and have to spend time saying "yes, yes, i actually am satisfied with this attack the way i submitted it the first time".

I agree; I think https://github.com/buttonmen-dev/buttonmen/issues/1143 would fix this nicely. Is that more like a fun UI thing, or more like super complicated?

blackshadowshade:
Honestly, seeing how long Jota took to get some of his pull requests accepted, I would expect that this would take two weeks of programming effort (including the to-ing and fro-ing on the pull request), and that's not counting the month of learning curve time to even get to the stage of being able to do the programming.

So, is it super-complicated? Probably no, but getting it to a stage where both Chaos and I are happy with it will take a while.

blackshadowshade:
I think that this should now be fixed. I've added the "Submit attack without turning down fire dice" action, which is added to the list of fire actions (becoming the default action) when this action is valid.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/716/ (if it passes)

jimmosk:
On the non-dev site, I just got the following message:

screen shot 2015-05-21 at 9 40 26 am

I see two problems:
1) "fire your attacking die" is confusing wording. My "attacking die" is my z(12). Perhaps it should say "It is not clear if you wish to use your Fire die to assist in the attack"?
2) How will the ambiguity over whether I want to use the Fire die be resolved if I choose a different setting than "Default Attack"? At first I thought that choosing Speed Attack or Skill Attack would cause the attack to go ahead without any further questions, but choosing Power Attack would give me the "how much do you want to turn down your Fire die?" message, so I did the latter... and got no such screen; the Power attack just went through as a one-on-one attack.

blackshadowshade:
There's an outstanding pull request addressing this issue, and I'd like to see what you think of it after Chaos merges it.

blackshadowshade:
Regarding the wording though, is it not standard nomenclature to "fire a die" in the rules?

jimmosk:
Huh, so to you "fire a die" means "increase an attacking die by means of decreasing another Fire die"?  No, that's not how I've seen that verb used.  For example, in https://github.com/buttonmen-dev/buttonmen/wiki/Tester-guide:-supported-skills is says,
"Dice with both Fire and Mighty skills do not grow when firing, only when actually rolling."
So in that case "firing" is what the fire die does to itself when it decreases to raise another die.

irilyth:
I feel like I've seen "to fire" as a verb used to describe what a Fire die does, somewhere, but now I can't find it. I thought it might have been in http://www.buttonweavers.com/ui/forum.html#!threadId=129 but I don't see it there now, and I don't see it anywhere else obvious either. But I agree that my impression was that it's something the Fire die does.

blackshadowshade:
In that case, sure, I'll change the message.

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/728/

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/799/

jimmosk:
Just following up on my comment above with the screencap on May 21. 
I was just in the same situation, got the Default Attack Is Ambiguous alert, chose Skill attack this time... and *still* wasn't given the option to turn down fire dice -- the skill attack went through as a one-on-one attack.  So how the heck would I ever actually be able to tell it I did want to use fire dice?

blackshadowshade:
jimmosk, we are still waiting for this pull request to be accepted. Once it goes through, that issue should be resolved.

blackshadowshade:
BTW, in this situation, if you choose Skill attack, there *is* no option to use Fire dice in this situation. The only way to use Fire dice here is with a Power attack.

jimmosk:
So until the pull request is accepted there isn't any way to use a Fire die in this situation?  (See my May 21 comment, where I mentioned that choosing Power Attack didn't allow me to use a Fire die.)

irilyth:
If you select a die with a showing-value equal to or greater than a target die, then no, right now there's no way to use a Fire die in that attack. This ticket will change that, but hasn't yet.

cgolubi1:
This doesn't really have a test plan any more, so let's make a new one.
* Look at the new code and make sure i understand the changes (i've kind of done this already, so it should be quick)
* Play some fire games by hand, with both skill and power attacks
* Regression-test non-fire games and expect no changes
* Regression-test fire games, and probably expect no changes (anything that was previously legal should still work even if the new code makes more things legal)
* Play some new RandomAI fire games and make sure there are no internal errors

cgolubi1:
Here's a bug - when you choose to do a power attack and get taken into the turndown screen, the message from the server is "Loaded data for game N".  That's definitely a blocker - we ruthlessly quashed instances of that last year and it took a long time.  No way i'm introducing a new one.  :>)

fire_loaded_data_message

cgolubi1:
Onward, though, this is a UI antipattern, and i'm not sure it's a bug, but i think it probably is, and i think it might not be too hard to fix.  A story in N screencaps:

I go to make my power attack, and am prompted to turndown dice, with the default being not to turn down:
turndown_sequence_1

I think, "Oh, i would like to turn down a die!", so i do:
turndown_sequence_2

Then i click "Submit", and get an error because i didn't move the selector off "Submit attack without turning down fire dice":
turndown_sequence_3

So far so good --- something's gotta be the default after all, and i said i thought no turndown should be the default, and i still stand by that.  But here's the kicker.  Now i think "Oh, right, right, good point" and move the slider to "Turn down fire dice":
turndown_sequence_4

Then i click Submit, and get *another* error, because my turndown value got lost:
turndown_sequence_5

That second error is the one that's gratuitous --- i'm somewhat confident that if you modified `Game.formAdjustFireDiceActive` and, in the new `no_turndown` case, parsed the valid values into `Game.activity` the same way is done in the `turndown` case, those values would be retained after an invalid submission.  So the player who was trying to do optional turndown on Power would get the first error (which i think is fine) but not the second.

src/ui/js/Game.js
```diff
@@ -867,12 +869,24 @@ Game.actionAdjustFireDiceActive = function() {
...
+  if (('Power' == attackType) &&
+    (attackerDiffFromMax > exactFiringAmount)) {
+    fireMessage += 'between ' + Math.max(0, exactFiringAmount) +
+                   ' and ' + attackerDiffFromMax;
+  } else {
+    fireMessage += exactFiringAmount;
+  }
+
+  fireMessage += '.';
+
```
cgolubi1: Nit: arguably all this logic here is evidence that the backend should be telling the UI more about what's valid.  But i don't think you need to tackle that today.

test/src/api/responderTest.php
```diff
@@ -4676,7 +4676,7 @@ public function test_interface_game_008() {
         array_unshift($expData['gameActionLog'], array('timestamp' => 'TIMESTAMP', 'player' => 'responder003', 'message' => 'responder003 chose to perform a Skill attack using [tm(8):3] against [(8):7]; responder003 must turn down fire dice to complete this attack'));
         $expData['gameActionLogCount'] = 7;
         $expData['gameState'] = 'ADJUST_FIRE_DICE';
-        $expData['validAttackTypeArray'] = array();
+        $expData['validAttackTypeArray'] = array('Skill');
```
cgolubi1: Fine.  I still don't like overloading `validAttackTypeArray`, but i guess in this case it's fairly unambiguous.  It'll bite us eventually, but hopefully not today.

irilyth:
Two thoughts about the UI on this:

If you change the number next to a Fire die, should you also have to select another option from another menu? Why not just say "changing the number means you want to turn down the value, not changing the number means you don't"? We could put up some (red?) text somewhere, saying "About to turn down Fire dice!" or something, if we want to make it more obvious. Or have the Submit button say "Don't turn down Fire dice" and change it to a red "Turn down Fire dice" if you change a value.

Is this going to bring up an extra screen every time you make a Power Attack in a game with Fire dice? I think that's really non-ideal. :^( I think https://github.com/buttonmen-dev/buttonmen/issues/1143 would vastly improve this situation, if anyone was interested in tackling that.

cgolubi1:
Okay, here's my responderTest diff which reproduces the message bug by tacking an extra couple of moves onto existing responder test 008:
```php
index 9324110..a55c9b8 100644
@@ -4686,6 +4686,44 @@ class responderTest extends PHPUnit_Framework_TestCase {
         $_SESSION = $this->mock_test_user_login('responder001');
         $this->verify_api_loadGameData_as_nonparticipant($expData, $gameId, 10);
+
+
+        ////////////////////
+        // Move 07 - responder003 abandons the Fire-assisted Skill attack and gets another attack
+        // [tm(8):3, sF(20):7] => [(4):1, (8):7, (12):11, z(20):17]
+        $this->verify_api_adjustFire(
+            array(),
+            'Cancelled fire attack.',
+            $retval, $gameId, 1, 'cancel', NULL, NULL);
+
+        array_unshift($expData['gameActionLog'], array('timestamp' => 'TIMESTAMP', 'player' => 'responder003', 'message' => 'responder003 chose to ab
+        $expData['gameActionLogCount'] = 8;
+        $expData['gameState'] = 'START_TURN';
+        $expData['validAttackTypeArray'] = array('Power', 'Skill', 'Shadow', 'Trip');
+        $expData['playerDataArray'][0]['activeDieArray'][0]['properties'] = array();
+        $expData['playerDataArray'][1]['activeDieArray'][1]['properties'] = array();
+        $retval = $this->verify_api_loadGameData($expData, $gameId, 10);
+
+
+        ////////////////////
+        // Move 08 - responder003 chose to perform a Power attack using [tm(8):3] against [(4):1]; responder003 may turn down fire dice to complete t
+        // [tm(8):3, sF(20):7] => [(4):1, (8):7, (12):11, z(20):17]
+        $this->verify_api_submitTurn(
+            array(),
+// BUG: the API message should be something informative
+//            'responder003 chose to perform a Power attack using [tm(8):3] against [(4):1]; responder003 may turn down fire dice to complete this at
+            'Loaded data for game ' . $gameId . '.',
+            $retval, array(array(0, 0), array(1, 0)),
+            $gameId, 1, 'Power', 0, 1, '');
+// BUG: something like this should be appended to the action log.
+//        array_unshift($expData['gameActionLog'], array('timestamp' => 'TIMESTAMP', 'player' => 'responder003', 'message' => 'responder003 chose to 
+//        $expData['gameActionLogCount'] = 9;
+        $expData['gameState'] = 'ADJUST_FIRE_DICE';
+        $expData['validAttackTypeArray'] = array('Power');
+        $expData['playerDataArray'][0]['activeDieArray'][0]['properties'] = array('IsAttacker');
+        $expData['playerDataArray'][1]['activeDieArray'][0]['properties'] = array('IsAttackTarget');
+
+        $retval = $this->verify_api_loadGameData($expData, $gameId, 10);
     }
```

cgolubi1:
To answer irilyth's comment: yes, this will bring up a new screen every time someone makes a power attack which could be turned down.  I don't think this is going to make the site better --- i'm pretty sure i've been on the record before as saying that i think Fire turndown of Power attacks is a sufficiently marginal benefit that it's not worth any drop in playability at all for players who aren't using it.

However, with the solution James is currently working on, it brings up the additional screen, and if you don't want to use fire, you only have to click one button ("Submit").  My preference would still be to just drop it, but i think this solution is an acceptable compromise given that other players do think fire turndown of power attacks is important, and that we are not going to implement #1143 immediately.  (One suggestion is, implement #1143 for Turbo, and once we come up with a UI we like, retrofit it for Fire too.  But i don't know how much that would delay Turbo, and i'm wandering off-topic for this issue.)

cgolubi1:
Okay, something else i found by accident while testing new randomAI games: if you talk to the API and choose action 'turndown' but don't specify any dice to turndown, you get an internal error:
```
[Sun Aug 16 14:27:15 2015] [error] [client 10.0.2.2] PHP Warning:  Invalid argument supplied for foreach() in /var/www/engine/BMInterface.php on line 3710, referer: http://localhost:8080/ui/game.html?game=1
```

The UI won't let you do that, but the API certainly will --- it shouldn't throw an internal error.

cgolubi1:
Okay, i did everything on my test plan, and i'm left with three issues:
* Fix the action logging and message return so that you don't return "Loaded data for game N"
* Ideally retain the turned down values in the UI if the player turns down values but leaves the action set to 'no_turndown'
* Fix internal error if the API receives an action of 'turndown' but no turned-down values

For the first issue, i'll note that, if you want to be consistent with the way the fire UI is currently implemented, what you need to do is:
1. Add an entry to the action log for the first half of the attack
2. Add a second entry to the action log with the results of what was done on the fire turndown screen

That is, the action logging should work by the same principles in the case of power overshooting, as it does for any other type of fire attack.

blackshadowshade:
So, I think I've addressed all the issues here.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/825/ (if it passes)

cgolubi1:
Okay, these changes aren't very major.  I think what i need is:
* Record some games with fire dice before i update my VM, then play them back, and make sure nothing has regressed problematically
* Play some games by hand and make sure the action logging looks good now
* Verify that when i make mistakes in the fire turndown UI, i get reasonable results
* Have random_AI send turndown with no turned down values, and verify that i no longer get an internal error

cgolubi1:
API change looks good:
```
...
ValueError: API adjust_fire_dice(255, turndown, [], [], 1, 1440679342) unexpectedly failed: At least one fire value must be turned down for a turndown action
```
and no new error log message when that happened.

cgolubi1:
So... am i being too picky?  I still don't like the action logging.  It's not as obviously broken, but i'll show you.

Here's two consecutive messages when you choose to turn down dice:
```
(2015-08-27 08:47:55)  c2 chose to perform a Power attack using [vz(20):5] against [F(4):1]; c2 must decide whether to turn down fire dice
(2015-08-27 08:47:55)  c2 turned down fire dice: wHF(4) from 4 to 1; Defender F(4) recipe changed to Fv(4), was captured; Attacker vz(20) rerolled 5 => 11
```

Here's two consecutive messages when you don't choose to turn down dice:
```
(2015-08-27 08:47:39)  c2 chose to perform a Power attack using [vz(20):9] against [F(6):2]; c2 must decide whether to turn down fire dice
(2015-08-27 08:47:41)  c2 performed Power attack using [vz(20):9] against [F(6):2]; Defender F(6) recipe changed to Fv(6), was captured; Attacker vz(20) rerolled 9 => 5
```

So, in the first case, the second message starts by describing what happened during the turndown action.  That seems good.

In the second case, the second message starts by repeating what happened in the first message, and doesn't describe what happened during the turndown action.  That seems less good.

cgolubi1:
Okay, i have some ideas about how we should handle the interaction of getting this last thing fixed and the September prod release, but there's a smoke detector in an unreachable location that needs attention and won't stop chirping (maybe it has something to do with all the fire overshooting...), so i have given up on formulating them.  Ping me on IRC.

cgolubi1:
We talked about this yesterday and decided we'd like to push this out earlier in the month anyway to give players more time with it.  So i think that means shadowshade will address this last small issue, and then we'll merge it early next month and let players try it out on dev.

blackshadowshade:
So, I've made that change for the action logging in the no_turndown case.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/846/ (if it passes)

cgolubi1:
Okay, so here i'll need to:
* Collect some games on this branch before i pull, then play them back, to make sure the changes are expected
* Look at the visual behavior and make sure it seems okay

cgolubi1:
This passed my tests, so i'm going to go ahead and merge it.  Once i push it to dev, we'll be opening a quest (opportunity for BFotM) for people to discuss what they think and drive a consensus about what to do.

PR #1601, opened 2015-04-22
Disabled Maximum and Mood (and Mad) when a die has Konstant

### LINKED ISSUE #1597: Konstant Maximum Focus reroll effect contradicts log message
In http://www.buttonweavers.com/ui/game.html?game=5283 I have a kfM(20), and during initiative, I used Focus to turn it down to show 14. When I attacked with it, the log said

> irilyth performed Skill attack using [k(20):18,kfM(20):14] against [M(4):4]; Defender M(4) was captured; Attacker k(20) does not reroll; Attacker kfM(20) does not reroll

as if Konstant took precedence over Maximum, but the die changed its showing-value to 20, as if Maximum took precedence.

Here's a screenshot:

bm

blackshadowshade:
Fixes #1597.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/685/ (if it passes)

### DIFF
--- a/src/engine/BMDie.php
+++ b/src/engine/BMDie.php
@@ -171,7 +171,7 @@ class BMDie extends BMCanHaveSkill {
 
             // if all hook results are FALSE, then roll the die
             if (empty($hookResultArray) ||
-                (0 == count(array_filter($hookResultArray, function($value) { return $value !== FALSE; })))) {
+                (0 == count(array_filter($hookResultArray, function ($value) { return $value !== FALSE; })))) {
                 $this->set__value(bm_rand($this->min, $this->max));
             }
         }
--- a/src/engine/BMDie.php
+++ b/src/engine/BMDie.php
@@ -167,7 +167,13 @@ class BMDie extends BMCanHaveSkill {
 
         if (!isset($this->value) ||
             ($this->doesReroll && !$this->has_flag('JustPerformedTripAttack'))) {
-            $this->set__value(bm_rand($this->min, $this->max));
+            $hookResultArray = $this->run_hooks(__FUNCTION__, array('die' => $this));
+
+            // if all hook results are FALSE, then roll the die
+            if (empty($hookResultArray) ||
+                (0 == count(array_filter($hookResultArray, function($value) { return $value !== FALSE; })))) {
+                $this->set__value(bm_rand($this->min, $this->max));
+            }
         }
 
         $this->run_hooks('post_roll', array('die' => $this,
@@ -593,7 +599,7 @@ class BMDie extends BMCanHaveSkill {
         if ($this->has_flag('HasJustSplit')) {
             $actionLogInfo['recipeBeforeSplitting'] = $this->flagList['HasJustSplit']->value();
         }
-        
+
         return($actionLogInfo);
     }
 
--- a/notes/engine/List_of_die_hooks_and_skills.txt
+++ b/notes/engine/List_of_die_hooks_and_skills.txt
@@ -6,8 +6,9 @@ add_skill
   Value
 pre_roll
   Mood
-post_roll
+roll
   Maximum
+post_roll
   Warrior
   TimeAndSpace
 attack_list
--- a/src/engine/BMSkillMaximum.php
+++ b/src/engine/BMSkillMaximum.php
@@ -15,22 +15,22 @@ class BMSkillMaximum extends BMSkill {
      *
      * @var array
      */
-    public static $hooked_methods = array('post_roll');
+    public static $hooked_methods = array('roll');
 
     /**
-     * Hooked method applied after rolling a die
+     * Hooked method applied while rolling a die
      *
      * @param array $args
      * @return boolean
      */
-    public static function post_roll(&$args) {
+    public static function roll(&$args) {
         if (!($args['die'] instanceof BMDie)) {
             return FALSE;
         }
 
         $die = $args['die'];
 
-        if (!$die->doesReroll) {
+        if (!$die->doesReroll && isset($die->value)) {
             return FALSE;
         }
 
--- a/src/engine/BMSkillMaximum.php
+++ b/src/engine/BMSkillMaximum.php
@@ -29,6 +29,11 @@ class BMSkillMaximum extends BMSkill {
         }
 
         $die = $args['die'];
+
+        if (!$die->doesReroll) {
+            return FALSE;
+        }
+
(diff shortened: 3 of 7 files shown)


blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/687/ (if it passes)

blackshadowshade:
For information, a die with Maximum now does not require an extra random number when rolling to its maximum value.

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/691/ (if it passes)

src/engine/BMDie.php
```diff
@@ -290,7 +290,13 @@ public function roll($isTriggeredByAttack = FALSE, $isSubdie = FALSE) {
...
         if (!isset($this->value) ||
             ($this->doesReroll && !$this->has_flag('JustPerformedTripAttack'))) {
-            $this->set__value(bm_rand($this->min, $this->max));
+            $hookResultArray = $this->run_hooks(__FUNCTION__, array('die' => $this));
+
+            // if all hook results are FALSE, then roll the die
+            if (empty($hookResultArray) ||
+                (0 == count(array_filter($hookResultArray, function ($value) { return $value !== FALSE; })))) {
+                $this->set__value(bm_rand($this->min, $this->max));
+            }
```
cgolubi1: I don't understand why this code change has the desired effect:
* the if() statement on line 291 has a condition that either you have no value or `$this->doesReroll` has to be set
* the only time your new hooks are going to return false is when `$this->doesReroll` is *not* set

So it seems like the combination of all the new code should be that when you have a `kM` die, it exits this code with `$this->value` still unset, and freaks out.

blackshadowshade: The point here is that the FALSE response from the hook is telling the code not to reroll the die *because* the die value has been set within the die hook (which currently can only be triggered by Maximum).

cgolubi1:
This fix does seem strangely complicated to me.  It seems like if `!$die->doesReroll` simply caused the roll step not to happen, it wouldn't be necessary to individually hook all the other things that happen at roll time.  So i'm curious why we can't do it that way.

I looked at the code further, and, as noted inline, i'm not convinced that it *does* work --- i bet your test gets around the problem by setting `$die->value`, but this actual game behavior comes into play when `$die->value` is not set.

That said, the purpose of responderTest and RandomAI and all those is to let me focus on testing the external behavior, and not get too tied up in my comprehension of the underlying implementation or lack thereof.  So, given all of the previous, my plan of attack is:
* Test my theory that this new code doesn't work by playing a RandomBM game with kM dice and seeing if i get internal errors when the kM die would reroll
* Assuming my theory is wrong:
    * This change should have no effect on any non-RandomBM buttons, so replay a bunch of games and look for regressions
    * Replay a bunch of RandomBM games and see what changes
    * Play a bunch of new RandomBM games and look for internal errors

src/engine/BMDie.php
cgolubi1: Within the `roll()` die hook?  I'm staring at the logic of your diff and just not seeing what you say here --- what i see is that on lines 33-34, `roll()` returns FALSE *if* the die value is already set.  So where does it return FALSE *and* set the die value?  Enh, there's no point arguing about it --- i'm provisionally willing to believe that you're right and i'm just blind here, and i'll test it and find out.

cgolubi1:
Okay, well, in fact, this does seem to work in a by-hand game.  I don't understand why it works, but it does:
fkm_doesnt_reroll

So i'll go ahead and do the rest of my testing.

src/engine/BMDie.php
blackshadowshade: Ah sorry, got the response around the wrong way.

A TRUE response from the die hook says that the die already has a die value set (which happens to be the maximum value), and thus the following bit within the if statement doesn't run.

cgolubi1: Ah hah!  That's better.

cgolubi1:
Okay, this seems fine.  At some fundamental level, i kind of don't get why the hook-based approach relies on the participation of the other skill, vs simply having `!$die->does_reroll` cause the `roll` step to be skipped.  But it works, so i'm not going to knock it.

cgolubi1:
Well, that was a slight mess --- i clicked "Merge" without thinking about the fact that we're in a merge freeze.  So i created and merged a revert, and now i can verify after pulling from upstream:
```
$ git log -4
commit 41254ded9515d0476229b43dfbb93d75f842f524
Merge: 2eaa51b e9d9c65
Author: cgolubi1 <walrus-github@glassonion.org>
Date:   Sun May 3 10:15:15 2015 -0400

    Merge pull request #1612 from buttonmen-dev/revert-1601-1597_ts_konstant
    
    Revert "Disabled Maximum and Mood (and Mad) when a die has Konstant"

commit e9d9c65c7378679c691a5bc835dd4f7be20331b2
Author: cgolubi1 <walrus-github@glassonion.org>
Date:   Sun May 3 10:14:47 2015 -0400

    Revert "Disabled Maximum and Mood (and Mad) when a die has Konstant"

commit 2eaa51b1c242ae9fc3ce4385165f2231a56486a9
Merge: 36d4670 03b117a
Author: cgolubi1 <walrus-github@glassonion.org>
Date:   Sun May 3 10:14:39 2015 -0400

    Merge pull request #1601 from blackshadowshade/1597_ts_konstant
    
    Disabled Maximum and Mood (and Mad) when a die has Konstant

commit 36d46702316d6591cd1c27c5d25abd51ff841aa1
Merge: 3811a38 2178d08
Author: cgolubi1 <walrus-github@glassonion.org>
Date:   Fri Apr 24 20:34:20 2015 -0400

    Merge pull request #1573 from blackshadowshade/1572_focus_warrior
    
    Warrior dice now cannot react to initiative
```

and:
```
leocrotta,[~/src/git/cgolubi1/buttonmen-master],10:16(0)$ git diff 36d46702316d6591cd1c27c5d25abd51ff841aa1
leocrotta,[~/src/git/cgolubi1/buttonmen-master],10:17(0)$ 
```

So i think i successfully reverted, but how do i in fact merge this when i'm ready?  I'll go figure that out now.

cgolubi1:
Okay, there we go --- i created a new pull to revert the revert, and it has the same contents as the original pull.  So i think we're okay, and i'll merge that one tomorrow.

PR #1602, opened 2015-04-22
Made konstant description more precise, and added interaction descriptions

### LINKED ISSUE #1587: Clean up Konstant help text
The help text for Konstant dice says "These dice do not reroll after an attack", but the actual rule is more like "After the first roll of the round, any time a Konstant die is re-rolled, it always rolls the same value that it was showing", because we concluded that you can re-roll them if they're the target of a Trip Attack, if they're also Chance dice, if they're also Ornery, etc, they just don't change their showing-value.

Also, the clarifications portion of the help text mentions how Chance and Konstant interact; it might be worth generalizing that to say "if another skill (such as Chance, Trip, Ornery, etc) causes the die to re-roll, it continues to show the same value". And, perhaps, also to say something like "if another skill causes the die to change its showing-value without rerolling (such as Focus, Fire, etc), the die's value does change, and then continues to show that new value".

blackshadowshade:
Fixes #1587.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/686/

### DIFF
--- a/src/engine/BMSkillChance.php
+++ b/src/engine/BMSkillChance.php
@@ -58,7 +58,8 @@ class BMSkillChance extends BMSkill {
     protected static function get_interaction_descriptions() {
         return array(
             'Focus' => 'Dice with both Chance and Focus skills may choose either skill to gain initiative',
-            'Konstant' => 'Dice with both Chance and Konstant skills always retain their current value',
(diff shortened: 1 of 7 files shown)


cgolubi1:
No reason not to accept this on the strength of the existing responderTest passing, and the new text looks fine to me.  I'll merge it when the freeze is lifted.

PR #1579, opened 2015-03-26
Addresses multiple issues regarding RandomBM

blackshadowshade:
Addresses part of #779.

See commit message for details of changes.

Requires database update 00779_05_RandomBM.sql

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/663/ (if it passes)

PR #1605, opened 2015-04-26
Added boom skill

### LINKED ISSUE #506: Boom Dice
Boom: b
the die frame is a bomb

---

Boom Dice were created for the Unexploded Cow set by Cheapass Games.  They are like normal dice and they have an additional attack, the Boom attack. 
 
Boom Attack: To make a Boom Attack, remove one of your Boom Dice from play (neither player will score it). Choose one of your opponent's dice, and reroll it. Note that the targeted die is never captured, just re-rolled.

blackshadowshade:
Fixes #506.

This pull request needs additional responder and Javascript tests before it's really complete. Also, the unit tests in BMBtnSkillRandomBMMixedTest.php will conflict with the changes to RandomBM in #1579.

Requires database update 00506_boom_01.sql

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/698/

### DIFF
--- a/deploy/database/data.game.sql
+++ b/deploy/database/data.game.sql
@@ -29,7 +29,7 @@ INSERT INTO die_status (name) VALUES
 ('NORMAL'),
 ('CAPTURED'),
 ('DISABLED'),
-('OUT_OF_GAME'),
+('OUT_OF_PLAY'),
 ('DELETED'),
 ('SELECTED'),
 ('DIZZY');
--- a/deploy/database/updates/00506_boom_01.sql
+++ b/deploy/database/updates/00506_boom_01.sql
@@ -1,0 +1,3 @@
+UPDATE die_status
+SET name = 'OUT_OF_PLAY'
+WHERE name = 'OUT_OF_GAME';
--- a/notes/engine/List_of_die_hooks_and_skills.txt
+++ b/notes/engine/List_of_die_hooks_and_skills.txt
@@ -12,6 +12,7 @@ post_roll
   TimeAndSpace
 attack_list
   Berserk
+  Boom
   Fire
   Konstant
   Queer
@@ -38,6 +39,7 @@ assist_values
   Fire
 capture
   Berserk
+  Boom
   Doppelganger
   Morphing
   Null
--- a/src/engine/BMAttackBoom.php
+++ b/src/engine/BMAttackBoom.php
@@ -1,0 +1,116 @@
+<?php
+/**
+ * BMAttackBoom: Code specific to boom attacks
+ *
+ * @author james
+ */
+
+/**
+ * This class contains code specific to boom attacks
+ */
+class BMAttackBoom extends BMAttack {
+    /**
+     * Type of attack
+     *
+     * @var string
+     */
+    public $type = 'Boom';
+
+    /**
+     * Determine if there is at least one valid attack of this type from
+     * the set of all possible attackers and defenders.
+     *
+     * If $includeOptional is FALSE, then optional attacks are excluded.
+     * These include skill attacks involving warrior dice.
+     *
+     * @param BMGame $game
+     * @param boolean $includeOptional
+     * @return boolean
+     */
+    public function find_attack($game, $includeOptional = TRUE) {
+        $targets = $game->defenderAllDieArray;
+
+        return $this->search_onevone($game, $this->validDice, $targets);
+    }
+
+    /**
+     * Determine if specified attack is valid.
+     *
+     * @param BMGame $game
+     * @param array $attackers
+     * @param array $defenders
+     * @return boolean
+     */
+    public function validate_attack($game, array $attackers, array $defenders) {
+        $this->validationMessage = '';
+
+        if (count($attackers) != 1) {
+            $this->validationMessage = 'There must be exactly one attacking die for a trip attack.';
+            return FALSE;
+        }
+
+        if (count($defenders) != 1) {
+            $this->validationMessage = 'There must be exactly one target die for a trip attack.';
+            return FALSE;
+        }
+
+        if ($this->has_dizzy_attackers($attackers)) {
+            // validation message set within $this->has_dizzy_attackers()
+            return FALSE;
+        }
+
+        if (!$this->are_skills_compatible($attackers, $defenders)) {
+            // validation message set within $this->are_skills_compatible()
+            return FALSE;
+        }
+
+        return TRUE;
+    }
+
+    /**
+     * Check if skills are compatible with this type of attack.
+     *
+     * @param array $attArray
+     * @param array $defArray
+     * @return boolean
+     */
+    protected function are_skills_compatible(array $attArray, array $defArray) {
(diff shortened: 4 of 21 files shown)


cgolubi1:
Okay.  Obviously there's a lot of work to be done on the test end on this one, so i don't want to make any rash promises about timing.  (But if we can do this, why can't we do Rage dice, which have no UI component, and would let me play Rold?  `:>)`  Anyway.)

As a note to myself, the conflict with #1579 is just that the numeric IDs for the various skills will change once #1579 is merged, so that's no cause for alarm, and doesn't conflict with *testing* this. 

So, in terms of testing, i should:
* Do the standard database update test
* Do a test in which i load some games in progress and completed games into the DB before updating to make sure there's no backwards-compatibility issue with previous games
* Play a game with Boom dice myself and report some initial impressions of the UI and whatnot
* Create a responderTest for a game with Boom dice
* Create a UI test to go with the responder test
* There should be no regressions to any existing games, so replay a number of RandomAI games looking for regressions
* Play a number of RandomAI games with boom dice looking for internal errors

cgolubi1:
Okay, first issue: it looks to me like boom dice can *only* perform boom attacks.  Here's output where i explicitly performed a power attack with a boom die:
boom_power_attack

But the effect was what i think the boom effect should be --- the attacking die was taken out of play, and the target die rerolled.

I don't know anything about boom dice, so for all i know they can't do power attacks.  But the description of the boom skill lists Boom as an additional attack, not a replacement for other attacks.  So i'm pretty sure this is a bug.

cgolubi1:
Second bug, which you can see in the same issue --- the action log entry doesn't say that the target die rerolled from 6 => 1, just that it wasn't captured.  It looks to me like this is missing logic in `BMGameAction::messageDefender()` --- you need to be able to invoke `message_value_change()` as needed in the absence of `$defenderRerollsEarly`.

cgolubi1:
Looks like the first bug is that any attack containing Boom dice turns into a Boom attack, since when i do a skill attack containing both Boom and non-Boom dice, i get an internal error:
skill_attack_with_boom_dice

The internal error is:
```
[Sun May 03 16:47:56 2015] [error] [client 10.0.2.2] PHP Warning:  assert(): Assertion failed in /var/www/engine/BMSkillBoom.php on line 42, referer: http://localhost:8080/ui/game.html?game=2
```

cgolubi1:
Third bug: you have probably a simple copy/paste-o in the error message for whether a Boom attack can be performed:
boom_attack_failure_says_trip

The error message says "There must be exactly one attacking die for a trip attack.", and it should say "boom attack".

cgolubi1:
The same issue exists if there's more than one defending die, i can see from the code.

cgolubi1:
Okay.  Let's start with those, and i'll get back to my test plan once they're fixed.  Obviously, until the "all attacks are boom attacks" and "internal error on skill attack" bugs (which i assume are the same bug under the hood) are fixed, it doesn't make any sense to run RandomAI tests or make a responder test.

cgolubi1:
This needs merge help in addition to the bugfixes, FYI.

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/707/

cgolubi1:
If i make a Default attack in which a single Boom die targets another die:
* If the attacker is showing a lower value than the defender, the Default attack is interpreted as a Boom attack
* If the attacker is showing a higher value than the defender, the Default attack is interpreted as a Power attack

The first one of those seems okay, but the second is wrong.  The backend needs to report that the attack is ambiguous when Power/Skill and Boom are both valid options, the same way it would do for Power/Berserk.

cgolubi1:
What do you think the interaction between Boom dice and other skills that limit capture should be?  I assumed that the purpose of Boom was to evade most restrictions on target dice, but i see:
```
Stealth dice cannot be the target of boom attacks.
```

That seems wrong to me --- the description of Stealth says that Stealth dice cannot be *captured* by any attack other than multi-die skill attacks, and we reasonably concluded that that means they can't be *targeted* by Trip attacks because Trip attacks might capture them.  But Boom attacks can't capture their targets.  So IMO Boom vs. Stealth should be legit.

cgolubi1:
You haven't addressed the second bug which i mentioned in https://github.com/buttonmen-dev/buttonmen/pull/1605#issuecomment-98502429 - the action log doesn't mention the target die rerolling after a boom attack.

cgolubi1:
Okay, so, rundown of the problems found so far.  At the moment, i'm playing a bunch of RandomBMMixed games with Boom dice to look for more internal errors.  There's still a fair bit of testing to do per above, but enough of the issues i'm mentioning here need logic changes that i'm waiting on it:
1. Game action log entry needs to say what value the target die rerolled.
2. Default attack assumes Power when Boom is also an option - this case should be reported as ambiguous.
3. Should stealth dice be targetable by Boom attacks?  The code says no, i think yes.

Item 3 isn't a blocker to dev testing, since dev testers may have opinions (but they can have opinions now too --- i don't think this is a "you have to see it in action to decide what you think about it" type question).  The other two are.

blackshadowshade:
So, I think I've addressed all the issues.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/713/

cgolubi1:
This needs merge help (unsurprisingly, given the modification affecting RandomBMMixed).

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/717/ (if it passes)

blackshadowshade:
I'll add the interaction with Turbo once #1607 has been accepted.

cgolubi1:
I can test this without the Turbo interaction change, since that change will only affect the interaction descriptions, and Turbo isn't implemented yet.

cgolubi1:
Okay, testing which remains:

* Play some games with Boom dice myself and verify that the three things i mentioned above have been fixed
* Play a number of RandomAI games with boom dice, and with RandomBMMixed, looking for internal errors
* There should be no regressions to any existing games, so replay a number of RandomAI games looking for regressions
* Create a responderTest for a game with Boom dice (maybe i can pick an interesting sample from one of the RandomAI games)
* Create a UI test to go with the responder test

cgolubi1:
Hmm, i noticed an API inconsistency here --- when activeDieArray is empty, the API passes it as an empty array.  Ditto for capturedDieArray.  But when outOfPlayDieArray is empty, the API doesn't pass it.

I'm inclined to file this as a separate bug rather than hold up this pull request, but IMO it *is* a bug --- testing against our API is painful because it's so internally inconsistent, so i don't think we should make it worse.

blackshadowshade:
I'm happy to fix the API inconsistency once boom has been accepted, since it shouldn't affect any other UI functioning.

cgolubi1:
I created a new issue, #1640, for the API inconsistency.

cgolubi1:
I now have a responder test to submit for Boom dice.  I'm going to merge this, then submit the responder test as a pull request.

Issue #91, opened 2013-10-10
Turbo swing skill needed


blackshadowshade:

jimmosk:
I'm eager for another skill to join the party, so I'd like this to be my BFotM for March.

jimmosk:
I'm going to roll this over into April; it's still the enhancement I'm most interested in seeing next (now that Time and Space is accomplished).

jimmosk:
How about seeing what advances can get made towards Turbo?  I'm picking this as my BFotM.

blackshadowshade:
So, I'm starting to work on this. I'd like people to start discussing how Turbo Swing should interact with other skills. Consider, but don't feel bound by, how the interactions worked on the old site. It's also useful to me to hear if you have no preference on certain interactions.

Skills that I think might be worth discussing:
- Berserk
- Chance
- Konstant
- Mad/Mood
- Mighty/Weak
- Morphing
- Ornery
- Radioactive

dwvanstone:
I remember turbo option dice on the old site (such as 1/30!) but I don't remember turbo swing dice (X!). I'm sure they were there but maybe I just went from one extreme to the other all of the time.

First pass!

Turbo + Mad/Mood: Turbo allows you to control the change every time you roll it, but the Mad/Mood quality comes in when it is rolled for some other purpose (such as when the opponent trips it or if it is ornery). That allows us to use both of the qualities.

Turbo + Berserk: The Berserk halves the die size, but as soon as you reroll the die you get to choose the size again. (The size limits are not halved.) 

Turbo + Radioactive: The Radioactive text already says that we lose the Turbo characteristic. I think that's fine.

Turbo + Ornery: When you roll, you can turbo; when you don't roll, it rerolls randomly.

Turbo + Mighty/Weak: When you roll, you can turbo to the original sizes; when you don't roll, mighty/weak comes into play.

Turbo + Chance: I think since you are controlling the Chance die being rerolled you can change its size at that moment too.

Turbo + Morphing: When you roll, you can turbo to the original sizes; when you don't roll, morphing comes into play.

Turbo + Konstant:  Ugh. What happens now for a konstant ornery weak die? Does it keep its value but change # of sides? Do we have the precedent that a die can have 4 sides but show a 7?

blackshadowshade:
You may need to rethink Turbo + Morphing.

[...]

PR #1607, opened 2015-04-28
Added new game state to accommodate Turbo swing

blackshadowshade:
Addresses part of #91.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/700/


src/engine/BMSkillTurbo.php
```diff
@@ -0,0 +1,47 @@
...
+     * @return string
+     */
+    protected static function get_description() {
+        return 'After your starting roll, you may change the size of a ' .
+               'Turbo Swing die or a Turbo Option die every time you roll it. ' .
+               'Decide on a size first that is valid for the Swing or Option '.
+               'type, then roll the new die as usual. If a Turbo Die is ' .
+               'rerolled by a player other than the player who has the Turbo ' .
+               'die (e.g., during a trip attack), then the size cannot be ' .
+               'changed.';
```
cgolubi1: The phrasing of this description feels a little awkward to me, but i don't have a better suggestion offhand.  If anyone wandering by wants to take a stab at wordsmithing, go for it.  Otherwise i'll try to come up with some concrete suggestions in the near future.

src/engine/BMSkillTurbo.php
```diff
@@ -0,0 +1,47 @@
...
+    /**
+     * Descriptions of interactions between this skill and other skills
+     *
+     * An array, indexed by other skill name, whose values are descriptions of
+     * interactions between the relevant skills
+     *
+     * @return array
+     */
+    protected static function get_interaction_descriptions() {
+        return array();
```
cgolubi1: IMO, the comment you put in the main description about the interaction between Turbo and Trip should be listed as an interaction description between Turbo and Trip.

test/src/engine/BMGameTest.php
```diff
@@ -5759,6 +5773,10 @@ public function test_berserk_round() {
         $this->assertGreaterThanOrEqual(1, $game->actionLog[1]->params['postAttackDice']['attacker'][0]['value']);
 
         $game->update_game_state();
+        $this->assertEquals(BMGameState::CHOOSE_TURBO_SWING, $game->gameState);
```
cgolubi1: Is it expected that games without turbo dice should move through `BMGameState::CHOOSE_TURBO_SWING`?  I didn't think games without chance/focus dice moved through `BMGameState::REACT_TO_INITIATIVE`, but maybe i'm wrong about that.

cgolubi1:
Okay, i made various code comments above, but, modulo those, the upshot of this is really just that i should run some regression tests of arbitrary games and declare victory.

test/src/engine/BMGameTest.php
blackshadowshade: At the moment, every game goes through every game state that exists, and if the game state is irrelevant, the code moves on to the next game state.

src/engine/BMSkillTurbo.php
cgolubi1: Okay, so, why does this seem awkward to me?  Basically it just feels too wordy for the amount of information:
* why say "a Turbo Swing die or a Turbo Option die" rather than "a Turbo Swing or Option die"?
* "Decide on a size first that is valid for the Swing or Option type, then roll the new die as usual" - this feels both too verbose and insufficiently informative.  I feel like it sort of implies that you have to take an action to reroll the die after "deciding on a size".  Maybe something more like "Before each reroll, you will get the chance to select any valid size for the Swing or Option type"?
* "If a Turbo Die is rerolled by a player other than the player who has the Turbo die" - again, this seems awkward.  Do we really not have a more concise phrasing for "the player who owns a die"?

cgolubi1:
I don't see any regressions due to the logic changes, so i want to iterate on the wording for the description, add the interaction description between Turbo and Trip, and then this should be good to go.

blackshadowshade:
@jimmosk, how would you rework the wording?

blackshadowshade:
Let's give this a go.

Base description:
After your starting roll, you may change the size of a Turbo Swing or Option die every time you roll it. Decide on a size first that is valid for the Swing or Option type, then roll the new die as usual.

Interactions with Boom, Trip
If a Turbo Die is rerolled because it is the target of a Boom/Trip attack, then the size cannot be changed.

Would these be okay?

cgolubi1:
On Fri, 08 May, 2015 at 15:23:59 -0700, James Ong wrote:

> Let's give this a go.
> 
> Base description:
> After your starting roll, you may change the size of a Turbo Swing or Option die every time you roll it. Decide on a size first that is valid for the Swing or Option type, then roll the new die as usual.
> 
> Interactions with Boom, Trip
> If a Turbo Die is rerolled because it is the target of a Boom/Trip attack, then the size cannot be changed.
> 
> Would these be okay?

Sure, i'd accept these.

Chaos

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/718/ (if it passes)

cgolubi1:
Okay.  Thanks.

Issue #974, opened 2014-06-15
BMInterface has too many public methods


blackshadowshade:
The last PMD warning, which is that BMInterface has too many public methods, I think is a valid concern.

I was wondering what my fellow devs consider about splitting out the forum-related methods into a new class, BMInterfaceForum.

blackshadowshade:
I'm pretty sure we're done with this, so I'm closing this.

PR #1611, opened 2015-05-02
Refactored part of BMInterface into BMInterfacePlayer

blackshadowshade:
Addresses part of #974.

So, I've put this pull request up for review so that we can discuss whether this approach that I'm using (and the infrastructure that I've had to add) is okay, or if there are things that we should change at this early stage.

Once this is accepted, I can then do similar refactoring for much of the rest of BMInterface.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/701/ (if it passes)


blackshadowshade:
One obvious question to answer is whether BMInterface->cast() should actually do in-place replacement of the BMInterface object, instead of returning an alternative BMInterface* object. There is one caveat, which is that I don't know if the in-place replacement is even possible in PHP.

cgolubi1:
So, here's my first question: i don't see the intuitive logic of the implementation here, in which BMInterfacePlayer extends BMInterface.  It seems to be adding a lot of code complexity (all of this "cast" business), both in BMInterface* itself and in apiResponder, and i don't understand why it is advantageous to made the interface more complex in this way.

I haven't thought about this as much as you have, but it seems like the issue is that the way ApiResponder is implemented right now, it always creates an object of type `BMInterface`, and sometimes you want an object of type `BMInterfacePlayer`.  But if that's all there is to it, it seems like the fix should be very easy --- `ApiResponder::process_request()` already decides whether to create a `BMInterface` or a `BMInterfaceNewUser`.  Why not simply extend *that* logic in the logged-in case, and create a `BMInterfacePlayer` instead of a `BMInterface` if the API method being invoked is one which needs one?

To summarize, my initial feedback is that i'm not convinced of the merit of the "cast"-based approach.  Tell me more about what problems it's the best practical solution for, and why it's the best practical solution for those problems.

blackshadowshade:
The first issue here is that we are calling BMInterface methods from other BMInterface methods, so we need an explicit mechanism for changing type.

The second issue is that having ApiResponder::process_request() determine the type of BMInterface* class requires having a huge switch or doing some function name analysis, which will need to be updated every time that we add new functionality to BMInterface*. This is just another thing that will make future programming and debugging more tricky.

However, I accept that having ApiResponder do the selection is probably the correct way if we see a performance hit from establishing multiple connections (unless I can also transfer the BMInterface->conn during the cast).

cgolubi1:
We don't have that many API methods, relatively speaking --- it doesn't seem like that big a deal to me to have ApiResponder know which ones are BMInterfacePlayer methods and which ones are BMInterfaceGame methods and which ones are BMInterfaceForum methods, and instantiating the correct object, the same way we *already* have ApiResponder know which ones are BMInterface methods and which ones are BMInterfaceNewUser methods.

> The first issue here is that we are calling BMInterface methods from other BMInterface methods, so we need an explicit mechanism for changing type.

Don't understand this either --- if we start out with `BMInterfacePlayer extends BMInterface`, then BMInterfacePlayer should already have access to all the BMInterface methods it needs to call, provided they are protected and not private.

Put another way: to me, adding all this extra complexity seems worse than doing nothing and just dealing with BMInterface being too large.  It seems to me that if there's a clean division between Player methods and other methods, then the decision about which BMInterface* to use should be simple enough that it can be done with a simple switch in ApiResponder.  If there's not a clean division, then we shouldn't create a messy division just for the sake of having a division.

If there's something obvious i'm not getting here, be patient with me and run me through it again.

blackshadowshade:
I'm happy to try to implement this with a switch statement in ApiResponder. Hopefully, I can make it somewhat smart, so that it won't just be a giant list of all functions in ApiResponder.

Regarding the first point, though, I'm thinking of the case where we have a number of different BMInterface* classes, and each is using methods from multiple other BMInterface* classes, not just the parent class. In this case, we'll still need a method to cast from one BMInterface* class to another.

cgolubi1:
> Regarding the first point, though, I'm thinking of the case where we have a number of different BMInterface* classes, and each is using methods from multiple other BMInterface* classes, not just the parent class. In this case, we'll still need a method to cast from one BMInterface* class to another.

To me, this still seems like a sign that something is wrong with this approach.  If you're looking at a BMGame object and want to get information about a button in the game, you don't cast the game to BMButton --- and if i suggested you should, you'd look at me funny.

Instead, you have a definition of BMGame, which is a game object, and BMButton, which is a button object, and there's a relationship in which a BMGame contains some BMButtons.

IMO, the only reason we think BMInterface works differently, is that it's kind of a misc bin rather than a tightly-defined object.  I don't want to revisit incredibly old arguments, but i think a lot of the complexity here comes from the decision to define BMInterface as both the front door (between the API and the rest of the code) and the back door (between each object and the database).  Anyway, though, we are where we are, and i'm not seriously suggesting revisiting that decision --- it'd be a huge amount of work and we can get some benefits without doing it.

Anyway, here's another idea to try on for size: instead of casting: (1) define relationships between different BMInterface objects the same way relationships between other real objects work, (2) allow a new BMInterface object to be instantiated with an existing database connection, and (3) make the initialization of a BMInterface object take a list of what other children are needed (e.g. `new BMInterface(['Player', 'Forum'], $this->isTest)`).  That way, the interface between multiple BMInterface objects is clean (e.g. "Okay, now i need a forum function, so i'm calling `$this->forum->some_public_function()`").

You'd still need logic in ApiResponder to figure out what BMInterface object you need, but the thing is, your proposed solution here has logic in ApiResponder *right now* to figure out what BMInterface object it needs --- you're defining which functions need the cast and which ones don't by duplicating the cast code all over the place.

blackshadowshade:
Okay, I'm happy to consider each of the BMInterface* classes as separate entities, mostly unrelated. I wonder, though, whether a BMInterfacePlayer object should even have a 'forum' property, because that really doesn't make logical sense. The analogy with BMButton and BMGame breaks down here because there is no sense of ownership between the BMInterface* classes.

If we do consider the BMInterface* classes to be separate, then we simply need to instantiate a new BMInterface* object when required, using the same connection and isTest state. That sounds like we need a new constructor, which when given a BMInterface object as an optional input, copies the connection and the properties from the input object.

How does that sound?

cgolubi1:
That sounds pretty reasonable to me - i agree that there might be no reason for BMInterfacePlayer to have a Forum object --- maybe the front door is always BMInterface, and it's the one that links to all the other objects needed to handle your particular request, each of which just links back to it, or something?  See what works.

Anyway, this general approach seems fine to me.  Thanks.

blackshadowshade:
Just a status update on this: I hate PHP's fake object orientation. In particular, the problem I'm hitting here is that there is no provision for a class to call getter functions internally when accessing one of its own properties. Anyway, it looks like I'm going to have to do something uglier than what might be possible if we were using a real object oriented language.

blackshadowshade:
Okay, what do you think about the approach that I've taken here?

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/722/ (if it passes)

cgolubi1:
FYI, there are merge conflicts here.  I assume you haven't bothered because i'm supposed to be commenting on the approach, but just a heads-up.

src/api/ApiResponder.php
```diff
@@ -144,7 +109,7 @@ protected function get_interface_response_createGame($interface, $args) {
         if (isset($retval)) {
             foreach ($playerIdArray as $playerId) {
                 if (isset($playerId)) {
-                    $interface->update_last_action_time($playerId, $retval['gameId']);
+                    $interface->player()->update_last_action_time($playerId, $retval['gameId']);
```
cgolubi1: I'm not convinced this actually works in this case --- it looks from the code below like in this case $interface is a BMInterfacePlayer object, so why do we need $interface->player() here?

src/api/ApiResponder.php
```diff
@@ -651,6 +610,72 @@ public function process_request($args) {
...
+
+    protected function select_interface_type($funcname) {
+        if (array_search($funcname, $this->playerFunctions, TRUE)) {
+            $interface = new BMInterfacePlayer($this->isTest);
+        } else {
+            $interface = new BMInterface($this->isTest);
+        }
+
+        return $interface;
+    }
```
cgolubi1: Not totally sure this is necessary --- in cases like the createGame case, can you just call `$interface->player()->create_game()`?  Ah, looking at more of the code, i see that at least part of the problem is that, in the createGame case, too much logic is in ApiResponder.  I wonder if this would be a good time to move more logic to BMInterfacePlayer, but i also don't think it's strictly necessary.

Anyway.  I think this is fine, and it is approximately what i had in mind (to the extent that i can remember) --- it's just a little untidy that every function needs to call `$interface->player()` to set the modification time.  Makes me wonder if all authenticated functions should instantiate both `$this->interface` and $this->interfacePlayer` or something.

cgolubi1:
> Just a status update on this: I hate PHP's fake object orientation. In particular, the problem I'm hitting here is that there is no provision for a class to call getter functions internally when accessing one of its own properties.

FWIW, things like changing from `$this->message =` to `$this->set_message()` don't bother me in the slightest --- it's basically just changing from an implicit setter/getter to an explicit one, and i know it's a lot of find/replace, but it doesn't strike me as particularly ugly.

cgolubi1:
Anyway, the short version is that i like this approach just fine --- i'm not sure if i'm supposed to be testing it now, or if you wanted an okay on paper now and you'll convert more and then have me test it, or what.  If you do want it tested now, i'll put it through the standard battery of RandomAI-ish regression checks this weekend, though that'll be easier once the merge conflicts are fixed because of changes to the API output in other pulls which have been merged.

blackshadowshade:
I'll deal with the merge conflicts and then this will be ready for testing. I'd like to take this refactoring in steps that are as small as possible, so each new BMInterface* class will be its own pull request, I think.

src/api/ApiResponder.php
blackshadowshade: It's correct that the interface is already a BMInterfacePlayer object. However, it's not a costly cast, and it keeps the code consistent.

The point is that the programmer shouldn't need to know what sort of interface is being used, rather, each time a certain interface function is being called, the only important thing to know is what sort of interface provides that function.

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/775/

src/api/ApiResponder.php
cgolubi1: Oh, i see.  Yeah, okay.

So the alternative i was thinking of is not to have a `select_interface_type()` function, and instead generally return a BMInterface, and on line 100 above use `$interface->player()->create_game()`.

I don't think i feel strongly --- it seems a little inconsistent that on line 100 you're saying "well, we know this is a BMInterfacePlayer object" and on line 112 in the same function you're saying "we can't make assumptions about what kind of BMInterface object this is".  But i don't think it's a big deal.

cgolubi1:
Okay, so, nits notwithstanding: this is a refactor, so what we want to see is that it doesn't break anything.  I should certainly play back a bunch of games and play some new games, but game replay isn't very diverse in how it accesses the API, so what i need to do is:
* Replay some games
* Play some new games
* Enumerate all the API functions, and try something by hand which accesses each one, ideally with all possible sets of optional args.

cgolubi1:
Here's the list of all API functions:
```
adjustFire
countPendingGames
createForumPost
createForumThread
createGame
createUser
dismissGame
editForumPost
joinOpenGame
loadActiveGames
loadActivePlayers
loadButtonData
loadButtonSetData
loadCompletedGames
loadForumBoard
loadForumOverview
loadForumThread
loadGameData
loadNextNewPost
loadNextPendingGame
loadOpenGames
loadPlayerInfo
loadPlayerName
loadPlayerNames
loadProfileInfo
login
logout
markForumBoardRead
markForumRead
markForumThreadRead
reactToAuxiliary
reactToInitiative
reactToReserve
savePlayerInfo
searchGameHistory
submitChat
submitDieValues
submitTurn
verifyUser
```

cgolubi1:
None of that led to any additional entries in the error log, so i think this pull works.  I'm going to go ahead and merge it --- shadowshade, take a look at my suggestion earlier about more consistently using `$interface->player()` instead of sometimes instantiating a `BMInterfacePlayer` object and sometimes using `$interface->player()`.  If you like that idea, you can integrate it into the next pull in this set of refactors, and if you don't, not a big deal.

blackshadowshade:
I like your suggestion of not instantiating the BMInterfacePlayer. I'll see if I can incorporate that in a small pull in the next couple of days.

PR #1618, opened 2015-05-12
Fixed redirect location after dismissing a completed game

### LINKED ISSUE #1246: Preference to always go to Monitor mode
There's a preference "Automatically Monitor after "Next game" runs out" right now, but this doesn't affect other things that would take you to the Overview page, like dismissing a game from the end-of-game page. As a user of this feature, I'd like it to be more like "Always Monitor when sent to Overview page", with an exception for when you actually explicitly click on "Overview" in the header -- but basically anything else that's going to send you to Overview should send you to monitor instead. (Or should go to http://dev.buttonweavers.com/ui/index.html?mode=preference if that's how this is implemented?)

I'd vote for replacing the current preference with this, rather than adding a new one, but we could poll people who use the current one to see if they prefer the current behavior.

blackshadowshade:
Fixes #1246.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/720/

### DIFF
--- a/src/ui/js/Game.js
+++ b/src/ui/js/Game.js
@@ -1432,7 +1432,7 @@ Game.formDismissGame = function(e) {
     messages,
     $(this),
     function() {
-      window.location.href = Env.ui_root;
+      window.location.href = Env.ui_root + 'index.html?mode=nextGame';
       return false;
     },
     Game.showLoggedInPage
--- a/src/ui/js/Game.js
+++ b/src/ui/js/Game.js
@@ -1432,7 +1432,7 @@ Game.formDismissGame = function(e) {
     messages,
     $(this),
     function() {
-      window.location.href = Env.ui_root + 'index.html?mode=nextGame';
+      window.location.href = Env.ui_root + 'index.html?mode=preference';
       return false;
     },
     Game.showLoggedInPage


cgolubi1:
I'll have to try this out to see how it works, and take a look at the code.  Not sure offhand if this should be "nextGame" or "preference" or something else.  Basically i just need to do some tests by hand, and make sure things look right both when i'm in monitor mode and when i'm not.

cgolubi1:
Okay, this definitely isn't right.

When i dismiss a game, regardless of whether i have the "Automatically Monitor after "Next game" runs out" preference selected, i get taken directly to my next waiting game.  That's not right --- the only way i should get taken to the next waiting game is (possibly) if i have both automonitor *and* "Redirect to waiting games when in Monitor mode" selected.

If i only have automonitor selected, i should get taken to the monitor mode overview.  And if i don't have automonitor selected at all, i should get taken to the general overview.

I bet you want "preference" instead of "nextGame", but i haven't looked into it.

blackshadowshade:
Okay, updated.

cgolubi1:
Okay, so this now:
* takes you to the overview not in monitor mode when you don't have either preference set
* takes you to the overview in monitor mode when you have the automonitor preference set (regardless of whether you were monitoring *before* dismissing the game, but that's probably fine, and certainly worth putting in front of people who actually use the automonitor preference)
* if you have automonitor and also "monitor takes you to next game" set, takes you to a game requiring action when you dismiss a game.

So that seems okay.

PR #1624, opened 2015-05-23
Hide stale games

### LINKED ISSUE #846: Let players delete old games
If an opponent takes too long to take their turn (I think on the old site it was 14 days), players should be allowed to end the game and take a win by default.

I guess this would probably mean adding an action log item saying that this happened and setting the game state to Completed, since that would make it clear to anyone looking at the game what happened (as well as making sure that it showed up in both players' completed games lists).

Hmm. What would be the consequences of having a game that was completed, with a winner, but without a matching round W/L/T score?

blackshadowshade:
Fixes #846.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/732/

### DIFF
--- a/src/ui/js/Overview.js
+++ b/src/ui/js/Overview.js
@@ -360,10 +360,12 @@ Overview.pageAddGameTable = function(
     tableFoot.append(footRow);
     tableFoot.click(function() {
       $('.staleGame').toggle();
-      $('.staleGame').is(":visible") ?
-        footCol.text('Hide stale games') :
-        footCol.text('Show stale games');
-    })
+      footCol.text(
+        $('.staleGame').is(':visible') ?
+        'Hide stale games' :
+        'Show stale games'
+      );
+    });
     tableBody.closest('table').append(tableFoot);
   }
 };
--- a/src/ui/js/Overview.js
+++ b/src/ui/js/Overview.js
@@ -359,7 +359,7 @@ Overview.pageAddGameTable = function(
     var staleToggle = $('<a>', {
       'id': 'staleToggle',
       'href': 'javascript:Overview.toggleStaleGame();',
-      'text': 'Show stale games',
+      'text': '[Show stale games]',
     });
 
     footCol.append(staleToggle);
@@ -373,8 +373,8 @@ Overview.toggleStaleGame = function() {
   $('.staleGame').toggle();
   $('#staleToggle').text(
     $('.staleGame').is(':visible') ?
-    'Hide stale games' :
-    'Show stale games'
+    '[Hide stale games]' :
+    '[Show stale games]'
   );
 };
 
--- a/src/ui/js/Overview.js
+++ b/src/ui/js/Overview.js
@@ -359,10 +359,10 @@ Overview.pageAddGameTable = function(
     var staleToggle = $('<a>', {
       'id': 'staleToggle',
       'href': 'javascript:Overview.toggleStaleGame();',
-      'text': '[Show stale games]',
+      'text': 'Show stale games',
     });
 
-    footCol.append(staleToggle);
+    footCol.append('[').append(staleToggle).append(']');
     footRow.append(footCol);
     tableFoot.append(footRow);
     tableBody.closest('table').append(tableFoot);
@@ -373,8 +373,8 @@ Overview.toggleStaleGame = function() {
   $('.staleGame').toggle();
   $('#staleToggle').text(
     $('.staleGame').is(':visible') ?
-    '[Hide stale games]' :
-    '[Show stale games]'
+    'Hide stale games' :
+    'Show stale games'
   );
 };
 
--- a/src/ui/js/Overview.js
+++ b/src/ui/js/Overview.js
@@ -363,7 +363,7 @@ Overview.pageAddGameTable = function(
     });
 
     footCol.append(staleToggle);
-    footRow.append(footCol)
+    footRow.append(footCol);
     tableFoot.append(footRow);
     tableBody.closest('table').append(tableFoot);
   }
@@ -376,7 +376,7 @@ Overview.toggleStaleGame = function() {
     'Hide stale games' :
     'Show stale games'
   );
-}
+};
 
 Overview.pageAddIntroText = function() {
   Overview.page.append($('<h1>', {'text': 'Welcome to Button Men!', }));
--- a/src/ui/js/Overview.js
+++ b/src/ui/js/Overview.js
@@ -260,6 +260,7 @@ Overview.pageAddGameTable = function(
   }
 
   var i = 0;
+  var staleGamesExist = false;
   while (i < gamesource.length) {
     var gameInfo = gamesource[i];
     var playerColor = gameInfo.playerColor;
@@ -273,6 +274,12 @@ Overview.pageAddGameTable = function(
       gameLinkTd.append($('<a>', {'href': 'game.html?game=' + gameInfo.gameId,
                                   'text': 'Play Game ' + gameInfo.gameId,}));
     } else if (gameType == 'awaitingOpponent') {
+      if (gameInfo.inactivityRaw > 14*24*60*60) {
+        staleGamesExist = true;
+        gameRow.addClass('staleGame');
+        gameRow.hide();
+      }
+
       gameLinkTd =
         $('<td>', { 'style': 'background-color: ' + opponentColor, });
       gameLinkTd.append($('<a>', {'href': 'game.html?game=' + gameInfo.gameId,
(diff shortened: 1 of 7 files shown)


cgolubi1:
Okay, what's going on here?

There certainly shouldn't be any regressions to gameplay --- i'm playing back a batch of games to test that theory, but i don't expect any surprises there.

There's not nearly as much stuff in place for testing regression to the overview --- i'm basically just going to have to eyeball it, which pretty much means:
* See what the overview page looks like when there are no games at all
* See what the overview page looks like when there are some games, but none which are stale
* See what the overview page looks like when there are stale games
* Look at the code, make sure i understand what it's doing, and see if i have any questions about it

cgolubi1:
So... i have two issues with the visuals here --- first, you have to click on something which is not highlighted to look like a link or a button:
show_stale_games

Second, once you've done that, the "hide stale games" link drops to the bottom, so the visual effect is that the bar you just clicked on disappears, and you've got all these stale games.

I think the "Show stale games" link needs to look clickable in some way --- i'm not really picky about how.  I'm weak on web design, but i *think* the issue is that the cursor doesn't change appearance when you mouse over it.  If the pointer turned into a hand, that would be a strong hint that you're now on an interactive element.

I also think it would be better if "Hide stale games" appeared in the same place where "Show stale games" was, and the stale games table were drawn below it.  That would give a better illusion of a button that i was clicking to open a particular view, and then could click the same button to close that view.

src/ui/js/Overview.js
```diff
@@ -273,6 +274,12 @@ Overview.pageAddGameTable = function(
       gameLinkTd.append($('<a>', {'href': 'game.html?game=' + gameInfo.gameId,
                                   'text': 'Play Game ' + gameInfo.gameId,}));
     } else if (gameType == 'awaitingOpponent') {
+      if (gameInfo.inactivityRaw > 14*24*60*60) {
```
cgolubi1: Make this a variable at the top of the file --- STALENESS_SECS = 14 * 24 * 60 * 60 or something.  It's a little icky to have a bunch of multiplication inline.  Or if you must do the multiplication inline, at least make STALENESS_DAYS a variable.

irilyth:
I haven't looked at this, but one idea is to mimic the look of the view/hide things that we already have. Chat log (which I guess is just "view", no "hide"); any others?

For chat log, it's clearly a link (it's colored and underlined like a link); and if there were going to be a "hide" link, I'd think that you'd draw it at the bottom of the log after you expand it.

src/ui/js/Overview.js
```diff
@@ -340,6 +347,27 @@ Overview.pageAddGameTable = function(
...
     tableBody.append(gameRow);
   }
+
+  if (staleGamesExist) {
+    var tableFoot = $('<tfoot>');
+    var footRow = $('<tr>');
+    var footCol = $('<td>');
+    // james: Do not use colspan="0" because this is incompatible with older
+    //        browsers. Instead, use overkill and specify 100 columns.
+    footCol.attr('colspan', 100);
```
cgolubi1: I'm suspicious of this --- every other use of colspan in the JS is a number under 10, usually 2, 4, or 6, and is specified as a string (that last one is a nit).  Why is this one 100?

What is the visual effect you are trying to accomplish by setting colspan?  The comment doesn't say at all, and it's not obvious to me.

cgolubi1:
Hmm, okay, i may be backing off on that second idea --- basically, there are four different cases for the visuals:
1. All the non-stale games i have are waiting for me, and stale games are hidden
2. All the non-stale games i have are waiting for me, and stale games are visible
3. Some non-stale games are waiting for my opponent, and stale games are hidden
4. Some non-stale games are waiting for my opponent, and stale games are visible

Let me screencap those four cases, using my dev games as the sample population.

cgolubi1:
Okay, screencaps - here's the transition between "games waiting for me" and whatever's below that, in each of the four cases:

Case 1:
stale_games_case_1

Case 2:
stale_games_case_2

Case 3:
stale_games_case_3

Case 4:
stale_games_case_4

And in case 2 and case 4, the transitions between "games waiting for opponent" and the "Hide stale games" bar are also relevant, so here are those:

Case 2:
stale_games_case_2_bottom

Case 4:
stale_games_case_4_bottom

cgolubi1:
So... it's a little ugly how all of those cases look different from each other.  I think the most glaring is that in case 3, the bottom of the table is broken.  That's true in both firefox and safari for me, and given that i think case 3 is probably the most common case, IMO it's worth fixing that bottom line of that table.

cgolubi1:
However, given all of those options, i'm not as exercised by the fact that the "Hide stale games" bar moves, especially since simply reloading the page will hide the stale games, as people are used to from the in-game view.

I do agree with irilyth's suggestion --- instead of drawing a bar for "Show stale games", why not simply draw a centered link which looks like the "[View full activity and chat history]" one inside the game?  That would make our UI consistent, and i bet the table bottom bug in case 3 would go away if you weren't trying to draw a big column in the existing table.

cgolubi1:
Okay, summarizing the previous N comments:
* The "Show stale games" link is implemented as a non-clickable-looking table row, which is bad UI design (because there's nothing to tell the user it's clickable) and looks inconsistent with other show/hide interfaces we have - suggestion is to make it a link like the action/chat log one.
* The visual effect of clicking on a bar and having it disappear (move to the end of the page) is jarring.  I think making it a link would make it less jarring.
* When the player has games waiting for an opponent, and also stale games, the table of games waiting for the opponent is missing its bottom bar.  Again, i'm guessing making the button a link would solve this bug.
* Number of days for staleness is hidden in an arithmetic expression in Overview.js - it should be a top-of-file variable.
* I want better documentation of why you're setting colspan to a weirdly big value, though i think i eventually figured it out.  Again, if you're not putting the bar in the table, this one goes away.

blackshadowshade:
Okay, that all seems reasonable. This shouldn't be too difficult to fix.

blackshadowshade:
So, I've addressed all issues apart from the missing bar. I presume that this has something to do with the .hide() function, but I haven't found it yet. I think that this is ready enough to go, and I'll see if I can track down the missing bar when I can.

blackshadowshade:
BTW, if anyone has any tips on the missing bar, I'd appreciate them. :)

I've added an empty test for Overview.toggleStaleGame, and I'd like some help in filling this out.

blackshadowshade:
This could be related:
http://stackoverflow.com/questions/1999311/hiding-trs-borders-being-included

blackshadowshade:
Oddly, when I test this simply, the border remains visible:
https://jsfiddle.net/gycu3vk9/

blackshadowshade:
Right, I think I've fixed this. It was all to do with Jota's very clever CSS evading my attention.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/758/

cgolubi1:
Yep, the bottom line works for me too now.

cgolubi1:
I notice this change made the borders around the header bar (the grey bar containing "Game #" and so forth) substantially narrower --- was that on purpose?  I don't think i care, just curious if it was intentional.

cgolubi1:
Slightly more relevantly: the visual for "[View full activity and chat history]" is:
* a blue link, not a black one
* surrounded by left and right square brackets, "[" and "]"

My initial suggestion had been that we match that style, to make page-expanding elements look consistent.  I don't insist, but i do think consistent would be slightly better.

blackshadowshade:
The change of the borders around the header bar was not intentional, but also not unintentional. I noticed it, and thought that it wouldn't matter much. Maintaining the border size would have added additional complexity to the CSS, which I thought would be unnecessary.

I'm happy to add the square brackets to the link. Regarding the link colour, I have no idea where that is encoded. I'll take a look.

cgolubi1:
> The change of the borders around the header bar was not intentional, but also not unintentional. I noticed it, and thought that it wouldn't matter much. Maintaining the border size would have added additional complexity to the CSS, which I thought would be unnecessary.

Sure, that's fine --- it's useful to me in testing if you mention the visual changes you're aware of in the pull request.  (But if not i can always just ask if i notice them.)

cgolubi1:
As for the color, i'm guessing it's in gui.css:
```
table.gameList a {
    color: black;
}
```

blackshadowshade:
Okay, this should be fixed now.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/764/ (if it passes)

blackshadowshade:
Oh wait, the brackets are blue now too. XD

cgolubi1:
Yeah, the code that i pointed to makes the links black for ALL entries in the overview table and ALL history page entries.  You don't want to simply get rid of it --- you probably don't want to change it at all.  You want to add a special case for tfoot or some other thing that only covers the stale games link.

cgolubi1:
If you don't want the brackets to be blue, put them outside of the <a href> rather than inside it, and that'll take care of that.  That's separate from my comment above, which is that you don't want to cause regressions for the entire Overview and History UIs --- those links should presumably stay black.

blackshadowshade:
Right, try this.

cgolubi1:
Yes. That is what i am talking about.  :>)  Ready to merge.

cgolubi1:
Oh, actually, let's wait to make sure jenkins passes first.  But then it'll be ready.

PR #1629, opened 2015-05-27
Exclude ApiResponder and DummyApiResponder from PHPDoc targets

blackshadowshade:
Addresses part of #1130.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/735/

### DIFF
--- a/deploy/jenkins/build.xml
+++ b/deploy/jenkins/build.xml
@@ -107,7 +107,7 @@
 
  <target name="phpdoc" description="Generate API documentation using PHPDocumentor">
   <exec executable="phpdoc">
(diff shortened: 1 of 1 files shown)


cgolubi1:
Hmm, i definitely agree that we should exclude DummyApiResponder.php, because that's test code which has to be in the non-test tree for implementation reasons.

But why exclude ApiResponder.php - that's part of the user-visible code path, and we should document it IMO.

blackshadowshade:
Okay, I'll change the exclusion to be only DummyApiResponder.php

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/747/

cgolubi1:
Thanks.

PR #1630, opened 2015-05-27
Only show link to show all logs when some logs are hidden

### LINKED ISSUE #784: when all log entries are already being shown, suppress option to show more

blackshadowshade:
Fixes #784.

This completes the work started in #1222 by Jota.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/736/

### DIFF
--- a/src/engine/BMInterface.php
+++ b/src/engine/BMInterface.php
@@ -668,8 +668,9 @@ class BMInterface {
                 $data['playerDataArray'][$gamePlayerIdx]['playerName'] = $playerName;
             }
 
-            $data['gameActionLog'] = $this->load_game_action_log($game, $logEntryLimit);
-            $data['gameActionLogCount'] = $this->count_game_action_log($game);
+            $actionLogArray = $this->load_game_action_log($game, $logEntryLimit);
+            $data['gameActionLog'] = $actionLogArray['logEntries'];
+            $data['gameActionLogCount'] = $actionLogArray['nEntries'];
             $data['gameChatLog'] = $this->load_game_chat_log($game, $logEntryLimit);
             $data['gameChatLogCount'] = $this->count_game_chat_log($game);
             $data['timestamp'] = $this->timestamp;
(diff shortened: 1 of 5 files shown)


cgolubi1:
From the code, i think this isn't going to work right --- as discussed at length in #1540 and addressed correctly when *returning* action log entries in #1571, some entries in the action log database are invisible and are not returned to the player.  So you can't just count them.

Since `load_game_action_log()` already loads all the entries, and thus already knows how many there are, why not just have it report a compound value containing the requested subset of entries, and also an integer saying how many total entries there are?  Why add a new function with a second select when you already have the data you need?

cgolubi1:
Also... am i just completely misunderstanding how this solution works?  You added new elements to the loadApiGameData response, so why didn't you have to modify all the responder tests?  I mean, it appears that you didn't, but i'm mystified as to why not.

cgolubi1:
Oh!  I see.  You did change responderTest, you just made so many changes it's not willing to show them to me.  Yeah, okay, that makes more sense.

blackshadowshade:
Okay, I'll do an iteration on this when I can.

cgolubi1:
This is still on shadowshade's plate to iterate on getting the necessary counts from `load_game_action_log()` and `load_game_chat_log()`, rather than needing to reproduce all the code to generate them correctly.

blackshadowshade:
Right, I think I've made the required changes.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/770/ (if it passes)

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/772/ (if it passes)

cgolubi1:
I think this looks reasonable now, and it seems to work.  I'm going to go ahead and make sure RandomAI can play games against the new API, then call this good.

cgolubi1:
Yep, RandomAI seems fine here.

PR #1635, opened 2015-06-06
Weak/mighty trip dice now only roll once when performing a trip attack

### LINKED ISSUE #1628: Weak Trip die shrinks too many times, and isn't logged
In http://dev.buttonweavers.com/ui/game.html?game=1673, I did a Trip Attack with a th(8) against another Weak die. The log said:

> irilyth performed Trip attack using [th(8):6] against [hpn(10):10]; Attacker th(8) rerolled 6 => 4; Defender hpn(10) recipe changed to hpn(8), rerolled 10 => 1, was captured

The defending die shrank and re-rolled once and was captured, as expected. The attacking die, however, seems to have shrunk *twice*, and is showing a 2 rather than a 4 after the attack, and neither time is mentioned in the log. Here's a screenshot:

bm

I think the th(8) should have shrunk to a th(6) before re-rolling, and then stayed a th(6) and retained its current value afterwards.

blackshadowshade:
Fixes #1628.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/740/

### DIFF
--- a/src/engine/BMSkillMighty.php
+++ b/src/engine/BMSkillMighty.php
@@ -37,6 +37,11 @@ class BMSkillMighty extends BMSkill {
             return;
         }
 
+        // don't trigger skill when the die has just performed a trip attack
+        if ($die->has_flag('JustPerformedTripAttack')) {
+            return;
+        }
+
         $die->grow();
     }
 
--- a/src/engine/BMSkillWeak.php
+++ b/src/engine/BMSkillWeak.php
@@ -37,6 +37,11 @@ class BMSkillWeak extends BMSkill {
             return;
         }
 
+        // don't trigger skill when the die has just performed a trip attack
+        if ($die->has_flag('JustPerformedTripAttack')) {
+            return;
+        }
+
         $die->shrink();
     }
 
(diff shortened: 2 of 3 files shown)


cgolubi1:
Okay, what am i going to do here?
* Identify a game using which i can see the new behavior in action
* Regression test a bunch of games and make sure all changes are expected
* Autoplay some new games and look for internal errors

cgolubi1:
Okay, so, Darthcliff and yakboy are the only non-random buttons with Ht or ht, and neither is implemented.  I'll start with those.

cgolubi1:
I played back 500 games, and the only regressions were of the form:
```
Failed asserting that 1 matches expected 0.
```

and were in RandomBMMixed games with either mighty trip or weak trip dice.  That's a good sign.

cgolubi1:
Hmm... i looked at one of the games more carefully --- it'll take me a little while to reformat it as a responder test, but probably not too long.  Anyway, the issue is that the behavior is right, but the action log doesn't mention the recipe change at all now:
recipe_change_not_mentioned

cgolubi1:
Shadowshade said he'd rather address the action log bug as a separate issue, so i've opened a separate issue for it, and am now merging this.

PR #1636, opened 2015-06-10
Added logic to handle logging of complex size changing that starts with berserk

### LINKED ISSUE #1620: Add a flag for "this die shrank due to Berserk"
Because we do all the logging after the attack is complete, and thus don't know what happened during the attack unless something flags it, we can't tell if a die shrank due to Berserk, and thus don't log this, which gets confusing if the die also changes size for some other reason (e.g. Mood Swing).

blackshadowshade:
Fixes #1620.

This should probably have a responder test added eventually. I tested using spindisc vs spindisc.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/743/

### DIFF
--- a/src/engine/BMDie.php
+++ b/src/engine/BMDie.php
@@ -791,19 +791,28 @@ class BMDie extends BMCanHaveSkill {
         }
 
         if ($this->has_flag('JustPerformedTripAttack')) {
-            $actionLogInfo['valueAfterTripAttack'] = $this->flagList['JustPerformedTripAttack']->value();
+            $actionLogInfo['valueAfterTripAttack'] =
+                $this->flagList['JustPerformedTripAttack']->value();
+        }
+
+        if ($this->has_flag('JustPerformedBerserkAttack')) {
+            $actionLogInfo['recipeAfterBerserkAttack'] =
+                $this->flagList['JustPerformedBerserkAttack']->value();
         }
 
         if ($this->has_flag('HasJustGrown')) {
-            $actionLogInfo['recipeBeforeGrowing'] = $this->flagList['HasJustGrown']->value();
+            $actionLogInfo['recipeBeforeGrowing'] =
+                $this->flagList['HasJustGrown']->value();
         }
 
         if ($this->has_flag('HasJustShrunk')) {
-            $actionLogInfo['recipeBeforeShrinking'] = $this->flagList['HasJustShrunk']->value();
+            $actionLogInfo['recipeBeforeShrinking'] =
+                $this->flagList['HasJustShrunk']->value();
         }
 
         if ($this->has_flag('HasJustSplit')) {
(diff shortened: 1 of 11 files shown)


cgolubi1:
This is complicated.  Deferring for now in hopes of trying to knock off some simpler pulls.

cgolubi1:
Okay, so, basically, what James did here is:
* Add a flag JustPerformedBerserkAttack which is associated with an action log entry type recipeAfterBerserkAttack
* Make a new parent class BMFlagJustPerformedTypeOfAttack and refactor BMFlagJustPerformedTripAttack to use it
* In BMGameAction, implement additional messaging using the Berserk information

So, what i want to do here is:
* Play a game by hand using spindisc as James suggests, and see if i like the result
* Play back a lot of games containing no Berserk dice, and make sure there are no changes to visible logging in those games
* Play back a bunch of games containing Berserk dice, and see what the differences are and make sure they are reasonable
* Play some new games containing size-changing skills, and look for internal errors
* See if i can easily make a short responderTest for this

cgolubi1:
The berserk mood behavior looks reasonable:
berserk_mood_behavior

cgolubi1:
Hmm, this doesn't look so good --- why is the intermediate value 0 sides?
berserk_mighty

cgolubi1:
My assumption is that it's because of your new `max_from_recipe()` function - i don't really understand how that works.  My request, orthogonal to fixing the bug above, would be add some unit tests to `max_from_recipe()` so its behavior can be seen.

cgolubi1:
Okay, i added a responder test in https://github.com/cgolubi1/buttonmen/tree/demonstrate_berserk_logging_bug which demonstrates the incorrect behavior here (and also, basically by chance, has a boom attack targeting a stealth die).

blackshadowshade:
Yes, I'm pretty sure you're exactly right, in that the regexp is actively searching for an equals sign, which will fail for a non-swing die.

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/753/ (if it passes)

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/756/ (if it passes)

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/757/

cgolubi1:
Okay, trying this again.

* Play a game by hand using spindisc, and make sure i still like the result
* Play a game by hand using RandomBMMixed, trying to get a berserk size-changing non-swing die
* Play back a lot of games containing no Berserk dice, and make sure there are no changes to visible logging in those games
* Play back a bunch of games containing Berserk dice, and see what the differences are and make sure they are reasonable
* Play some new games containing size-changing skills, and look for internal errors

cgolubi1:
This seems good to the extent of my testing ability, but it's new logic and i want to give the testers a crack at it, so i'll merge it after the code freeze.

PR #1637, opened 2015-06-19
Added logic to disable "Mark thread as read" button when there is text in the reply box

### LINKED ISSUE #1336: Prompt for confirmation before leaving a forum page with unsaved changes
I just lost a long forum post when I accidentally clicked on the button immediately below my post, which is "Mark thread as read", rather than the button way off on the right, "Post reply". In general if the user tries to do something that will navigate them away from an unfinished forum post, it'd be better to prompt for confirmation than to instantly discard it.

blackshadowshade:
Fixes #1336.

I'm thinking that there should probably be a QUnit test for this, but I don't yet know how to construct one.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/745/ (if it passes)

### DIFF
--- a/src/ui/js/Forum.js
+++ b/src/ui/js/Forum.js
@@ -323,7 +323,7 @@ Forum.showThread = function() {
     'maxlength': Forum.FORUM_BODY_MAX_LENGTH,
   });
   replyBodyTextArea.on('change keyup paste', function() {
-    if ('' == $(this).val().trim()) {
+    if ('' === $(this).val().trim()) {
       if ('disabled' == $('#markReadButton').attr('disabled')) {
         $('#markReadButton').removeAttr('disabled');
         $('#markReadButton').removeAttr('title');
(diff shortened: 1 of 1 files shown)


cgolubi1:
Okay, let me try this out.  There probably should be a QUnit test, but my recollection is that mocking this kind of dynamic behavior isn't 100% trivial, so i'm inclined to not hold up the pull for that if i like the visible behavior.

cgolubi1:
Yeah, this seems to work.  My nit-level comment is that i'm not sure trim() is correct --- if i setup my complicated whitespace pattern, why should i be able to accidentally delete it?  So if it were me, i wouldn't have used trim().

But i don't actually think that'll affect any real users any time soon, so i'm just going to throw the comment out there, and approve this.

blackshadowshade:
My reason for the trim is the fact that I will sometimes add extra spaces accidentally because I use the space as a PageDown key in my browser. Also, it's almost impossible to add tabs to a reply box like this because it causes the focus to change by default.

cgolubi1:
Makes sense.  I'm on the fence about what the right answer is, and i think it's not important enough to worry about.  Either way this looks like a big win over the current behavior.

cgolubi1:
One question i do want to get on the record is, should the "Mark board as read" box behave the same way when you're creating a new thread?  Again, a lesser case that affects fewer people, but might be more likely to bite people if they get used to the UI protecting them in the "comment thread" case.

blackshadowshade:
Sounds like a good question for the testers.

PR #1639, opened 2015-06-21
Update responderTest list of skills to handle berserk/speed interaction

cgolubi1:
I noticed some bitrot between the live implementation of skill interaction descriptions, and the list responderTest uses to regression-test those interactions.  Here's a quick fix for responderTest to match the live implementation.

http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/411/ (assuming it passes)


blackshadowshade:
Looks fine.

PR #1641, opened 2015-06-22
Add responder test for Boom dice

cgolubi1:
Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/412/ (assuming it passes)


blackshadowshade:
I note some interesting syntactic mismatches with the position of the '&' on the Mad Swing dice, but that can wait for a different issue.

blackshadowshade:
Okay, I think the issue here is that the recipes in the database should really have the '&' after the brackets, to be consistent with Mood Swing. Otherwise, the test looks fine.

PR #1647, opened 2015-06-23
Made API always return outOfPlayDieArray

### LINKED ISSUE #1640: API inconsistency between activeDieArray, capturedDieArray, and outOfPlayDieArray
When there are no out of play dice, outOfPlayDieArray does not appear in the API result.  However, when there are no active or captured dice, activeDieArray and capturedDieArray still appear, just with size 0.  These three arrays should be treated consistently.

blackshadowshade:
Fixes #1640.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/751/

### DIFF
--- a/src/engine/BMGame.php
+++ b/src/engine/BMGame.php
@@ -3282,13 +3282,12 @@ class BMGame {
         $canStillWinArray = $this->get_canStillWinArray();
 
         foreach ($this->playerIdArray as $playerIdx => $playerId) {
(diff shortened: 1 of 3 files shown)


cgolubi1:
Looks extremely straightforward.  I'm sufficiently convinced that weird regressions are unlikely that i'm not interested in figuring out how to regression-test the expected across the board changes.  So i'll just throw a bunch of new RandomAI games at this to make sure nothing blows up weirdly, then call it good.

cgolubi1:
Seems fine.

PR #1648, opened 2015-06-24
Boom attacks now do not trigger side-effects from other skills

### LINKED ISSUE #1645: Boom vs Radioactive
In http://dev.buttonweavers.com/ui/game.html?game=1712, I did

> irilyth performed Boom attack using [b(4):4] against [%Ho(1,8):6]; Defender %Ho(1,8) recipe changed to %Ho(2,10), rerolled 6 => 3, was not captured; Attacker b(4) showing 4 split into: b(2) showing 1, and b(2) showing 1

That seems pretty correct, as both halves of my Boom die vanished, but only one of them is showing up on the out-of-play line; probably both should.

blackshadowshade:
Fixes #1645.

This addresses the interactions between boom and the following skills: 
- morphing
- null
- radioactive
- value
- mighty
- weak
- mood
- mad
- ornery

This will need quite a bit of testing by the testers, I expect.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/762/

### DIFF
--- a/src/engine/BMSkill.php
+++ b/src/engine/BMSkill.php
@@ -221,6 +221,7 @@ class BMSkill {
                      'BMSkillReserve',
                      'BMSkillChance',
                      'BMSkillFocus',
+                     'BMSkillBoom',
                      'BMSkillSpeed',
                      'BMSkillTrip',
                      'BMSkillQueer',
@@ -244,8 +245,7 @@ class BMSkill {
                      'BMSkillNull',
                      'BMSkillMaximum',
                      'BMSkillTimeAndSpace',
-                     'BMSkillWarrior',
-                     'BMSkillBoom');
+                     'BMSkillWarrior');
         // fires last
     }
 
--- a/src/engine/BMSkillMighty.php
+++ b/src/engine/BMSkillMighty.php
@@ -42,6 +42,11 @@ class BMSkillMighty extends BMSkill {
             return;
         }
 
+        // don't trigger skil if the die has already left play
+        if ($die->outOfPlay) {
+            return;
+        }
+
         $die->grow();
     }
 
--- a/src/engine/BMSkillMighty.php
+++ b/src/engine/BMSkillMighty.php
@@ -42,7 +42,7 @@ class BMSkillMighty extends BMSkill {
             return;
         }
 
-        // don't trigger skil if the die has already left play
+        // don't trigger skill if the die has already left play
         if ($die->outOfPlay) {
             return;
         }
--- a/src/engine/BMSkillMood.php
+++ b/src/engine/BMSkillMood.php
@@ -40,6 +40,10 @@ class BMSkillMood extends BMSkill {
             return FALSE;
         }
 
+        if ($args['die']->outOfPlay) {
+            return FALSE;
+        }
+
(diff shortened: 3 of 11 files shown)


cgolubi1:
Okay.  Sorry i fell off the wagon about this --- we need it to push Boom to prod.

Basically, this change amounts to modifying the order in which die skills fire, and then changing several other skills:
# For Mighty, Mood, Morphing, Ornery, Radioactive, Weak: bail out of applying that skill if the die is out of play
# For Null, Value: only apply the skill if the defender has been captured

So, the theory is that these changes should only affect Boom.  I don't see any reason that would fail in the first case (though we should play back some games just in case).  In the second case, what if `$defender->captured` is somehow not set for some other reason?  But, again, the only fix i can think of is to play back games.

So my plan here is:
* Play back a lot of games that don't contain Boom.  There should be no regressions, but i want to try a couple thousand games here (which i have in hand right now, thanks to procrastinating on this pull.  `:>P`)
* Play back a bunch of games that do contain Boom.  Most of them should have some sort of change, see what they look like.
* Play a fair number of new games, looking for internal errors, including all of:
** Miscellaneous games with no restriction
** Games containing known Boom buttons
** RandomBMMixed games
* Look at some of those Boom games by hand, and make sure the behavior looks right

cgolubi1:
My first set of non-Boom tests had one "regression":
```
1) responderTest::test_interface_game_00050
--- Expected
+++ Actual
@@ @@
-'responder004 performed Trip attack using [tv(10):7] against [ft(12):2]; Attacker tv(10) rerolled 7 => 5; Defender ft(12) recipe changed to ftv(12), rerolled 2 => 6, was not captured. '
+'responder004 performed Trip attack using [tv(10):7] against [ft(12):2]; Attacker tv(10) rerolled 7 => 5; Defender ft(12) rerolled 2 => 6, was not captured. '
```

That seems obviously like a fix of an (unreported?) related bug, not a regression.  So, not a problem.

cgolubi1:
Incidentally, i don't see any evidence that anyone previously reported that bug (that when you unsuccessfully trip attack using a value trip die, the target gains "value"), and that's pretty surprising to me.  Maybe we just didn't have any value or null trip dice until RandomBMMixed gained trip?  I'm not going to sidetrack this pull request to follow up, but we should almost certainly have a regression test for this case, to make sure we don't lose this property.

cgolubi1:
Second round of random games seemed fine --- there's an interaction between Radioactive and Morphing in the code which apparently responderTest doesn't know about (implying that we have no responder tests around the interaction of radioactive and morphing, but that's not today's problem), but otherwise i just saw a second bugfix for the "trip should not add value" unreported bug.

cgolubi1:
Okay, no surprises in the third and fourth rounds of random games, so let's call this good from that standpoint.

cgolubi1:
Regression-related changes seem basically as expected.

cgolubi1:
Found a Violette vs. Aysez game in which a boom die targeted a radioactive die and the right thing happened (no split, target stayed radioactive, attacker was taken out of game).

cgolubi1:
Okay, i haven't managed to break this yet, so let's merge it.

PR #1646, opened 2015-06-22
Fixed recipes with Mad Swing to use post-fix notation

blackshadowshade:
Fixes #1641.

Requires database update 01643_mad_order.sql

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/748/

Issue #217, opened 2013-11-27
push new features out to dev.buttonweavers.com


cgolubi1:
As features are merged into the main codebase, redeploy dev.buttonweavers.com with those changes.

Generic procedures for use in deployments:

Use a checked-out copy of buttonmen-master to find a sort order for database updates:
```bash
cd deploy/database/updates/
UPDATES="<space-separated list of DB updates>"
git log --reverse --first-parent --oneline $UPDATES
UPDATES="<space-separated list of DB updates, now in the order reported by that output>"
```

Database update plan:
```bash
# in the correct git working directory
git pull origin dev
cd deploy/vagrant
vagrant provision

# on the sandbox or dev VM
UPDATES="<space-separated ordered list of DB updates, as determined above>"

mysqldump -u root buttonmen | sed -e 's/),(/),\n(/g' > ~/buttonmen_db.$(timefmt).sql
cd /buttonmen/deploy/database/updates
for update in $UPDATES; do
  echo "APPLYING: $update"
  cat $update | mysql -u root buttonmen
done

mysqldump -u root buttonmen | sed -e 's/),(/),\n(/g' > ~/buttonmen_db.$(timefmt).sql
```

How to diff database dumps:
```bash
diff --side-by-side --width=234 --left-column buttonmen_db.* | less
```

Within less, search for lines that don't end with `(` to skip past common lines: `/!\($`

cgolubi1:
This issue is getting slow to load because it links to so many other issues, and it would be more useful from my POV to have a dev push issue corresponding to each prod release, so that when i go to do the prod release i have the exact set of corresponding dev releases in one place.  So i'm going to close this out and switch to a model where i cycle the dev push ticket monthly.

cgolubi1:
The new issue is #1915

PR #1649, opened 2015-06-28
pull changes from master to dev on 2015-06-28

cgolubi1:
* User-visible:
    * #1648: Boom attacks no longer trigger side-effects from other skills.  Also fixes a bug in which unsuccessful Trip attacks used to confer Value or Null.
    * #1646: recipes containing Mad swing dice should have the `&` after the die size, not before
* Ideally not user-visible:
    * #1641: test for boom dice
    * #1647: API always returns `outOfPlayDieArray`, even when it's empty

Database updates:
```
01643_mad_order.sql
```

Once this pull is merged, i will load the changes on dev.buttonweavers.com as part of #217.


cgolubi1:
Testing on vagrant.

cgolubi1:
The only DB difference is the button recipes.  The button recipe changes take effect immediately, including to open games, and don't appear to cause any ill effects.  Loading on dev now.

PR #1676, opened 2015-08-07
Added interaction between Doppelganger and Rage

### LINKED ISSUE #1674: Doppelganger vs Rage
When a Doppelganger die captures a Rage die, it gains Rage, and doesn't lose it. This seems relatively correct, since Rage says "If a Rage Die participates in an Attack, it loses its Rage ability." while Doppelganger says "When a Doppelganger Die performs a Power Attack on another die, the Doppelganger Die becomes an exact copy of the die it captured." so it seems to me that the attacking die doesn't have Rage at the time it participates in the attack, but only gains Rage after the attack, and thus wouldn't lose it then.

If that seems wrong to others, then there's a bug.

If that seems right to others, then it'd be nice if the Doppelganger and Rage mentioned this as an "interaction with other skills" thing, so that people would know for sure what to expect before making the attack.

blackshadowshade:
Fixes #1674.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/803/ (if it passes)

### DIFF
--- a/src/engine/BMSkillDoppelganger.php
+++ b/src/engine/BMSkillDoppelganger.php
@@ -75,6 +75,8 @@ class BMSkillDoppelganger extends BMSkillMorphing {
             'Radioactive' => 'Dice with both Radioactive and Doppelganger first decay, then ' .
                              'each of the "decay products" are replaced by exact copies of the ' .
                              'die they captured',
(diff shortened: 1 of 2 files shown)


cgolubi1:
Yep.

PR #1677, opened 2015-08-08
Added skill for The Flying Squirrel, and added description for button special skills

### LINKED ISSUE #369: need Flying Squirrel skill
This is more of a lack of skill than a skill.  I'm not sure if it should be made into a die type or just be kept as a button special.  My notes were made assuming it would be a die type:
(Flying) Squirrel: Q
The Flying Squirrel can not make skill atttacks.
Used to implement special button rules for The Flying Squirrel from The Japanese Beetle set

### LINKED ISSUE #1200: Button specials need descriptions
Some button specials have been implemented already. Their ```BMBtnSkill``` classes should have a static ```get_description()``` method on them that returns the description of that special as a string. (And as new ones are implemented, they should be given descriptions as well.)

blackshadowshade:
Fixes #369. Partly addresses #1200.

Requires database update 0369_btn_skill_descriptions

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/812/ (if it passes)


cgolubi1:
Okay, so, as i understand this, the button special skills description changes are only instantiating the text in the `get_description()` functions --- it's not yet doing anything with the text like passing it via the API.  I should double-check that assumption.  In general:
* Look at the code and see if anything catches my eye as needing checking
* Play some TFS games by hand
* Play back a bunch of RandomAI games --- there should be no regressions
* Play a bunch of new RandomAI games --- there should be no internal errors
* Play new RandomAI games with TFS --- there should be no internal errors

blackshadowshade:
The descriptions actually do come up under the button description pages.

src/engine/BMBtnSkillRandomBMFixed.php
```diff
@@ -44,4 +44,14 @@ public static function specify_recipes(array $args) {
...
         return TRUE;
     }
+
+    /**
+     * Description of skill
+     *
+     * @return string
+     */
+    public static function get_description() {
+        return '5 dice, no swing dice, two of them having a single skill' .
```
cgolubi1: You need a space after "skill" here.

cgolubi1:
This is untidy - because we had been using the database entries to describe each RandomBM type, there's a mismatch here:
random_bm_mixed_description

First, the "Special" description does not actually say what skills are chosen among, so that's wrong (and is new).

Second, the database description hasn't been kept up-to-date as we've added skills, so it's wrong (and that's not new).

Seems like the way to go is:
* Change the database description to be something way simpler - "RandomBM with a mix of skills" or something
* In the Special description, actually do populate the list of included skills, but do it live using the actual set of skills in the code so it stays accurate.

cgolubi1:
Okay, the TFS parts of this look fine so far.  So basically that just leaves:
* The one-character typo in the skill description for RandomBMFixed
* The untidiness of the discrepancy between DB entry and special for RandomBMMixed

If the latter doesn't seem trivial to fix, i'm happy to see it fixed as a separate issue/pull.  If it is trivial, let's do it here.

blackshadowshade:
I've simplified all the descriptions of BMBtnSkillRandomBM*, since we now have recipe descriptions in the Special section. I'd like to address the live populating of the skill list in another pull request.

I've added a database update, and added a line to the opening post on this pull request to reflect that.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/818/

cgolubi1:
> I'd like to address the live populating of the skill list in another pull request.

Yep, that's definitely reasonable.

cgolubi1:
Okay:
* Do the usual database "update vs. rebuild" regression check
* Bring up the site and make sure the RandomBM pages look visually sane

cgolubi1:
Seems fine.

PR #1678, opened 2015-08-08
Removed need to explicitly instantiate different BMInterface types

blackshadowshade:
Continues to address #974, based on comments in #1611.

I was able to create a game of Avis vs Bauer with this new code, so game creation still seems to work.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/813/ (if it passes)

### DIFF
--- a/src/api/ApiResponder.php
+++ b/src/api/ApiResponder.php
@@ -25,11 +25,6 @@ class ApiResponder {
         'login',
     );
 
-    // functions that handle player-related information
-    protected $playerFunctions = array(
-        'createGame',
-    );
-
     /**
      * Constructor
      * For live invocation:
@@ -97,7 +92,7 @@ class ApiResponder {
             $previousGameId = NULL;
         }
 
-        $retval = $interface->create_game(
+        $retval = $interface->player()->create_game(
             $playerIdArray,
             $buttonNameArray,
             $maxWins,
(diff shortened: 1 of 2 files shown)


blackshadowshade:
This is not yet complete, since it is causing a unit test to fail, and I don't yet know how to fix it.

blackshadowshade:
Right, this should be ready, Jenkins willing.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/814/ (if it passes)

cgolubi1:
The corpus in responderTest really should be sufficient to catch most regressions in this type of code, because it does go through ApiResponder and the changes here are reasonably small.  I want to check:
* Look at BMInterface.php for long enough to understand whether all `$this->message` settings have been changed to `$this->set_message()` or only some of them, and, if the latter, why
* Create some valid and invalid continuation games and make sure i get the correct message in both cases.

That'll probably do it.

cgolubi1:
Actually, in fact, `test_interface_game_003()` tests continuation failure, but anyway i tried changing the opponent in the continuating screen, and i got the expected error with the expected level of friendliness.

cgolubi1:
As for `set_message()`, i've convinced myself that BMInterface now uses `set_message()` and not `$this->message =` consistently, so this change is just making the continuation error case consistent.

So this looks fine.

PR #1680, opened 2015-08-15
Fixed weak and mighty so that they don't trigger when the die doesn't reroll

### LINKED ISSUE #1679: Konstant Weak die shrinks when used
In http://www.buttonweavers.com/ui/game.html?game=6922, I got this result:

> irilyth performed Skill attack using [k(8):8,hk(8):6] against [nq(4):2]; Defender nq(4) was captured; Attacker k(8) does not reroll; Attacker hk(8) changed size from 8 to 6 sides, recipe changed from hk(8) to hk(6), does not reroll

Konstant says "These dice do not reroll after an attack"; Weak says "When a weak die rerolls for any reason". I think the hk(8) should not have rerolled (as the log says it didn't), and thus also should not have changed size.

(When I used it again, I got

> irilyth performed Skill attack using [k(8):8,(8):7,hk(6):5] against [M(10):10]; Defender M(10) was captured; Attacker k(8) does not reroll; Attacker (8) rerolled 7 => 4; Attacker hk(6) changed size from 6 to 4 sides, recipe changed from hk(6) to hk(4), does not reroll

and it was changed to show 4 -- I was guessing it might turn into hk(4):5. Not relevant to the bug, but I was curious.)

blackshadowshade:
Fixes #1679.

Tested on a manually altered database state, with a skill attack using a hk(4) and a Hk(12) to capture an opponent's die.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/817/ (if it passes)

### DIFF
--- a/src/engine/BMSkillMighty.php
+++ b/src/engine/BMSkillMighty.php
@@ -52,6 +52,10 @@ class BMSkillMighty extends BMSkill {
             return;
         }
 
+        if (!$die->doesReroll) {
+            return;
+        }
+
         $die->grow();
     }
 
--- a/src/engine/BMSkillWeak.php
+++ b/src/engine/BMSkillWeak.php
@@ -52,6 +52,10 @@ class BMSkillWeak extends BMSkill {
             return;
         }
 
+        if (!$die->doesReroll) {
+            return;
+        }
+
         $die->shrink();
     }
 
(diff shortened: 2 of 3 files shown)


cgolubi1:
Regression-checking logic, not much to see here:
* Bring up game 33 from the responderTest corpus and make sure the new state looks visually right
* Play back a bunch of games with no mighty or weak, make sure no changes
* Play back a bunch of games with mighty or weak, make sure all changes are expected
* Play new games with mighty or weak, look for internal errors

cgolubi1:
Everything looks fine.  Merging.

Issue #1234, opened 2014-08-24
Create Game with random button doesn't follow chosen Button Set or Die Skill


jimmosk:
The Create Game screen automatically filters the pulldown list of allowable buttons to reflect choices made about Button Set and/or Die Skill.  But the "Random button" entry still appears at the top of this pulldown list.  This implies to me that if you select that entry, a button will be randomly chosen from the options in the pulldown list, but that's not what happens.  I just created two games, one specifying a Die Skill and the other specifying a Button Set, and chose the Random button entry.  Both games gave the players buttons chosen randomly from among all possible buttons.  

I think in this case the random buttons should follow the chosen options, limiting itself to only those that appear in the pulldown list.  But until that's implemented, I think the "Random button" entry at the top of said list ought to not show up if you've made restrictions to Button Set or Die Skill.

PR #1686, opened 2015-08-20
Removed Random Button option for when creating a new game with filters specified

blackshadowshade:
Provides a short-term fix for #1234.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/829/

### DIFF
--- a/src/ui/js/Newgame.js
+++ b/src/ui/js/Newgame.js
@@ -563,9 +563,16 @@ Newgame.updateButtonList = function(player, limitid) {
     });
   }
 
-  Newgame.activity.buttonList[player] = {
-    '__random': 'Random button',
-  };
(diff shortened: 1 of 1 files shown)


cgolubi1:
Seems plausible.  Nothing to do here but try it out and see how it looks and make sure game creation still works under various combinations of random and not-random.

cgolubi1:
* Okay, if i select "Random Button", and then limit by set, the selector goes back to 'Choose your button'.  That seems good.
* If i create an open game picking "Random Button", my opponent can still also pick "Random Button"
* If i use the selector to limit myself to Berserk buttons and pick Bruno and give my opponent "Random Button", they get a random button.

Nothing complicated here.  Merging.

PR #1687, opened 2015-08-20
Added button special information to game screens

### LINKED ISSUE #1200: Button specials need descriptions

blackshadowshade:
Fixes #1200.

I have added the button skills before the die skills, but not otherwise discriminated them, because this was the easiest thing to implement.

Tested with Giant vs RandomBMMixed, Echo vs Temperance, and Washu vs Shadow Warriors.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/834/

### DIFF
--- a/src/engine/BMBtnSkill.php
+++ b/src/engine/BMBtnSkill.php
@@ -43,7 +43,7 @@ class BMBtnSkill {
         }
 
         $skillDescription = array(
-            'code' => $skill,
+            'code' => '',
             'description' => $description,
             'interacts' => array(),
         );
--- a/src/engine/BMBtnSkill.php
+++ b/src/engine/BMBtnSkill.php
@@ -36,8 +36,8 @@ class BMBtnSkill {
                 return NULL;
             }
 
-            $skill = new $skillClass;
-            if ($skill instanceof BMBtnSkillArtificialReenable) {
+            $skillInstance = new $skillClass;
+            if ($skillInstance instanceof BMBtnSkillArtificialReenable) {
                 return NULL;
             }
         }
--- a/src/engine/BMBtnSkill.php
+++ b/src/engine/BMBtnSkill.php
@@ -25,13 +25,26 @@ class BMBtnSkill {
      * Complete description of skill, packaged for front end
      *
      * @param string $skill
-     * @return array
+     * @return array|NULL
      */
-    public static function describe($skill) {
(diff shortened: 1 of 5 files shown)


cgolubi1:
This looks quite reasonable, and i don't think there's any need to play RandomAI games and look for internal errors or anything, so i'll just bring it up and see how it looks, covering each of the cases of:
* Game has no button skills or die skills
* Game has button skills and no die skills
* Game has no button skills, and die skills
* Game has button skills and die skills
* Game has more than one different button skill
* Game has more than one identical button skill

I'll look for any weirdness in the apache error log, and call that good enough.

cgolubi1:
Hmm, why does the mouseover text for Giant say "Giant (Giant): Cannot win initiative"?  That doesn't seem right.

cgolubi1:
Nothing in the error log.  So the only issue i see is the "Giant (Giant)" thing.

src/engine/BMBtnSkill.php
```diff
@@ -20,4 +20,34 @@ class BMBtnSkill {
...
+            }
+
+            $skillInstance = new $skillClass;
+            if ($skillInstance instanceof BMBtnSkillArtificialReenable) {
+                return NULL;
+            }
+        }
+
+        $skillDescription = array(
+            'code' => $skill,
```
cgolubi1: Yeah, okay, now i understand what you're doing here that's leading to the doubled "Giant", and it's definitely a bug.  I guess it happens that the UI doesn't assume codes are single-character strings found in recipes, but there's no reason someone couldn't make that assumption in using API output, so i don't think it makes sense to break that.

I think your easiest solution is to leave 'code' empty (or not pass it), and have the UI not display the parens if there is no code.  If you wanted to pass a new parameter (e.g. 'type' => 'die' or 'type' => 'button'), you could, but that would be a lot more test updating, so i don't think you need to.  I think overloading 'code' is a bug.

blackshadowshade:
Okay, done. Tested with Giant vs Echo.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/841/ (if it passes)

cgolubi1:
Looks right to me based on the code.  Checking it out using the test games i made for the previous testing.

PR #1689, opened 2015-08-24
Sped up Overview page by adding indices to database table game_player_map

blackshadowshade:
Continues to address #1580.

Requires database update 01580_database_index_02.sql

Tested with Aug 2 backup of live site, logged in as irilyth (with a password manually copied across from my account). Load times drop from 5 seconds to 1 second.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/839/ (if it passes)

### DIFF
--- a/deploy/database/schema.game.sql
+++ b/deploy/database/schema.game.sql
@@ -43,7 +43,8 @@ CREATE TABLE game_player_map (
     last_action_time   TIMESTAMP DEFAULT 0,
     was_game_dismissed BOOLEAN DEFAULT FALSE NOT NULL,
     is_button_random   BOOLEAN DEFAULT FALSE NOT NULL,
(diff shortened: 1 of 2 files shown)


cgolubi1:
To do:
* the usual DB-update-vs-DB-rebuild test
* a DB-update test with the prod database already loaded, to make sure nothing weird happens
* before-and-after speed test of the prod database (i.e. what shadowshade tried, but under vagrant)

cgolubi1:
The speed test change wasn't very dramatic for me --- on virtualbox vagrant, irilyth's page loads for me in 1 second without this patch.  But i can see the difference even if i didn't both to time it, and i don't see any regressions here, so merging on the strength of shadowshade's testing.

PR #1692, opened 2015-08-31
Actively swallow a space character when there is an explicit handler for it

### LINKED ISSUE #1683: Using space bar to select dice also scrolls page
I often do things like this:

* Click on a die. That die is now selected (outlined in red).
* Hit tab. The next die to the right is now outlined in a thin dotted blue line.
* Hit the space bar.

Previously, that last step would select the outlined die (changing it to being outlined in red), which was great. As of a recent Firefox update (I think from 39.x to 40.0), that last step now selects the die and *also* scrolls the page, typically taking the entire battlemat area of the page out of view.

This doesn't happen when navigating with the keyboard among HTML forms, e.g. if I go to http://www.buttonweavers.com/ui/prefs.html, click in the "Real name" text box, hit tab (now the checkbox by "Make email address public" is selected, and hit the space bar, the checkbox is toggled and the page doesn't scroll.

Issue #180 originally introduced the keyboard navigation stuff, FWIW.

blackshadowshade:
Fixes #1683.

Tested on the game page. I'm not aware of other space handlers, but I might have missed some.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/844/

### DIFF
--- a/src/ui/js/Env.js
+++ b/src/ui/js/Env.js
@@ -212,14 +212,15 @@ Env.addClickKeyboardHandlers = function(
   if (spaceHandlerCallback || returnHandlerCallback) {
     element.keydown(function(eventData) {
 
-      // if a space handler was specified, invoke it on spacebar
-      if (spaceHandlerCallback && eventData.which == Env.KEYCODE_SPACEBAR) {
-        spaceHandlerCallback.call(element, eventData);
-      }
-
       // if a return handler was specified, invoke it on return
       if (returnHandlerCallback && eventData.which == Env.KEYCODE_RETURN) {
         returnHandlerCallback.call(element, eventData);
+      }
+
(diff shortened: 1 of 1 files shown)


blackshadowshade:
Tested further on AWS instance with my screen magnification at >200% to force the page to extend further than the end of the screen. Space and return both seem to work correctly for when the dice are highlighted, and when other elements are highlighted.

blackshadowshade:
Okay, I've put the order of keypress actions back to the way they were. After testing on my local MAMP stack, this fix continues to work.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/845/ (if it passes)

cgolubi1:
Looks fine to me.  Assuming it passes jenkins, i'm happy to merge it.

PR #1693, opened 2015-08-31
Add Unix account for irilyth

cgolubi1:
Not bothering with jenkins because jenkins doesn't test vagrant module changes.


cgolubi1:
Probably this is not too dangerous to do during the change freeze - if you merge it, i'll make sure to roll it onto dev before prod in case anything goes pear-shaped.

blackshadowshade:
This looks fine.

PR #1694, opened 2015-08-31
pull changes from master to dev on 2015-08-31

cgolubi1:
* User-visible:
    * #1692: fix spacebar keypress handling scrolling mismatch with newer versions of firefox
* Ideally not user-visible:
    * #1693: add Unix account for irilyth so he can do systems work if needed

Database updates: none

Once this pull is merged, i will load the changes on dev.buttonweavers.com as part of #217.


cgolubi1:
These are small changes and have both been tested under vagrant, so i'm going to go ahead and load this on dev.

PR #1700, opened 2015-09-07
Switched forum button locations

### LINKED ISSUE #1696: Swap order of Cancel and Save Edits buttons
When adding a new forum post, the Post Reply button is the next form element after the text box, which means that if you hit tab and then return, it posts the reply. This is pretty nice.

When entering or editing chat text in a game, if you're in the text box, hitting tab and then return also completes the attack or edits the text. Also nice.

When editing a forum post, the Cancel button is the next form element after the text box, which means that if you do the same thing, it cancels your edit. This is not so nice.

When posting a new forum thread, it's the same as when editing. Also not nice.

It'd be more consistent if the same combination did the same thing everywhere, and I think it'd be better if the first thing you get when you tab out of the text box is more like "do it" than like "throw away my text".

blackshadowshade:
Fixes #1696.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/857/

### DIFF
--- a/src/ui/js/Forum.js
+++ b/src/ui/js/Forum.js
@@ -198,18 +198,18 @@ Forum.showBoard = function() {
   contentTd.append($('<textarea>', {
     'maxlength': Forum.FORUM_BODY_MAX_LENGTH
   }));
-  var cancelButton = $('<input>', {
-    'type': 'button',
-    'value': 'Cancel',
-  });
-  contentTd.append(cancelButton);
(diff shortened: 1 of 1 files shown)


cgolubi1:
Test plan: pull this up in a test site and verify that i think it does indeed address the reported bug.

cgolubi1:
Looks fine to me.

PR #1621, opened 2015-05-15
Footer

blackshadowshade:
I'm putting this pull request up so that I can get some help with working out how to get the QUnit tests to pass.

This is based on Jota's footer incomplete pull request #1218. The current status of the pull request is that the footer is now flush to the bottom of the page. There are still a number of other comments from #1218 that remain unaddressed, and I'll deal with them once I get the QUnit tests passing.

PR #1702, opened 2015-09-08
Fill footer with some content

### LINKED ISSUE #648: Need a footer for the website
Currently, we are missing quite a number of features that are considered to be standard on any website, including:
- webmaster / contact details
- support / FAQs / link to issue tracker
- legal disclaimers and privacy policy (see #551)
- copyright (maybe)
- deployment date / time / version (e.g., production vs development)

blackshadowshade:
Fixes #648.

I don't consider this pull request to be complete yet, but I'd like guidance on what else needs to be done.

I've shifted some content into the footer as a placeholder. I expect the actual details of what goes in the footer to be altered in future pull requests.

I haven't addressed all of the queries that were left on #1621 and #1218, so feel free to repost those if they are still relevant.

I'd like to know if I've manipulated the QUnit tests correctly to get them to pass. In particular, I'm rather uncertain what the failing tests were trying to test, so I've tried to make the changes as minimal as possible.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/863/


cgolubi1:
I'm not going to make a formal test plan for this yet --- the next step is to bring it up under virtualbox and see how i think it looks, and then look at the code more closely.

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/886/

cgolubi1:
Whew.  Okay.  I think this is going to take some iteration, and i'm a little daunted getting started, but we should get the show on the road.  Let me start with some screencaps and some requests/observations based on the visuals.

cgolubi1:
First off, for whatever reason, i don't think your 100% height numbers are correct.  Maybe there's some extra padding somewhere, maybe it's the impact of the tabbed browser, i dunno, but the effect of what you have, in both Firefox and Safari, is that it defaults to enforcing just enough height that there has to be a scrollbar.  That doesn't seem right --- it should default to enforcing just enough height that there *doesn't* have to be a scrollbar:
front_page_with_footer

cgolubi1:
Second: you've heard this movie before, but it's still true --- don't put unnecessary vertical whitespace in the middle of elements we're trying to group.  Let's not have a blank line between the copyright line and the issue line.  The footer *should* look concise, it's the footer.

cgolubi1:
Third, i think whitespace *should* be used to separate elements which are not the same.  Right now, if you have a full page, the action log and the line and the footer are all right next to each other.  To my eye, the right fix here is to have a blank line *above* the `<hr>`, i.e. the page body should end with a blank line, because it's fine (desirable) for the footer to look compact and not have any internal whitespace.  You probably want to insert the blank line at the Login.js level, to make sure every page has it:
no_whitespace_above_footer

cgolubi1:
As for the code itself, i'm finding the diff very confusing because you moved a bunch of functions around, and i can't figure out what the change actually looks like.  So far, everything i've tried to actually track down and figure out whether you removed it from existence or not, it's seemed like you just moved it.  But i think i do need to drill down into this in more detail to make sure i can't think of any likely regressions to other page layout functionality.

Anyway, though, i think you can iterate on the three changes suggested above while i work on that:
* make the default page height not introduce a scrollbar
* remove internal whitespace within the footer
* insert whitespace above the line demarcating the footer

cgolubi1:
BTW, a hint (to me) about the first problem is that if you split the CSS definitions for "html" and "body", and set html's height to 99% and body's height to 98%, the scrollbar goes away, but with both at 99%, the scrollbar is still there.  To me, that implies that each of these elements introduces some amount of padding which you have to account for.

AdmiralJota:
Do we need to use percentages at all? Why not just set a min-height in pixels on div.c_body that will fit comfortably on 95% of desktop users' screens? I don't think there's any need to try to force the footer to always be at the <i>very</i> bottom of the actual window, just so long as it doesn't look awkwardly high up on empty pages.

blackshadowshade:
Jota, what sort of behaviour would you then expect on a smaller screen, e.g., a mobile device?

blackshadowshade:
Alternatively, Jota, what min-height would you suggest?

AdmiralJota:
Based on some Googling, it looks like 768 is the smallest common vertical resolution. The header renders at about 70px for me. If we expect the footer to be roughly similar, plus at least 100px for the menus, toolbars, taskbar, tabs, etc... that would suggest something between 480px-520px for the min-height for #c_body, if we wanted to make sure it didn't scroll for 95% of desktop users.

The largest common vertical resolution is apparently 1080px. Even with just a 480px body, the footer would be about halfway down the screen, which seems fine.

As for mobile devices, my aging iPhone 4s apparently has a resolution of 960x640. Even in landscape mode, that isn't much scrolling to get down to the footer (possibly none at all, depending on how big the location bar is). I think it should be alright; especially since <i>some</i> amount of scrolling is to be expected on a display the size of a smartphone.

Of course, the real test is for someone to try it out and see how it actually looks on their screen.

blackshadowshade:
This might be one of those cases where it might be worth spinning up a t1.micro instance or two with various manually-changed css types.

cgolubi1:
I guess my question is, why *not* use percentages?  We've found that 100% is too much, but if the goal we're trying to get is to get the footer to fit comfortably at the bottom of the screen, i actually think the current approach is very close to correct.

cgolubi1:
(That was really a question for Jota: what's the benefit to getting rid of the percentages altogether?)

cgolubi1:
Here's a slightly wacky visual effect, told in three screencaps:

I'm logged out on the front page:
step1_logged_out

Then i login:
step2_logged_in

Then i reload:
step3_reload

As you can (maybe) see, after logging in, an additional HR is drawn at the bottom of the page.  When i reload, it disappears again.

cgolubi1:
The same effect holds on other pages as well (i tried game.html) --- when i login, i get an additional HR at the bottom of the footer, then it disappears again when i reload.

cgolubi1:
Anyway, you've fixed all the issues i asked you to fix in the latest pull.  Now tracing through the Login.js logic as advertised.

blackshadowshade:
(For reference, the current valid Jenkins link for this pull request is http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1006/)

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1009/ (if it passes)

cgolubi1:
Okay, i think i pretty much understand the logic changes to the existing code (basically just rearranging the functions on the page to match execution order better, and splitting arrangeHeader and arrangePage into multiple functions for modularity), so i'm comfortable with this, and will merge it assuming jenkins passes and let the dev testers check it out.

AdmiralJota:
I'm uncomfortable with percentages that aren't just neatly dividing up the screen (like 50%/50% and the like), since when make one thing 99% or 98% or what have you, allotting 1% or 2% to something else, you never really know how big 2% will actually be on any given display. 

Then you wind up with situations where you're fudging by changing 99% to 98% to make it work in one browser on one monitor (like you were suggesting), but you don't know that you're not going to just wind up with the same problem at 98% on a different monitor, or at 97%, or if by that point you're now leaving so much space that it's not quite at the bottom on a third display, defeating the point, or...

If the goal is to actually have it be right at the bottom, then I'd suggest putting some javascript into the page to measure the height of the window and the height of the footer and to calculate the optimal height for the body based on that. But that's a pain to do, and I personally don't think it's worth it.

(All of that being said, when I say "I'm uncomfortable using percentages", I mean <i>I'm</i> uncomfortable using them. If what I just said doesn't bother you, then I'm not going to fight hard over it if <i>you</i> still want to use them.)

cgolubi1:
Jota: makes sense, but i see that as a danger no matter what you do --- counting percentages may get you the wrong result on some browser, but at least at the beginning, each browser starts with approximately 100% of its height.  It doesn't feel to me like as much of a slippery slope as guessing pixel count, where we really just don't have any idea how many pixels tall browsers are.  So i'm less uncomfortable with percents than with pixels.

cgolubi1:
Going ahead and merging this, and letting the dev testers have a whack at it.

PR #1706, opened 2015-09-11
Optimised button query, and added more database indices

blackshadowshade:
Continues to address #1580.

So, this pull request doesn't lead to huge improvements in performance, but it does stop the select_full_join count from increasing on:
- viewing button information (buttons.html and buttons.html?...)
- loading the create game page (create_game.html)

I watched the select_full_join count on my MAMP stack using the following query:

    SELECT * from information_schema.global_status where variable_name='select_full_join';

Then, I looked up the slow queries, which had been logged because I had a my.cnf with the following contents:

    [mysqld]
    log-slow-queries = /Users/james/slow.log
    long_query_time = 20
    log-queries-not-using-indexes

The button query was rewritten because MySQL views don't support indexing.

Requires database update 01580_database_indices_03.sql

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/865/ (if it passes)

### DIFF
--- a/deploy/database/schema.stats.sql
+++ b/deploy/database/schema.stats.sql
@@ -2,14 +2,16 @@ DROP TABLE IF EXISTS dhs_player;
 CREATE TABLE dhs_player (
     dhs_player_id   SMALLINT UNSIGNED NOT NULL,
     dhs_player_name VARCHAR(40) NOT NULL,
-    bw_player_id    SMALLINT UNSIGNED
+    bw_player_id    SMALLINT UNSIGNED,
+    INDEX (bw_player_id)
 );
 
 DROP TABLE IF EXISTS dhs_button;
 CREATE TABLE dhs_button (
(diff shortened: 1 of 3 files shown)


blackshadowshade:
After this pull has been merged, it may be useful to restart the MySQL server on the dev and live sites, so that the global indices are reset---if not, then I'd like a dump of the global status of the live site to #1580.

It might also be a good idea to turn on logging of slow queries on at least the dev site, if not the live site.

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/887/

cgolubi1:
This seems straightforward - i think the usual database non-regression test combined with loading the button data pages by hand should suffice.  If anything else broke, it should have been major enough that any responderTest should have caught it.

cgolubi1:
Seems fine.  Merging.

PR #1708, opened 2015-09-12
Hide "Mark thread as read" when creating thread, and disable it when editing reply

### LINKED ISSUE #1336: Prompt for confirmation before leaving a forum page with unsaved changes

blackshadowshade:
Fixes #1336.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/869/ (if it passes)

### DIFF
--- a/src/ui/js/Forum.js
+++ b/src/ui/js/Forum.js
@@ -490,6 +490,9 @@ Forum.quotePost = function() {
 };
 
 Forum.editPost = function() {
+  $('#markThreadReadButton').attr('disabled', 'disabled');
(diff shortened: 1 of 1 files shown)


cgolubi1:
Let me just bring this up on a test site and see how it looks.

cgolubi1:
Seems to work.  Merging.

PR #1715, opened 2015-09-17
Removed fire message overriding

### LINKED ISSUE #1489: Report of attack results preempted by Fire dice turndown message
This occurred on http://buttonweavers.com, not the dev site, though I don't know whether it would have happened differently there: In game http://buttonweavers.com/ui/game.html?game=2989 I made an attack that involved turning down a Fire die. Normally after an attack is made, text at the top (in green) shows what the results of the attack was. But after this turn (see image), all I got was "Successfully turned down fire [sic] dice" - which is not particularly interesting or helpful. 
The attack results *are* displayed as normal in the "Recent game activity" section at the bottom, so it's more of a peculiarity than a bug... but I still like always being able to see the results up there at the top.

screen shot 2015-01-03 at 9 31 27 pm

blackshadowshade:
Fixes #1489.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/874/

### DIFF
--- a/src/engine/BMGame.php
+++ b/src/engine/BMGame.php
@@ -1593,7 +1593,6 @@ class BMGame {
             'newValues' => $newValues,
         );
 
-        $this->message = 'Successfully turned down fire dice.';
         return TRUE;
     }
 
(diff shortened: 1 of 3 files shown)


cgolubi1:
I'm just going to throw this one over the fence --- i figured it would be that easy, and the responderTest changes look exactly like i think they should.

PR #1718, opened 2015-09-23
Added attack type to fire turndown message

### LINKED ISSUE #1710: Attack not specified on FIre turndown page (outside of history)
When I go to the Fire turndown page, nothing specifies what kind of attack I'm in the middle of performing (skill versus power). 

I did recently find it in the history, but I was expecting to find it in the attack area.

This becomes important if a user did a "Default Attack" which became a power attack allowing overshooting with Fire dice. A newbie may not know what is happening.

blackshadowshade:
Fixes #1710.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/879/

### DIFF
--- a/src/ui/js/Game.js
+++ b/src/ui/js/Game.js
@@ -883,7 +883,7 @@ Game.actionAdjustFireDiceActive = function() {
     fireMessage += exactFiringAmount;
   }
 
-  fireMessage += '.';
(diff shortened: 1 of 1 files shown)


cgolubi1:
This does not look very complicated.  Seeing if it works.

cgolubi1:
Looks good.

PR #1725, opened 2015-09-30
Fixed default attack when fire overshooting is off

### LINKED ISSUE #1721: Overshooting preferences off, but ambiguous default attack message appears
In a game I'm doing a default attack where the target button value matches my value and my ovreshooting preference is off:

screen shot 2015-09-27 at 10 50 23 am

This should not be ambiguous!

blackshadowshade:
Fixes #1721.

Testing pull request: #2111.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1232/ (if it passes)

### DIFF
--- a/src/engine/BMAttackDefault.php
+++ b/src/engine/BMAttackDefault.php
@@ -196,7 +196,10 @@ class BMAttackDefault extends BMAttack {
         if ($fireTurndownAvailable > 0) {
             // power attack with the possibility of fire assistance
             if (in_array('Power', $validAttackTypes) && ($attacker->value < $attacker->max)) {
-                return TRUE;
+                if (($attacker->value < $defender->value) ||
+                    $game->fireOvershootingArray[$game->attackerPlayerIdx]) {
+                    return TRUE;
+                }
             }
 
             // skill attack with the need for fire assistance


blackshadowshade:
Tested Raupe vs Raupe, with one player with fire overshooting on, and the other with fire overshooting off. I always used a speed die to target another die with the exact same value.

cgolubi1:
This seems like the right fix to me, and the logic change is only within fire code, so it seems like just replaying a bunch of fire games, playing a reasonable selection of new fire games, and verifying what shadowshade tested, should suffice.

cgolubi1:
Verified by hand that:
* fire-assisted power attacks still work when the attacker has a lower value than the defender, including both cases in which the attacker winds up with the same values as the defender, and cases in which the attacker winds up overshooting
* fire assistance is not suggested when the attacker has a greater value than the defender
* fire assistance is not suggested when the attacker has the same values as the defender

cgolubi1:
Okay, played back on the order of 50-60 games so far, no issues seen.  I'd like to play back some more, and also play some new games, when i get back to this.

cgolubi1:
Per discussion elsewhere, i now think i'm not planning to merge this in December, though i'll do more of the planned testing as time permits.

blackshadowshade:
I'm pretty sure that this should be okay without a new Jenkins, since it doesn't impact any of the unit tests.

blackshadowshade:
Rebased to apply after turbo.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1308/ (if it passes)

cgolubi1:
This has been waiting for a replay loop since before replay_loop existed, so i stood one up.  It's playing a mix of 75% Fire, 25% anything novel games, plus replay games.  We can't be a rush on the order of a week, so let's say, if we don't see anything weird here in a week, this is good to go.

PR #1727, opened 2015-10-02
Allow game rejection

### LINKED ISSUE #205: Offer the option to reject a game

blackshadowshade:
Fixes #205.

This pull request does the following:
- add a new section to the Overview page called "New games", showing games that have not yet been accepted
- adds a user preference to auto-accept games, default on

Currently, the game page is minimal for a not-yet-accepted game or for a rejected game.

Requires database update 00205_reject_game_02.sql

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/898/

### DIFF
--- a/src/api/ApiSpec.php
+++ b/src/api/ApiSpec.php
@@ -23,6 +23,7 @@ class ApiSpec {
     private $automatableApiCalls = array(
         'loadNextPendingGame',
         'loadNextNewPost',
+        'loadNewGames',
         'loadActiveGames',
         'loadCompletedGames',
         'loadPlayerInfo',
@@ -343,6 +344,10 @@ class ApiSpec {
             'permitted' => array(
                 'logEntryLimit' => 'number',
             ),
+        ),
+        'loadNewGames' => array(
+            'mandatory' => array(),
+            'permitted' => array(),
         ),
         'loadNextPendingGame' => array(
             'mandatory' => array(),
--- a/src/api/ApiSpec.php
+++ b/src/api/ApiSpec.php
@@ -158,15 +158,10 @@ class ApiSpec {
                 'previousGameId' => 'number',
             ),
         ),
-        'acceptGame' => array(
+        'reactToNewGame' => array(
             'mandatory' => array(
                 'gameId' => 'number',
-            ),
-            'permitted' => array(),
-        ),
-        'rejectGame' => array(
-            'mandatory' => array(
-                'gameId' => 'number',
+                'action' => 'string',
             ),
             'permitted' => array(),
         ),
--- a/src/api/ApiSpec.php
+++ b/src/api/ApiSpec.php
@@ -522,6 +522,7 @@ class ApiSpec {
                     'arg_type' => 'string',
                     'maxlength' => 100,
                 ),
+                'autoaccept' => 'boolean',
                 'autopass' => 'boolean',
                 'fire_overshooting' => 'boolean',
                 'monitor_redirects_to_game' => 'boolean',
--- a/src/api/ApiSpec.php
+++ b/src/api/ApiSpec.php
@@ -158,6 +158,18 @@ class ApiSpec {
                 'previousGameId' => 'number',
             ),
         ),
+        'acceptGame' => array(
+            'mandatory' => array(
+                'gameId' => 'number',
+            ),
+            'permitted' => array(),
+        ),
+        'rejectGame' => array(
+            'mandatory' => array(
+                'gameId' => 'number',
+            ),
+            'permitted' => array(),
+        ),
         'dismissGame' => array(
             'mandatory' => array(
                 'gameId' => 'number',
--- a/src/api/DummyApiResponder.php
+++ b/src/api/DummyApiResponder.php
@@ -272,6 +272,34 @@ class DummyApiResponder {
         return $game;
     }
 
+    protected function get_interface_response_loadNewGames() {
+        $data = array(
+            'gameIdArray' => array(),
+            'opponentIdArray' => array(),
+            'opponentNameArray' => array(),
+            'myButtonNameArray' => array(),
+            'opponentButtonNameArray' => array(),
+            'nWinsArray' => array(),
+            'nLossesArray' => array(),
+            'nDrawsArray' => array(),
+            'nTargetWinsArray' => array(),
+            'isAwaitingActionArray' => array(),
+            'gameStateArray' => array(),
+            'statusArray' => array(),
+            'inactivityArray' => array(),
+            'inactivityRawArray' => array(),
+            'playerColorArray' => array(),
+            'opponentColorArray' => array(),
+        );
+
+//        for ($gameIdx = 1; $gameIdx <= 25; $gameIdx++) {
+//            $funcname = 'add_active_game_data_'.$gameIdx;
+//            $this->$funcname($data);
+//        }
+
+        return array($data, "All game details retrieved successfully.");
+    }
+
     protected function get_interface_response_loadActiveGames() {
         // Use the same fake games here which were described in loadGameData
         $data = array(
--- a/src/api/DummyApiResponder.php
+++ b/src/api/DummyApiResponder.php
@@ -292,11 +292,6 @@ class DummyApiResponder {
             'opponentColorArray' => array(),
         );
 
-//        for ($gameIdx = 1; $gameIdx <= 25; $gameIdx++) {
-//            $funcname = 'add_new_game_data_'.$gameIdx;
-//            $this->$funcname($data);
-//        }
-
         return array($data, "All game details retrieved successfully.");
     }
 
@@ -1233,12 +1228,14 @@ class DummyApiResponder {
         return array(NULL, "function not implemented");
     }
 
-    protected function get_interface_response_acceptGame() {
-        return array(TRUE, 'Successfully accepted game');
-    }
-
-    protected function get_interface_response_rejectGame() {
-        return array(TRUE, 'Successfully rejected game');
+    protected function get_interface_response_reactToNewGame($args) {
+        if ('accept' == $args['action']) {
+            return array(TRUE, 'Successfully accepted game');
+        } elseif ('reject' == $args['action']) {
+            return array(TRUE, 'Successfully rejected game');
+        } else {
+            return FALSE;
+        }
     }
 
     protected function get_interface_response_dismissGame() {
--- a/src/api/DummyApiResponder.php
+++ b/src/api/DummyApiResponder.php
@@ -1233,6 +1233,14 @@ class DummyApiResponder {
         return array(NULL, "function not implemented");
     }
 
+    protected function get_interface_response_acceptGame() {
+        return array(TRUE, 'Successfully accepted game');
+    }
+
+    protected function get_interface_response_rejectGame() {
+        return array(TRUE, 'Successfully rejected game');
+    }
+
     protected function get_interface_response_dismissGame() {
         return array(TRUE, 'Dismissing game succeeded');
     }
--- a/src/api/DummyApiResponder.php
+++ b/src/api/DummyApiResponder.php
@@ -1229,10 +1229,8 @@ class DummyApiResponder {
     }
 
     protected function get_interface_response_reactToNewGame($args) {
-        if ('accept' == $args['action']) {
-            return array(TRUE, 'Successfully accepted game');
-        } elseif ('reject' == $args['action']) {
-            return array(TRUE, 'Successfully rejected game');
+        if (array_key_exists('action', $args)) {
+            return array(TRUE, 'Successfully ' . $args['action'] . 'ed game');
         } else {
             return FALSE;
         }
--- a/src/api/DummyApiResponder.php
+++ b/src/api/DummyApiResponder.php
@@ -1112,6 +1112,7 @@ class DummyApiResponder {
                                 'dob_day' => 0,
                                 'gender' => '',
                                 'image_size' => NULL,
+                                'autoaccept' => TRUE,
                                 'autopass' => TRUE,
                                 'fire_overshooting' => FALSE,
                                 'uses_gravatar' => FALSE,
--- a/src/api/DummyApiResponder.php
+++ b/src/api/DummyApiResponder.php
@@ -293,7 +293,7 @@ class DummyApiResponder {
         );
 
 //        for ($gameIdx = 1; $gameIdx <= 25; $gameIdx++) {
-//            $funcname = 'add_active_game_data_'.$gameIdx;
+//            $funcname = 'add_new_game_data_'.$gameIdx;
 //            $this->$funcname($data);
 //        }
 
--- a/src/ui/js/Overview.js
+++ b/src/ui/js/Overview.js
@@ -436,21 +436,28 @@ Overview.pageAddGameTableNew = function() {
 
     decideTd = $('<td>');
     gameRow.append(decideTd);
-    acceptLink = $('<a>', {
-      'text': 'Accept',
-      'href': '#',
-      'data-gameId': gameInfo.gameId,
-    });
-    acceptLink.click(Overview.formAcceptGame);
+
+    if (Api.new_games.games[0].isAwaitingAction) {
+      acceptLink = $('<a>', {
+        'text': 'Accept',
+        'href': '#',
+        'data-gameId': gameInfo.gameId,
+      });
+      acceptLink.click(Overview.formAcceptGame);
+
+      decideTd.append('[')
+              .append(acceptLink)
+              .append('] / ');
+    }
+
     rejectLink = $('<a>', {
       'text': 'Reject',
       'href': '#',
       'data-gameId': gameInfo.gameId,
     });
     rejectLink.click(Overview.formRejectGame);
+
     decideTd.append('[')
-            .append(acceptLink)
-            .append('] / [')
             .append(rejectLink)
             .append(']');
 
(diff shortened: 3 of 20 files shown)


blackshadowshade:
Both irilyth and cgolubi1 have expressed a strong preference for having auto-accept to default to off. I would submit that this should be the subject of a later pull request, because I anticipate that it will cause significant changes in our test cases.

blackshadowshade:
Then again, I could be wrong. I'll take a look and see how doable it is to have the default set to off.

blackshadowshade:
Okay, I think I've dealt with the changes that are required to have the auto-accept default set to off.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/899/ (if it passes)

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/901/ (if it passes)

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/902/

src/api/ApiSpec.php
```diff
@@ -157,6 +158,18 @@ class ApiSpec {
...
+                'gameId' => 'number',
+            ),
+            'permitted' => array(),
+        ),
+        'rejectGame' => array(
+            'mandatory' => array(
+                'gameId' => 'number',
+            ),
+            'permitted' => array(),
+        ),
```
cgolubi1: It doesn't make sense to me that `acceptGame` and `rejectGame` should be two different API functions.  Why is this different from e.g. `reactToAuxiliary`, where something has been offered to you and you have to decide whether or not to do it (using the same API function either way with a different argument)?  I'd call this function something like `reactToOfferedGame` and make it behave like the other react guys.  At the BMInterface level, you're implementing it that way anyway (i.e. the function is `save_join_game_decision()` regardless), so why have two functions at the API level?

src/api/DummyApiResponder.php
```diff
@@ -272,6 +272,34 @@ protected function mock_active_game($args) {
...
+            'inactivityArray' => array(),
+            'inactivityRawArray' => array(),
+            'playerColorArray' => array(),
+            'opponentColorArray' => array(),
+        );
+
+//        for ($gameIdx = 1; $gameIdx <= 25; $gameIdx++) {
+//            $funcname = 'add_new_game_data_'.$gameIdx;
+//            $this->$funcname($data);
+//        }
```
cgolubi1: Nit: get rid of this commented-out block.

src/ui/js/Overview.js
```diff
@@ -369,6 +378,94 @@ Overview.pageAddGameTable = function(
...
+
+      decideTd.append('[')
+              .append(acceptLink)
+              .append('] / ');
+    }
+
+    rejectLink = $('<a>', {
+      'text': 'Reject',
+      'href': '#',
+      'data-gameId': gameInfo.gameId,
```
cgolubi1: This is a total nit, but something looks weird about this attribute with a hyphen in it to me.  I don't think that's the variable style of the UI - are you sure camelcase or an underscore (or just calling it `gameId` wouldn't be more consistent)?

test/src/api/responderTest.php
```diff
@@ -1252,6 +1341,22 @@ public function test_request_loadOpenGames() {
...
+
+        $_SESSION = $this->mock_test_user_login();
+        $this->verify_invalid_arg_rejected('loadNewGames');
+
+        $args = array('type' => 'loadNewGames');
+        $retval = $this->verify_api_success($args);
+        $dummyval = $this->dummy->process_request($args);
+
+        $this->assertEquals('ok', $dummyval['status'], 'Dummy load of new games should succeed');
+    }
```
cgolubi1: Seems to me that `loadNewGames` is in some way the simple part of this functionality, and you're not testing it very deeply anyway --- what's really important is to have a responderTest on accepting and rejecting new games --- you could pattern the API wrapper off `verify_api_joinOpenGame()` and the test itself off `test_request_joinOpenGame()`, and that would be pretty simple and add regression-testing of the logic around accepting and rejecting games.

cgolubi1:
Okay, so, after a first pass through the code, what i need to do here:
* There's a database change, so do a database rebuild/update test
* Make sure things like the history page still work fully given changes to `get_all_games()`
* Do side-by-side comparisons of the new and old Overview pages and look for visual warts
* Try out the new functionality by hand in all combinations of autoaccept/not and accept/reject of games
* Try out the existing "open game" functionality in all combinations of button-specified/not, and make sure that still works

Things to look for:
* does it cause any problems that REJECTED > END_GAME anywhere else in the code?
* `BMInterface::create_game()` is defined to default to `$autoAccept = TRUE` - is it actually called with that overridden?
* Do i agree that `Overview.pageAddGameTableNew()` should be a totally separate function, rather than extra logic in the existing function?

src/ui/js/Overview.js
blackshadowshade: I seem to remember that I copied this directly from somewhere else in the code.

src/api/ApiSpec.php
blackshadowshade: That sounds reasonable. I'm happy to rework this.

test/src/api/responderTest.php
blackshadowshade: Okay, sounds reasonable.

cgolubi1:
Accepting a game appeared to work, but rejecting a game didn't:
reject_game_internal_error

The error log says:
```
[Wed Oct 07 12:33:49 2015] [error] [client 10.0.2.2] PHP Fatal error:  Uncaught exception 'LogicException' with message 'Infinite loop detected when advancing game state. Final game state: REJECTED' in /var/www/engine/BMGame.php:1956\nStack trace:\n#0 /var/www/engine/BMInterface.php(991): BMGame->proceed_to_next_user_action()\n#1 /var/www/engine/BMInterface.php(456): BMInterface->save_game(Object(BMGame))\n#2 /var/www/api/ApiResponder.php(473): BMInterface->save_join_game_decision('2', '3', 'reject')\n#3 /var/www/api/ApiResponder.php(606): ApiResponder->get_interface_response_rejectGame(Object(BMInterface), Array)\n#4 /var/www/api/responder.php(21): ApiResponder->process_request(Array)\n#5 {main}\n  thrown in /var/www/engine/BMGame.php on line 1956, referer: http://localhost:8080/ui/index.html
```

cgolubi1:
Is this related to the fact that Ed, one of the chosen buttons, has an auxiliary die?  (I picked random vs random.)  I'll try Soldiers to check that.

blackshadowshade:
Ah, how could I have missed that! I'll add the appropriate logic to BMGame->proceed_to_next_action().

cgolubi1:
No.  I get the same thing with Changeling vs. Hammer.  I haven't looked at the code that's throwing this exception at all, but above i noted that i was suspicious of REJECTED being greater than END_GAME in the state table, because i had a sense that a bunch of the code expects END_GAME to be the biggest state.  So that would be what i would take a look at.

cgolubi1:
Summarizing my comments so far:
* Game rejection doesn't work
* There should be a responderTest regression-testing game rejection (and also game acceptance, but that one can wait until acceptGame and rejectGame are combined)
* acceptGame and rejectGame should be combined into one API function that takes an argument
* a nit about removing commented-out test code from DummyApiResponder
* a nit about an argument name in the UI

And i think once this comes back over the fence with rejection working, my previous test plan will still be valid.

src/ui/js/Overview.js
blackshadowshade: Right, I've found where I copied it from, namely the Dismiss Game code.

If you do a global search, though, on "data-", you'll find things like "data-boardId", "data-postId", "data-column", "data-direction" throughout all the API code.

blackshadowshade:
So, in a follow-up about the hyphen in the argument name, there are loads of these scattered throughout the UI. Do you want them all changed to underscores?

blackshadowshade:
Okay, after talking to Chaos about this, I'll leave the hyphen in for the moment, and after this pull request has been accepted, I'll submit a pull request doing a global replace of all the relevant hyphens with underscores.

blackshadowshade:
I believe that I've addressed all the issues. 

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/907/ (if it passes)

cgolubi1:
Changes look reasonable.  Going back to my test plan.

cgolubi1:
Okay, so, c1 vs c2, neither player has autoaccept.  Now, c1 creates two open games:
* Niles vs Any Button
* Niles vs Hammer

Now, c2 joins both games:
* In the first game, c2 selects Iago and joins the game.  That game starts immediately.
* In the second game, c2 simply joins the game.  Hmm... that game also starts immediately.

But i got different behavior before.  How to reproduce it?

Ah!  I see.  This is about Random Button.  Okay, here we go:

Scenario 1:
* c1 specifies that c1's own button should be "Random button" and c2's button should be "Any button"
* c2 joins the game and selects a button
* The game starts immediately

Scenario 2:
* c1 specifies that c1's own button should be "Random button" and picks a (non-random) button for c2
* c2 joins the game
* The game does not start immediately --- c2 has to accept the game

Scenario 3, in which c1 specifies "Random button" for both players, works like scenario 2 --- c2 has to accept again.

Scenario 4, in which c1 specifies a non-random button for c1 and "Random button" for c2, works like scenarios 2 and 3 --- c2 has to accept again.

So, what should happen?  It seems like scenario 1 is the outlier --- in "Random button vs button of c2's choice", c2 doesn't get to see the random button before accepting.  In all other scenarios, c2 gets to see the random button before accepting.

Shadowshade, what did you have in mind here?

blackshadowshade:
To be honest, I hadn't actually considered the Random Button cases explicitly. Perhaps the acceptance of a Random Button matchup should start automatically, so that the button remains random until the game starts.

cgolubi1:
Okay, so, in summary, i think the test cases here are:
* Fixed button vs fixed button (p2 has no autoaccept)
* Fixed button vs random button (p2 has no autoaccept)
* Random button vs fixed button ( p2 has no autoaccept)
* Random button vs random button (p2 has no autoaccept)

* Fixed button vs fixed button (p2 has autoaccept)
* Fixed button vs random button (p2 has autoaccept)
* Random button vs fixed button ( p2 has autoaccept)
* Random button vs random button (p2 has autoaccept)

* Fixed button vs fixed button (open, p2 has no autoaccept)
* Fixed button vs random button (open, p2 has no autoaccept)
* Fixed button vs any button (open, p2 has no autoaccept)
* Random button vs fixed button (open, p2 has no autoaccept)
* Random button vs random button (open, p2 has no autoaccept)
* Random button vs any button (open, p2 has no autoaccept)

* Fixed button vs fixed button (open, p2 has autoaccept)
* Fixed button vs random button (open, p2 has autoaccept)
* Fixed button vs any button (open, p2 has autoaccept)
* Random button vs fixed button (open, p2 has autoaccept)
* Random button vs random button (open, p2 has autoaccept)
* Random button vs any button (open, p2 has autoaccept)

That's 20 cases, having enumerated them, let's just try them all.

cgolubi1:
* Fixed button vs fixed button (p2 has no autoaccept) - p2 has to accept
* Fixed button vs random button (p2 has no autoaccept) - p2 is shown the identity of the random button and has to accept
* Random button vs fixed button ( p2 has no autoaccept) - p2 is shown the identity of the random button and has to accept
* Random button vs random button (p2 has no autoaccept) - p2 is shown the identity of the random buttons and has to accept

* Fixed button vs fixed button (p2 has autoaccept) - p2 autoaccepts
* Fixed button vs random button (p2 has autoaccept) - p2 autoaccepts
* Random button vs fixed button ( p2 has autoaccept) - p2 autoaccepts
* Random button vs random button (p2 has autoaccept) - p2 autoaccepts

* Fixed button vs fixed button (open, p2 has no autoaccept) - game starts immediately
* Fixed button vs random button (open, p2 has no autoaccept) - p2 has to accept
* Fixed button vs any button (open, p2 has no autoaccept) - game starts immediately
* Random button vs fixed button (open, p2 has no autoaccept) - p2 has to accept
* Random button vs random button (open, p2 has no autoaccept) - p2 has to accept
* Random button vs any button (open, p2 has no autoaccept) - game starts immediately

* Fixed button vs fixed button (open, p2 has autoaccept) - game starts immediately
* Fixed button vs random button (open, p2 has autoaccept) - p2 has to accept (BUG - p2 has autoaccept)
* Fixed button vs any button (open, p2 has autoaccept) - game starts immediately
* Random button vs fixed button (open, p2 has autoaccept) - p2 has to accept (BUG - p2 has autoaccept)
* Random button vs random button (open, p2 has autoaccept) - p2 has to accept (BUG - p2 has autoaccept)
* Random button vs any button (open, p2 has autoaccept) - game starts immediately

cgolubi1:
So the upshot of all that nonsense is two bugs:
* In the case in which p1 creates an open game where p1 has "Random button" and p2 has "Any button", and p2 joins, the game starts immediately.  In all other cases of "Random button", p2 sees the identity of the random button when deciding whether to accept.  Those two cases should be consistent.
* In the case in which p1 creates an open game and either player has "Random button", the UI ignores autoaccept when p2 joins, and always makes p2 decide whether to accept the game even if p2 has autoaccept.

I believe shadowshade's current proposal is to not resolve Random Button until the game has been accepted.  If implemented, that should cause both of these inconsistencies to go away.

blackshadowshade:
Okay, that should be fixed now.

cgolubi1:
## specified opponent who does not have autoaccept

* Fixed button vs fixed button (p2 has no autoaccept) - p2 has to accept
* Fixed button vs random button (p2 has no autoaccept) - p2 is shown the identity of the random button and has to accept
* Random button vs fixed button ( p2 has no autoaccept) - p2 is shown the identity of the random button and has to accept
* Random button vs random button (p2 has no autoaccept) - p2 is shown the identity of the random buttons and has to accept

## specified opponent who has autoaccept

* Fixed button vs fixed button (p2 has autoaccept) - p2 autoaccepts
* Fixed button vs random button (p2 has autoaccept) - p2 autoaccepts
* Random button vs fixed button ( p2 has autoaccept) - p2 autoaccepts
* Random button vs random button (p2 has autoaccept) - p2 autoaccepts

## open game joined by opponent who does not have autoaccept

* Fixed button vs fixed button (open, p2 has no autoaccept) - game starts immediately
* Fixed button vs random button (open, p2 has no autoaccept) - game starts immediately
* Fixed button vs any button (open, p2 has no autoaccept) - game starts immediately
* Random button vs fixed button (open, p2 has no autoaccept) - game starts immediately
* Random button vs random button (open, p2 has no autoaccept) - game starts immediately
* Random button vs any button (open, p2 has no autoaccept) - game starts immediately

## open game joined by opponent who has autoaccept

* Fixed button vs fixed button (open, p2 has autoaccept) - game starts immediately
* Fixed button vs random button (open, p2 has autoaccept) - game starts immediately
* Fixed button vs any button (open, p2 has autoaccept) - game starts immediately
* Random button vs fixed button (open, p2 has autoaccept) - game starts immediately
* Random button vs random button (open, p2 has autoaccept) - game starts immediately
* Random button vs any button (open, p2 has autoaccept) - game starts immediately

cgolubi1:
I was going to look for equality checks against the END_GAME state.  Here's what i found:
* In `BMInterface::build_game_log_query_restrictions()` - i believe this one will mean that once someone creates a display for rejected games (and there *should* be a display for rejected games, i always disliked the thing on DHS where once a game went away it was just gone and you couldn't find out anything about what had been proposed --- if we ever do get our own autoreject bug, it'll make life much easier for the players in sending bug reports), rejected games will still show the chat from a previous game.  So that's easy enough to fix now while we have the context.
* In `BMInterface::set_logEntryLimit()` - this one actually may literally never come up, just mentioning it for completeness.

cgolubi1:
I'm ready to merge this.  Shadowshade, you should fix the `build_game_log_query_restrictions` thing soon when you have a chance.

I have a couple of thoughts, namely:
* We should decide intentionally what we want to do about Random Button (i.e. is it good that the players see the selected Random Buttons before deciding whether to accept the game?)
* I think that, at a high level, it should be quite easy to show some UI screen for each of the cases of rejected game, game waiting for someone to decide whether to join, and open game.  I'm not necessarily talking about embedding the forms in the individual pages (we should do that too, but it's another step), but we should be able to show a display like the auxiliary/reserve/swing one in each of those cases, it shouldn't be a lot of code and it shouldn't be difficult code.  That's not a blocker for accepting this pull, but i expect people are going to ask about it because the "nothing to see here, move along" screen is ugly, and it's a pity for us to show that screen when the backend is actually passing along full API contents that we could use to assemble a page.

Neither of those is for this pull request, but i think they'll both come up this month and will be important.

blackshadowshade:
I think we should pass the question about random buttons on to the testers and see what they think. I guess it depends on why people use "Random button"---is it laziness, indecision, for the thrill of the unknown choice, or as a way of getting unlikely matchups (including because no one would play this matchup)?

I'll certainly move on to showing some sort of UI screen for the cases where we have no decent UI, but I'll need to take a look at how the pages are currently built before I know how to do this.

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/910/

cgolubi1:
Great.  Let's do this thing.

PR #1740, opened 2015-10-19
TimeAndSpace now doesn't trigger if the die doesn't reroll

### LINKED ISSUE #1729: Boom Time & Space dice get an extra turn if they Boom Attack while odd
In http://www.buttonweavers.com/ui/game.html?game=7561, I did this attack:

> irilyth performed Boom attack using [b^(20):5] against [q(12):11]; Defender q(12) rerolled 11 => 5, was not captured; Attacker b^(20) does not reroll, was taken out of play

That seems fine, but then it said

> irilyth gets another turn because a Time and Space die rolled odd

That seems wrong: My b^(20):5 was odd when it made its Boom Attack, and it's true that it didn't reroll, but that also means that it didn't "roll odd".

blackshadowshade:
Fixes #1729.

Tested with artificially altered die recipes in the database, using odd-valued b^ and k^ dice.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/913/

### DIFF
--- a/src/engine/BMSkillTimeAndSpace.php
+++ b/src/engine/BMSkillTimeAndSpace.php
@@ -42,6 +42,10 @@ class BMSkillTimeAndSpace extends BMSkill {
             return;
         }
 
+        if (!$die->doesReroll) {
+            return;
+        }
+
         if ($die->value & 1) {
             $game->nextPlayerIdx = $game->activePlayerIdx;
         }


cgolubi1:
I should be able to do most of my testing here looking for regressions in the usual way:
* Play back a bunch of games and make sure there are no regressions
* Play a bunch of new games with TimeAndSpace and RandomBMMixed, look for internal errors, and verify at least one example of this bug being fixed

cgolubi1:
Okay, so i played back 1500 games and found half a dozen regressions, all of which were either boom attacks using `b^` dice or skill attacks using `k^` dice.  I looked at one of the `k^` games, and it seemed to be valid and to play fine under the new codebase.

cgolubi1:
I played 200 RandomAI T&S vs. RandomBMMixed games, and 100 T&S vs. T&S games, and saw no internal errors.  This looks like it works to me, so i'm going to merge it.

PR #1741, opened 2015-10-22
Improve "Skills in this game" section

### LINKED ISSUE #1697: Make "Skills in this game" section more user-friendly
There are two main points here, suggested by @irilyth and @nkmcalli, amongst others:

- Button skills should be visually separated from Die skills
- Die skills should show their single character abbreviation

blackshadowshade:
Fixes #1697.

This implements four improvements:
- button specials are separated from die skills
- die skill codes are shown explicitly in the list of skills
- hint text now appears both on click and on hover
- hint text is now larger than before

Tested with Giant vs Echo, Prudence vs Giant, Bauer vs Avis.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/916/

### DIFF
--- a/src/ui/js/Game.js
+++ b/src/ui/js/Game.js
@@ -1690,6 +1690,7 @@ Game.pageAddSkillListFooter = function() {
   $.each(Api.game.gameSkillsInfo, function(skill, info) {
     skillDesc = skill;
     if (info.code) {
+      skill += ' (' + info.code + ')';
       skillDesc += ' (' + info.code + ')';
     }
     skillDesc += ': ' + info.description;
--- a/src/ui/js/Game.js
+++ b/src/ui/js/Game.js
@@ -1726,8 +1726,11 @@ Game.pageAddSkillListFooter = function() {
         items: 'span'
       })
       .on('click', function() {
+        // james: without closing other tooltips, it's possible that
+        // the same tooltip might occur twice on a browser page due
+        // to the combination of hover and click
+        $('.ui-tooltip').hide();
         $(this).tooltip('open');
-        return false;
       });
 
     gameSkillDiv.append(indivSkillSpan);
--- a/src/ui/js/Game.js
+++ b/src/ui/js/Game.js
@@ -1707,10 +1707,10 @@ Game.pageAddSkillListFooter = function() {
       isFirstInteraction = true;
       $.each(info.interacts, function(otherSkill, interactDesc) {
         if (isFirstInteraction) {
-          skillDesc += '\n\nInteraction with other skills in this game:';
+          skillDesc += '<br/><br/>Interaction with other skills in this game:';
           isFirstInteraction = false;
         }
-        skillDesc += '\n * ' + otherSkill + ': ' + interactDesc;
+        skillDesc += '<br/> * ' + otherSkill + ': ' + interactDesc;
       });
 
       if (isFirstDieSkill) {
@@ -1762,7 +1762,8 @@ Game.skillSpan = function(skill, skillDesc) {
   skillSpan
     .tooltip({
       content: skillDesc,
-      items: 'span'
+      items: 'span',
+      position: { my: 'bottom', at: 'top-5', collision: 'flipfit'},
     })
     .on('click', function() {
       // james: without closing other tooltips, it's possible that
--- a/src/ui/js/Game.js
+++ b/src/ui/js/Game.js
@@ -1680,72 +1680,100 @@ Game.pageAddGameNavigationFooter = function() {
 
 // Display a footer-style message with the list of skills in this game
 Game.pageAddSkillListFooter = function() {
-  var gameSkillDiv = $('<div>', {
+  var hasButtonSpecial = false;
+  var hasDieSkill = false;
+  var isFirstButtonSpecial = true;
+  var isFirstDieSkill = true;
+  var isFirstInteraction;
+  var skillDesc;
+
+  var buttonSpecialDiv = $('<div>', {
+    'text': 'Button specials in this game: ',
+  });
+  var dieSkillDiv = $('<div>', {
     'text': 'Skills in this game: ',
   });
 
-  var firstSkill = true;
-  var firstInteract;
-  var skillDesc;
-  var indivSkillSpan;
   $.each(Api.game.gameSkillsInfo, function(skill, info) {
     skillDesc = skill;
+
     if (info.code) {
+      // add a die skill ...
       skill += ' (' + info.code + ')';
       skillDesc += ' (' + info.code + ')';
-    }
-    skillDesc += ': ' + info.description;
+      skillDesc += ': ' + info.description;
 
-    firstInteract = true;
-    $.each(info.interacts, function(otherSkill, interactDesc) {
-      if (firstInteract) {
-        skillDesc += '\n\nInteraction with other skills in this game:';
-      }
-      skillDesc += '\n * ' + otherSkill + ': ' + interactDesc;
-    });
-
-    if (!(firstSkill)) {
-      gameSkillDiv.append('&nbsp;&nbsp;');
-    }
-
-    indivSkillSpan = $('<span>');
-
-    indivSkillSpan.append(
-      $('<span>', {
-        'text': skill,
-        'class': 'skill_desc',
-      })
-    );
-    indivSkillSpan.append($('<span>', {
-      'text': 'i',
-      'class': 'info_icon',
-    }));
-    indivSkillSpan
-      .tooltip({
-        content: skillDesc,
-        items: 'span'
-      })
-      .on('click', function() {
-        // james: without closing other tooltips, it's possible that
-        // the same tooltip might occur twice on a browser page due
-        // to the combination of hover and click
-        $('.ui-tooltip').hide();
-        $(this).tooltip('open');
+      // ... with interactions
+      isFirstInteraction = true;
+      $.each(info.interacts, function(otherSkill, interactDesc) {
+        if (isFirstInteraction) {
+          skillDesc += '\n\nInteraction with other skills in this game:';
+          isFirstInteraction = false;
+        }
+        skillDesc += '\n * ' + otherSkill + ': ' + interactDesc;
       });
 
-    gameSkillDiv.append(indivSkillSpan);
+      if (isFirstDieSkill) {
+        hasDieSkill = true;
+        isFirstDieSkill = false;
+      } else {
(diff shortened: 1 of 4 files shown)


src/ui/js/Game.js
```diff
@@ -1680,53 +1680,102 @@ Game.pageAddGameNavigationFooter = function() {
 
 // Display a footer-style message with the list of skills in this game
 Game.pageAddSkillListFooter = function() {
-  var gameSkillDiv = $('<div>', {
+  var hasButtonSpecial = false;
+  var hasDieSkill = false;
+  var isFirstButtonSpecial = true;
+  var isFirstDieSkill = true;
+  var isFirstInteraction;
```
cgolubi1: I kind of have a pet peeve about this isFirstWhatever logic style --- it seems to me that having an "if info.interacts.length > 0" block before the loop accomplishes what you want here, and is much tidier.

src/ui/js/Game.js
```diff
@@ -1680,53 +1680,102 @@ Game.pageAddGameNavigationFooter = function() {
...
 
-    firstInteract = true;
-    $.each(info.interacts, function(otherSkill, interactDesc) {
-      if (firstInteract) {
-        skillDesc += '\n\nInteraction with other skills in this game:';
+      if (isFirstDieSkill) {
+        hasDieSkill = true;
+        isFirstDieSkill = false;
+      } else {
+        dieSkillDiv.append('&nbsp;&nbsp;');
```
cgolubi1: Similarly, here, you can collect all of your descriptions into an array of descriptions, and then use `.join('&nbsp;&nbsp;')` to combine them.  See the assembly of `capturedDieText` in `Game.gamePlayerStatus()` for an example.

test/src/ui/js/test_Game.js
```diff
@@ -1136,19 +1136,33 @@ test("test_Game.pageAddSkillListFooter", function(assert) {
   });
 });
 
-test("test_Game.pageAddLogFooter", function(assert) {
+test("test_Game.pageAddSkillListFooter", function(assert) {
```
cgolubi1: This diff is weird looking, so i may be confused about what's going on here, but did you delete the original "test_Game.pageAddLogFooter" test and replace it with this new test, instead of adding the new test and retaining the old one?  If so, why?

src/ui/js/Game.js
cgolubi1: Huh, i guess the old code did it this way too.  `:>(`  Well, ick, but i guess i can't fault you for retaining that.

cgolubi1:
Okay, so, mostly, i think what i need to do is bring this up on a test site and see how it looks, and whether it seems to implement what shadowshade says it does.  :>)

cgolubi1:
So, here's the visual without any mouseovers selected:
specials_and_skills

My feeling is, we fought long and hard back in the day to keep the game part of the game mat from getting too tall, so that more lines of action log and chat log would be visible in people's browsers.

I think it's a slightly bad idea to split up button specials from game skills in the first place --- i'd prefer we keep them together, and i don't really understand the argument for why it makes the game better to split them up.

But i think it's a really bad idea to add vertical lines in order to do it.

If multiple people care enough to tell me they disagree, and think the visual i've pasted here is better than what we have right now, that's fine, and i'll concede.  But i am not enthusiastic about the look-and-feel change here (to how the visuals look when skills are not highlighted).

test/src/ui/js/test_Game.js
blackshadowshade: Huh. I'm pretty sure that my intent was to add the test, but I must have forgotten to duplicate the original pageAddLogFooter test. My bad. I'll add that back in.

blackshadowshade:
I'll let @irilyth comment on the reasoning for separating the two types of skills.

src/ui/js/Game.js
blackshadowshade: I'm happy to refactor this to use a .join().

cgolubi1:
In firefox, the click behavior is weird.
* When i hover over the link (which i have to in order to click it), the hover text shows up.
* When i click (with the hover text showing), the hover text blinks, and then reappears.
* When i click again, the hover text disappears.
* When i keep clicking, the hover text stays disappeared.

How many of those things are intentional?

blackshadowshade:
From my point of view, the first two are intentional. I tried playing around with the JQueryUI a little bit, but I couldn't get the other two steps to disappear.

cgolubi1:
> From my point of view, the first two are intentional. I tried playing around with the JQueryUI a little bit, but I couldn't get the other two steps to disappear.

Hmm... philosophically, i'd rather the UI do fewer things not brokenly, than do more things brokenly.  But that's certainly something we can discuss, and if it's a big improvement to the tablet experience, that could be good.  But if we can, i'd rather not make the experience worse on browser.

cgolubi1:
The other thing is, even when i click on the tooltip, the hover text still goes away when i mouse away from the skill.  So i feel like that's not going to get what your anonymous requester was asking for, because the text will still disappear on them when they move the mouse.

cgolubi1:
Hmm, also, wrt the new UI look, one thing i notice is that there's a blank line *between* "Button specials" and "Skills" - do we need that blank line, or was it just expedience?

blackshadowshade:
I added the blank line because without it, to my eye, things looked rather cluttered.

blackshadowshade:
I think that the experience is no worse on browser, because currently, we only have activity on-hover.

The click stuff is really intended for the case where hover doesn't work, for some reason, including:
- the browser doesn't support it correctly
- the player is using a touchscreen

irilyth:
So, hmm. I agree that not stretching the page vertically is good, but I think separating button abilities from die skills is valuable, mostly because it's otherwise not very obvious that a button even *has* a special ability, which I think can be really important. Especially e.g. if you're the *opponent* of someone like TFS, and knowing that TFS can't make Skill Attacks is really important to know for strategy while you're playing, and you might not think to hover over the TheFlyingSquirrel "skill".

But, I think the extra blank line doesn't add much, and deleting that would be fine. Maybe also putting the words "Die skills in this game" and "Button special abilities in this game" in bold or something?

Another idea: It'd be even more obvious that a button had a special ability if it was ner the button image and recipe on the game mat. If we had something there that said "Note! The Flying Squirrel has a button special ability", then I think it would matter less that the text under the game is more cluttered, because the note in the mat would catch your eye in a very obvious way. That exact text seems too bulky, but I bet we can come up with some shorter text, or an icon, or something.

So, I think we should figure out how to add something to the mat near the button art and recipe, and delete the blank line. And if we do add something to the mat, maybe even just have the skills and abilities on the same line -- just separating them with "Skills: blah blah; Abiilities; blah" would be good in that case.

blackshadowshade:
So, put a tooltip on the button link inside the playmat if there's a button special?

blackshadowshade:
No wait, that won't work on tablets or when there's no hover.

jbscare:
I think a more distinctive icon than a tooltip would be good, something like a small version of https://cdn3.iconfinder.com/data/icons/picons-weather/57/53_warning-512.png or something. Maybe that after the recipe, and some text underneath, if a fourth line of text would fit ok there? "Note! Button special ability (details below)" or something.

cgolubi1:
> I added the blank line because without it, to my eye, things looked rather cluttered.

Okay, i'd say take another go at this, and see if something like boldfacing can remove the need for this line --- for me, it's that much more action log text which i can't see on my screen by default, and that's a pretty high cost for blank space.

irilyth:
I think a more distinctive icon than a tooltip would be good, something like a small version of https://cdn3.iconfinder.com/data/icons/picons-weather/57/53_warning-512.png or something. Maybe that after the recipe, and some text underneath, if a fourth line of text would fit ok there? "Note! Button special ability (details below)" or something.

blackshadowshade:
If vertical space is the problem, what about getting rid of half the space between the playmat and the "Go to your next pending game" link? (as well as losing the new blank line)

cgolubi1:
> If vertical space is the problem, what about getting rid of half the space between the playmat and the "Go to your next pending game" link? (as well as losing the new blank line)

Well, that might be okay, but fundamentally, i think that button skill descriptions and game skill descriptions are far more similar to each other than they are to any other item on the play mat.  So, i guess i question the assertion that they can't be on consecutive lines.  Can we really not make that work visually?

irilyth:
> If vertical space is the problem, what about getting rid of half the space between the playmat and the "Go to your next pending game" link?

I think that'd be worth doing (or at least looking at to make sure it doesn't look bad, but it sounds to me like it'd look good) independent of what we do about the skills-in-this-game section.

irilyth:
Various things about clicking on skills:

- It seems to me like it'd be ideal if clicking on the link didn't make the hover text vanish and reappear, but that's probably not too big a deal. (Just based on Chaos's explanation.)

- My expectation would be that clicking once would cause a persistent box to pop up, and clicking a second time would cause it to vanish, and a third time would pop it back up again. In particular, I think it shouldn't vanish if you just move the mouse away; and if you do vanish it and then click again, it should pop up again.

- It's true that the current site doesn't do anything when you click on a skill name, but when I want to look soemthing up, I frequently click on it anyway -- so if clicking on it and moving the mouse will cause the box to pop up, and then vanish, and then *nothing else I do causes it to come back*, that's definitely worse for mouse-based browsers than the current situation.

cgolubi1:
Incidentally, now that i've looked at this and thus can try to iterate on it quickly, i'd like to suggest that it be split into multiple pulls --- probably dividing on the lines of the bullet points you put in the description would work well.  I see a number of different concerns with different pieces of it (do we like the split of die skills / button specials, does the new clicking behavior work, do we like the bigger mouseover font, etc), and i think it will be very painful to try to adjudicate all of those at the same time, whereas if you split them up, then where we *can* make progress, we can make it relatively quickly.

In particular, i imagine:
* Die skill codes are pretty non-controversial.  That pull could probably be merged immediately
* We need to argue about information vs whitespace for die skills / button specials, but we don't need any new tools to do it --- we just need to have the argument and pick something
* The other two, we'll need to talk about browser support and technological options a bit

blackshadowshade:
Okay, I'll get a pull request in for you combining the first two points, since I really don't want to have to deconflict the spacing pull request when the skill code one is accepted. I realise that this may delay the pull request a little, but I'm fine with that.

Once that is accepted, then I'll put up a second pull request that we can use as a starting point for the skill hints.

blackshadowshade:
I'm closing this pull request, in favour of #1799.

PR #1742, opened 2015-10-23
"New game" shortcuts that preserve chat now dismiss the previous game

### LINKED ISSUE #1657: Automatically dismiss the current game when creating a follow-up game
When I use the buttons to create a follow-up game (the ones after "Challenge X to another game, preserving chat" on the completed-game page), I invariably want to dismiss the original game immediately afterwards. (If I ever want to look at it again, I can always find it in the follow-up game's chat log.) It'd be cool if this happened automatically; perhaps as a user preference, if we're not confident that everyone wants this.

blackshadowshade:
Fixes #1657.

Tested with 1-round Haruspex vs Haruspex matches.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/917/ (if it passes)

### DIFF
--- a/src/api/ApiResponder.php
+++ b/src/api/ApiResponder.php
@@ -88,6 +88,7 @@ class ApiResponder {
         }
         if (isset($args['previousGameId'])) {
             $previousGameId = $args['previousGameId'];
+            $interface->dismiss_game($_SESSION['user_id'], $args['previousGameId']);
         } else {
             $previousGameId = NULL;
         }
--- a/src/ui/js/Game.js
+++ b/src/ui/js/Game.js
@@ -1740,7 +1740,7 @@ Game.pageAddNewGameLinkFooter = function() {
     Game.page.append($('<div>', {
       'text':
         'Challenge ' + Api.game.opponent.playerName +
-        ' to another game, preserving chat:',
+        ' to another game, preserving chat, and dismiss this game:',
     }));
 
     linkDiv = $('<div>');


cgolubi1:
I'm not 100% sure the wording is the best (it might confuse people), but i also don't think that's a blocker since i don't have an explicit suggestion (i would have put "(and dismiss this game)" in parens, but i'm not actually sure that's better).  We can always improve it later, so i'm not going to sweat that.

What i do want to test:
* See how the UI looks visually and if i have any issues there
* Make sure the functionality works as intended for me with various continuation game types
* Look at the code in BMInterface and see if i can see a way to break this by submitting invalid game parameters, so that the old game gets dismissed but the new game doesn't get created.

blackshadowshade:
So, independent of whether this actually solves the issue, irilyth has raised the problem that this fix may actually conflict logic-wise with the game rejection paradigm. We'll put this on hold for the moment.

irilyth:
Not the paradigm, the current implementation -- unless you think that "rejected games should silently vanish" is part of the paradigm.

cgolubi1:
So IMO with #1749 there is no longer a conflict between this pull and irilyth's game workflow, and i should test it again.  irilyth, please weigh in on that when you've had a chance to look at #1749.

irilyth:
Blargh, I failed to note that this was waiting for input from me, so I suppose I have only myself to blame for not signing off on my own BFOTM. :^p  Oh well, next month.

I agree that #1749 now makes this fix non-workflow-breaking for me; please merge it. Thanks!

cgolubi1:
Okay, so, from a cleanliness/safety point of view, i don't like the fact that dismiss is attempted, and its failure or success is totally ignored.  In particular, if the game has already been dismissed, `dismiss_game()` fails with a message "You have already dismissed game N" --- this is harmless, but what if dismiss_game failed with some less harmless error?  This is really multiple related issues:
* The code assumes that attempting to dismiss a game and failing because it was already dismissed is totally harmless (has no side effects).  That's true now, but what if it became less true later?
* The code doesn't differentiate between attempting to dismiss a game which was already dismissed (not an error because of the point above) and some other error which is actually unexpected.
* Errors are not propagated to the player at all.

cgolubi1:
shadowshade: i said in my test plan:

> see if i can see a way to break this by submitting invalid game parameters, so that the old game gets dismissed but the new game doesn't get created.

In future, if you get bored waiting for me to approve pulls, read the parts of my test plans where i say "let's see if shadowshade thought of X" and decide whether you thought of X.  :>)

Anyway, spoilers: you did not.  You dismiss the game (and ignore the return value, natch, but that's not relevant here) before you create the new game.  So if game creation fails (and you can reproduce this trivially via the UI by changing the opponent, which will make game creation fail in a followup game), the old game is dismissed anyway.

So:
* Create the new game before dismissing the old game
* Only dismiss the old game if creating the new game succeeded
* Have dismiss_game differentiate between failing because the game was already dismissed and some other reason (i suggest passing a flag to dismiss_game which says that it's harmless if the game has already been dismissed, and returning true in that case if that flag was passed --- among other things, that encodes in dismiss_game our expectation that dismissing a game should be idempotent)
* Construct the return success code and message intentionally based on the combination of create game's result and dismiss_game's result (i don't feel too strongly about what you do here, just do it on purpose)

blackshadowshade:
Okay, I must admit to being out of my depth here. I don't know how to implement what you're asking.

AdmiralJota:
Maybe something like these?

https://www.dropbox.com/s/znybpta8pn0txv3/ApiResponder.php?dl=0
https://www.dropbox.com/s/ltgre96x41ms8kk/BMInterface.php?dl=0

(Untested.)

blackshadowshade:
Jota, I guess my hesitation here with your proposed solution is that the fix is hard-coded into the interface, which I think is the wrong level. In my mind, the correct place to implement this behaviour is in the front-end, so that someone coming along and writing a different UI isn't bound to the same behaviour.

My problem with doing everything in the front-end, though, is that I don't really understand callbacks as well as I should.

cgolubi1:
On Sun, 08 Nov, 2015 at 12:39:34 -0800, James Ong wrote:

> Jota, I guess my hesitation here with your proposed solution is that the fix is hard-coded into the interface, which I think is the wrong level. In my mind, the correct place to implement this behaviour is in the front-end, so that someone coming along and writing a different UI isn't bound to the same behaviour.
> 
> My problem with doing everything in the front-end, though, is that I don't really understand callbacks as well as I should.

Does this have anything to do with the frontend?  Being "dismissed"
is a backend property of a game.  If you dismiss a game via the UI,
it will be dismissed too if you ask for the game via python.  This seems
like clearly API and backend functionality to me.

Chaos

blackshadowshade:
I think it does. What we're doing is putting a link there on the page that does multiple things: create a game and dismiss a game. The very fact that the link does that doesn't necessarily mean that creation of a follow-up game should always dismiss the previous game in every possible implementation of the UI.

irilyth:
Why wouldn't you think that creating a follow-up game would dismiss the previous game in every implementation of the UI?

I feel like "creating a follow-up game always dismisses the previous game" is exactly what we want, and not a UI thing at all.

cgolubi1:
Okay, so, Jota's suggested fix is not too dissimilar to mine, but i'd be happy to see the logic of dealing with the message move into ApiResponder --- e.g. something like:
* After calling `create_game()`, if you realise you need to call dismiss_game too (and Jota's logic seems right for how to make that determination):
    * First cache `$interface->message`
    * Then call `dismiss_game()` (with the `isAncillary` flag set to true as Jota suggests - not sure that's the best name, but the extra flag seems reasonable)
    * Within `dismiss_game()`, the `isAncillary` flag only controls whether the condition in which dismiss fails because the game has already dismissed returns false or true
    * After returning from `dismiss_game()`, reassemble the message - if `dismiss_game()` returned TRUE, just reinstate the cached message.  If `dismiss_game()` returned false, assemble the message as the cached message plus a notice that dismissing the previous game failed.

I think that pretty much accomplishes the same thing as Jota's suggestion, but is a much lighter hand on `BMInterface` --- the only change there is modifying `dismiss_game()` so that it takes an optional parameter under which the return value is TRUE if the game was already dismissed, so that's more like a two-liner.

blackshadowshade:
That sounds like a reasonable compromise, and if we decide in the future that we do want to separate out "pure create game" from "create game and then dismiss game", it's simple enough to add an extra function to ApiResponder without touching anything further back.

AdmiralJota:
Is ApiResponder allowed to set the message on BMInterface? I thought that the fact that set_message was protected would prevent that. That's the only reason I did it that way, so if that is allowed, then I am all in favor of your suggestion.

blackshadowshade:
No, ApiResponder isn't allowed to set the message on BMInterface. My immediate thought is that we need to keep a clone of the BMInterface, although that may also prove to be suboptimal.

AdmiralJota:
Would a clone help with the case when we want to convey that creating the game succeeded but dismissing failed?

cgolubi1:
On Mon, 09 Nov, 2015 at 08:10:13 -0800, James Ong wrote:

> No, ApiResponder isn't allowed to set the message on BMInterface. My immediate thought is that we need to keep a clone of the BMInterface, although that may also prove to be suboptimal.

Yeah, so i'd argue two things:

1. We need a complete refactor of how we assemble API messages to hand back to the client, and a possible design choice for that is that ApiResponder should own the content of those messages.  The current implementation was basically built to get message information to the client expediently (without a lot of backend work needed to make it happen), and it's had the upshot that a lot of the time the message the customer gets doesn't make a lot of sense (e.g. the ubiquitous "Loaded data for game N" bugs).  You could imagine a situation in which:
 * Whereas right now every public BMInterface method returns obj or FALSE and sets message as an incidental property, instead public BMInterface methods should return a tuple or dict containing a return value and a message.
 * Whereas right now ApiResponder blindly passes along the message from BMInterface to the client, ApiResponder would have some ownership of the message to pass along to the client, and therefore could do things like assemble its own message where appropriate.
Or we could do something else, but my point is that i think we need to refactor API message passing, and we should open an issue for that.

2. Given the intention to tidy that above, i'd be comfortable solving this problem in the short term by creating a public BMInterface "assemble message" method, to which we could pass the various pieces of message in this case and have BMInterface assemble them (and the internal logic of the new method could be along the lines of `assemble_message(part_one, part_two, boolean)` where if the boolean is true, return "part_one" . "part_two" and if the boolean is false, just return part_one, or something like that.

blackshadowshade:
Okay, I'm closing this pull request because it apparently doesn't approach anything near what will be necessary to address the problem comprehensively.

PR #1743, opened 2015-10-23
Added game description to overview page

### LINKED ISSUE #1508: Display Description on Game Page
I'd like to see the game description on the game page (between the game # and your button columns maybe).  Because:  when I'm playing in tournaments the description shows UBFC or MadRobin, and I like to play those games either first or last.  Having the description on the game page would let me pick those out quickly.

I suppose if a description is very long, this would either look terrible or not work.

An alternative would be to have a flag column, and a way for users to flag specific games themselves, but that seems like a much more complicated feature.  And the description gives more detail.

What do you think?

blackshadowshade:
Fixes #1508.

This should show game descriptions for new games, active games, and completed games.

Tested with 1-round haruspex vs haruspex games with descriptions.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/922/

### DIFF
--- a/src/api/DummyApiResponder.php
+++ b/src/api/DummyApiResponder.php
@@ -299,6 +299,7 @@ class DummyApiResponder {
         // Use the same fake games here which were described in loadGameData
         $data = array(
             'gameIdArray' => array(),
+            'gameDescriptionArray' => array(),
             'opponentIdArray' => array(),
             'opponentNameArray' => array(),
             'myButtonNameArray' => array(),
@@ -326,6 +327,7 @@ class DummyApiResponder {
 
     protected function add_active_game_data_1(&$data) {
         $data['gameIdArray'][] = 1;
+        $data['gameDescriptionArray'][] = 'Game 1';
         $data['opponentIdArray'][] = 2;
         $data['opponentNameArray'][] = "tester2";
         $data['myButtonNameArray'][] = "Avis";
@@ -345,6 +347,7 @@ class DummyApiResponder {
 
     protected function add_active_game_data_2(&$data) {
         $data['gameIdArray'][] = 2;
+        $data['gameDescriptionArray'][] = 'Game 2';
         $data['opponentIdArray'][] = 2;
         $data['opponentNameArray'][] = "tester2";
         $data['myButtonNameArray'][] = "Avis";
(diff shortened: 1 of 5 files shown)


cgolubi1:
Okay, the task here is just to bring this up on a test site and see how it looks.  I'm a little worried about the timing for November if there's any controversy around how it renders for e.g. games with long descriptions, but i'll take a look and post a screencap and see where we are.

cgolubi1:
One comment: in the "Open games" page, and in the game itself, the description is italicized.  I think it should be italicized on the overview too, for consistency.

cgolubi1:
Okay, so it's not terrible with long descriptions.  I really do think we need the italics, though:

description_screencap

cgolubi1:
Also, not sure how others feel, but my preference would be to have the description column later, after the basic game info columns - maybe i'd order them: Game #, Your Button, Opponent's Button, Opponent, Description, Score, Inactivity?

cgolubi1:
My third comment is that i do think you need to solve the long description problem --- i know long descriptions can look ugly on the open games page, but the thing is the overview page is used frequently and its purpose is to be concise. So i don't think this visual passes muster.  My druthers would be to take a description substring of a *short* number of characters (something like 30) and replace everything after that with ..., because my druthers would be to have all the descriptions fit on one line to keep the overview compact:

description_with_long_game

cgolubi1:
Okay, in summary:
* Italicize the description for consistency with the rest of the site
* I suggest putting the columns in a different order, so the (possibly blank) description field isn't the first thing
* Include a substring of the description, not the whole thing

blackshadowshade:
So, these changes should now be addressed.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/923/ (if it passes)

cgolubi1:
Okay.  This has become a prod blocker because #1749, which is a prod blocker, depends on it.  So let's bring it up and hope the issues above have been addressed.

cgolubi1:
Good.  I agree that the bugs are fixed.  I bet there's some spiffy way to dynamically decide how much text to show based on the player's browser width so the column won't wrap, but for now, let's go with this.  Merging.

PR #1744, opened 2015-10-23
Set an explicit API message when joining an open game

### LINKED ISSUE #1274: API should return a reasonable message for the joinOpenGame API method
When you join an open game, the API returns a message of e.g. "Loaded data for game 3.Loaded data for game 3.".  This should be something user-friendly like "Successfully joined game 3" instead.

blackshadowshade:
Fixes #1274.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/919/

### DIFF
--- a/src/engine/BMInterface.php
+++ b/src/engine/BMInterface.php
@@ -3129,6 +3129,7 @@ class BMInterface {
             $game = $this->load_game($gameId);
             $game->hasPlayerAcceptedGameArray[$emptyPlayerIdx] = TRUE;
             $this->save_game($game);
+            $this->set_message('Successfully joined game ' . $gameId);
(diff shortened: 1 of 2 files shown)


cgolubi1:
Easy enough.  I'll bring it up in a test site and make sure it looks visually fine when a player joins an open game.

cgolubi1:
Yeah, seems okay.  The UI doesn't give you a message from the server at all when you join an open game, so it's not any different as a result of this.

cgolubi1:
The UI not giving feedback is a real bug (and one which would matter if joining the open game failed), so opened #1755 for that.  Meanwhile, merging this.

PR #1762, opened 2015-11-03
Added insult die skill

### LINKED ISSUE #359: Insult / Parry / Japanese Beetle Die Skill needed
Insult: I 
Can not be taken by Skill attacks.

Insult from Gaming Guardians, Parry from Blademasters, and the Button Special for Japanese Beetle are all the same:

blackshadowshade:
Fixes #359.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/929/ (if it passes)

### DIFF
--- a/src/engine/BMAttackSkill.php
+++ b/src/engine/BMAttackSkill.php
@@ -360,7 +360,13 @@ class BMAttackSkill extends BMAttack {
         }
 
         $def = $defArray[0];
-    
+
+        if ($def->has_skill('Insult')) {
+            $this->validationMessage =
+                'Dice with the Insult skill cannot be attacked by skill attacks.';
+            return FALSE;
+        }
+
         if ($def->ownerObject instanceof BMGame) {
             $ownerButton = $def->ownerObject->buttonArray[$def->playerIdx];
             if ($ownerButton->has_skill('TheJapaneseBeetle')) {
--- a/src/engine/BMSkill.php
+++ b/src/engine/BMSkill.php
@@ -248,7 +248,8 @@ class BMSkill {
                      'BMSkillNull',
                      'BMSkillMaximum',
                      'BMSkillTimeAndSpace',
-                     'BMSkillWarrior');
+                     'BMSkillWarrior',
+                     'BMSkillInsult');
         // fires last
     }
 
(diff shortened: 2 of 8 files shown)


cgolubi1:
So... how do we feel about stealth insult dice?  (For those who need a reminder and don't feel like reading the source code, insult dice cannot be attacked by skill attacks.)  I guess our practice up to this point has been not to disallow pathological combinations of skills in RandomBMMixed, but to handwavily expect that when we start allowing custom recipes which might wind up on the prod site, there will be combinations of skills we will disallow.

Invincible dice strike me as fairly pathological as pathological things go.  I'm provisionally okay with the above (`dI` will only be able to come up in random BMs, and people won't be able to choose it intentionally), but i wanted to flag it explicitly.

cgolubi1:
In terms of testing, this doesn't seem very complicated, honestly.  I could do some regression tests, but the only game logic you changed is an explicit check for the defender having the `Insult` skill, so i don't see how that's going to come up in games without the `Insult` skill.

There are 10 buttons with Insult which will be playable after this update.

So, testing, probably something like:
* Play games by hand involving the 10 now-playable buttons (EDG, Arashi, Johnny, Renee, Danielle, Scott, Magistra, Inez, Jean-Paul, Chantal) and make sure they all look visually okay since we haven't seen them before
* Have RandomAI play games involving the Insult buttons against other skills, see if anything blows up, and review some of the games to make sure they look okay
* Have RandomAI play a bunch of RandomBMMixed games, and spot-check the pairing of Insult with other skills (i do want to make sure `dI` doesn't lead to internal errors, pathological or not

cgolubi1:
Okay, as noted above, i don't see any issues here.  This can be merged when we're out of merge freeze.

cgolubi1:
Merging.

irilyth:
We talked about this a bit on IRC, and I think the consensus was that Stealth Insult dice are very powerful, but don't break the game, so let's not ban them in RandomBM Mixed (which frequently ends up with very imbalanced buttons (strong or weak) anyway), but obviously we'd want to keep them in mind when thinking about proposed new buttons. (And one could imagine someone making a case for a button having an invincible die, if it had some other tradeoffs at a button level, just like any other skills.)

PR #1799, opened 2015-11-09
Added skill codes to in-game skill help, and separated button specials

blackshadowshade:
Addresses part of #1697.

Tested with Giant vs skapheles.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/945/


blackshadowshade:
So, I removed the line between Button Specials and Die Skills, as requested in #1741. To my eye, the section now looks overly crowded, but I will defer to consensus opinion:

screenshot

cgolubi1:
Hmm.  I want to flag AdmiralJota to have an opinion about the visuals of the two lines --- my first instinct is to wonder if changing the centering so that both lines are left-justified within a centered box would help.

In particular, it feels to me like the problem is the long introductory string on each line blending into the list of skills, so that makes me want to try something like either putting them in a table (for which left-justifying would be a proxy) or using coloring or styles in some fashion to distinguish the intro strings from the list of skills.

Jota has a really good instinct for this kind of visual, so i'm going say his name a couple more times and see if he offers an opinion.  :>)

blackshadowshade:
Did you already mention @AdmiralJota? Or did I need to mention Jota's name a couple more times?

AdmiralJota:
I see what you mean about the intro string blending into the list, although I fear that left justifying them might tend to make things look lop-sided, with the much shorter button specials being over to the left while the die skills stretch all the way across (in this case, anyway).

Maybe splitting them side-by-side, like we do with turn logs and chat logs?

Something like this (window intentionally narrowed to force it to wrap):

bm1

Or maybe this:

bm2

(That second one may need some awkward HTML/CSS to get it to look right in the general case -- although on the other hand, maybe it's no worse than what the turn/chat tables use.)

These would both have the added benefit that if there were fewer skills, then they'd fit on one line. (And of course, if there were *only* button specials or *only* die skills, then we could just center whichever one was there.)

AdmiralJota:
Do I need to explicitly mention @cgolubi1 my comment in order to find out how well my suggestions were received?

irilyth:
I sort of like the two-column approach, although since most buttons don't have special abilities, I predict that it'd be a little jarring when the customary one centered line got split into two. On the other hand, that would work towards the goal of making it super extra bonus obvious when a button *does* have a special ability, so maybe that's a good thing.

cgolubi1:
Hmm.  Well, it does make it more obvious which is which without taking up more vertical space.  So that's a bonus.  I guess i just don't like these long tags and the repetition of "in this game".  I wonder if something like:
```
In this game: Button specials: Giant
              Die skills: Focus Poison Reserve Shadow Speed Stealth Stinger TimeAndSpace
```

would work.

blackshadowshade:
We could drop "In this game" in its entirety, I reckon.

cgolubi1:
So, first off, i'd be in favor of trying dropping "in this game".

Meanwhile, i thought of something.  I wonder if this would be nifty, or if it would be too complicated and/or look ugly.

One of the things that's been bugging me is that there's no visual differentiation between the tag "Button specials in this game:" and the values.  I keep wondering about whether we should use color somehow, but not having a suggestion about how to do so.  So here's one --- what if we somehow background-colored or outlined each skill name (e.g. made an oval around it) *based on the player color of the player whose button had that skill*?  Offhand, i really like the idea because it solves the differentiation problem and conveys extra information using a signaling mechanism we already use (the distribution of colors to players).

Unsolved problems:
* The API doesn't actually provide the information about which skills go to which player.  (So if we decide to do this, we'd probably wind up deferring actually doing it to another issue.  But i think we should go ahead and think about whether we like the idea.)
* What visual element do we want here?  I *think* it could be as simple as a background oval of the color (because we already put all the game mat text on these colors, so implicitly we already think that's legible).
* What do we do if both buttons have a particular skill?  The obvious answer is splitting the oval in half and having half the background be in each color, but i bet that's not trivial to whip up in CSS.

What do y'all think?  Is this idea worth considering?

blackshadowshade:
If you're just looking to distinguish the title from the values, bold-facing might be enough.

cgolubi1:
> If you're just looking to distinguish the title from the values, bold-facing might be enough.

Yeah, if you want a short-term answer, do that.  I'm curious what people think about the more ambitious answer, but for now, could you mock up and show me the two-line version you have earlier in this post with:
* The labels boldfaced
* "in this game" dropped

blackshadowshade:
Screenshot as requested:

screenshot

Looking at that, I think that boldfacing doesn't do what I expected it to do.

dwvanstone:
I would suggest simply adding a few pixels (5?) between the two lines. That may be enough separation.

It's visually confusing for me because of the dark blue circles (which I assume indicate there's further info in pop-up text). I don't know if those dark blue circles are necessary and may clean up the appearance.

irilyth:
I've also never liked the dark blue circles, FWIW.

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/947/ corresponds to the version without "in this game" and with boldfacing

AdmiralJota:
I don't think anyone felt the dark blue circles were an aesthetic improvement over not having them, but I think they were determined to be the least bad way to try communicate to the user that it was worth hovering over them.

If we get a click-to-popup thing working nicely some day, then we can probably remove them and just keep the dotted underline. Also, it occurs to me that adding a CSS style of "cursor: help;" probably wouldn't hurt either.

dwvanstone:
I think the circles make it harder to parse the sentence visually. The dotted underline is nice and seems to communicate that there's something there. I do believe that the circles communicate "hover" more effectively, but the drawback is that it's visually more confusing. However, this seems to be far away from this particular issue. 

The reason why it was related to this issue is to articulate why I thought having the two lines close together made things look crowded and I was suggesting a way to aleviate the crowdedness.

blackshadowshade:
What about we lose the blue info circles then when we bring in click behaviour?

AdmiralJota:
Sounds good to me.

blackshadowshade:
In that case, do we still need on-hover behaviour, or do we make the behaviour exclusively on-click?

cgolubi1:
I don't want to discourage people chatting about hover vs. click, but just to note that i think we're not doing anything with click/hover in this pull.

cgolubi1:
So... i think the version with the boldfacing and without "in this game" is better than the previous version.  I agree that it's still a little crowded looking --- i like dwvanstone's idea of adding a couple of pixels of space, but not an entire blank line.

blackshadowshade:
For those in the know with CSS, is this just a matter of manipulating the line-height property?

blackshadowshade:
capture

Here's a screenshot with extra line height.

cgolubi1:
Yes.  I think that gets the job done.  Any objections?

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/950/ (if it passes)

blackshadowshade:
Regarding the colour idea, what happens when a player captures a die with a skill? Does the skill colouring change then?

cgolubi1:
> Regarding the colour idea, what happens when a player captures a die with a skill? Does the skill colouring change then?

I think not --- i think the informational text is about what skills each button brought into the game, not necessarily about the current state of dice.

AdmiralJota:
For future reference, you can also manipulate spacing around/between things using the padding and margin properties (e.g. "padding-top: 5px;") (Padding goes inside the element; margin goes outside. That can be important if you have things like borders and backgrounds around things.).

And I agree with Chaos about the coloring. On the topic of what to do if both buttons have the same skill, a second way to deal with that would be to list that skill twice, once with each color (maybe even grouping the ones for each button together). Easier CSS, but maybe weirder logic.

dwvanstone:
Yes, I think that extra space did the trick!

cgolubi1:
Based on the discussion, i think we ran this one to ground and it was ready to merge.

cgolubi1:
Hmm, though, actually, that's a lie --- i think we did a bunch of discussion based on screencaps that shadowshade posted, but i didn't test the final result myself.  So that should be pretty straightforward, and i'll work up a test plan this morning, but i'm not going to quickly merge it before the emergency dev release.

cgolubi1:
Okay, so, basically this pull is now simply a refactor of how the skill interactions look, so my test plan is just to bring this up in vagrant and look at the game mat for:
* A game with no skills
* A game with die skills only
* A game with button skills only
* A game with die skills and button skills
* A game with skills which interact
* A game in a state other than active

See if it looks like what we've been discussing on this pull, and, if so, merge it.

cgolubi1:
Okay, so, the game with no skills case is broken:
game_with_no_skills

cgolubi1:
See e.g. http://www.buttonweavers.com/ui/game.html?game=8591 for how the no skills case looks right now.  I think something analogous to that is just fine.

cgolubi1:
A second problem i notice here - in a new game with RandomBMMixed, RandomBMMixed doesn't show up as one of the skills.  Other button specials do show up in a new game.  That's not a regression, though, so i'm going to make a separate issue for it.

cgolubi1:
Okay.  So you have to fix the no skills case, and then we can merge this.

blackshadowshade:
Okay, this should be fixed.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1002/ (if it passes)

cgolubi1:
Okay.

Issue #1608, opened 2015-04-28
BMGame has too many properties


blackshadowshade:
BMGame is becoming more and more complex, and it seems sensible to refactor some of the structure into a new class BMPlayer, as was proposed right back at the beginning, when we were discussing the structure of the Buttonmen code.

Issue #712, opened 2014-03-13
all number inputs to the backend should be cast to int immediately


cgolubi1:
James and i talked about this in the context of pull #707: we like the idea of having the backend use integers in integer format exclusively, and provide integers to the frontend via JSON.  That means that, whenever we bring in numbers in string form, we should cast them to integer right away.  So:
- Whenever numbers are read from the database, cast them to (int) right after fetching them
- Whenever numbers are submitted via POST, and validated via ApiSpec::verify_argument_of_type_number(), they should also be cast to int if they are strings
- Whenever numbers are read from session variables, they should be cast to int
- Whenever numbers are generated in the code, they should be generated as integers, not as strings
- We should not need or have casts to int in places in the code other than the above

Make it so.

PR #1809, opened 2015-11-11
Start moving some game properties to new class BMPlayer

blackshadowshade:
Begins to address issue #1608.

I've moved responsibility for playerId, button, activeDieArray, capturedDieArray, outOfPlayDieArray, waitingOnAction, and isPrevRoundWinner to BMPlayer. To remove the corresponding properties completely from BMGame, I have altered the generic getter __get() to allow access to arrays of BMPlayer properties.

The intent is to alter the generic setter __set() in a similar way, and then to shift the responsibility for more properties across to BMPlayer.

This pull request is a pure refactor, and should not change any user-visible behaviour whatsoever.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/949/


blackshadowshade:
So, I've progressed this by altering the generic setter also.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/974/ (if it passes)

src/engine/BMPlayer.php
```diff
@@ -0,0 +1,264 @@
...
+
+
+    // setter methods
+    /**
+     * Set ID
+     *
+     * @param int|null $value
+     */
+    protected function set__playerId($value) {
+        if (!is_null($value)) {
```
cgolubi1: For consistency, shouldn't you throw an InvalidArgumentException if this isn't an int too?

src/engine/BMPlayer.php
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+    /**
+     * Getter
+     *
+     * @param string $property
+     * @return mixed
+     */
+    public function __get($property) {
+        $funcName = 'get__'.$property;
```
cgolubi1: I just dinged dwvanstone on horizontal whitespace in a different pull, so i feel i should mention it here too.  :>)

test/src/engine/BMGameTest.php
```diff
@@ -1732,9 +1728,9 @@ public function test_update_game_state_end_turn() {
                                                    array($die2));
         $this->object->nRecentPasses = 0;
         $this->object->gameState = BMGameState::END_TURN;
+        $this->object->activePlayerIdx = 0;
         $this->object->update_game_state();
         $this->assertEquals(BMGameState::START_TURN, $this->object->gameState);
-        $this->assertTrue(isset($this->object->activeDieArrayArray));
```
cgolubi1: For my understanding, why remove these asserts instead of making them trigger on `$this->object->player[0]->activeDieArray`?

test/src/engine/BMPlayerTest.php
```diff
@@ -0,0 +1,32 @@
...
+
+    /**
+     * Tears down the fixture, for example, closes a network connection.
+     * This method is called after a test is executed.
+     */
+    protected function tearDown() {
+
+    }
+
+    public function test_dummp() {
```
cgolubi1: You should probably go so far as to call this `test_dummy`...

But also don't you need some real tests of your new class?

cgolubi1:
I'm having to look at the diff of BMGame.php using the CLI, so i can't give you line-by-line comments with github, but i see:
```diff
@@ -256,12 +226,12 @@ class BMGame {
      */
     protected $gameState;
 
-    /**
-     * Boolean array whether each player needs to perform an action
-     *
-     * @var array
-     */
-    protected $waitingOnActionArray;
+//    /**
+//     * Boolean array whether each player needs to perform an action
+//     *
+//     * @var array
+//     */
+//    protected $waitingOnActionArray;
```

You probably want to remove that commented-out block.

cgolubi1:
So, this is probably me not understanding this refactor, but i don't get why you still need:
```php
 * @property      array $activeDieArrayArray     Active dice for all players
```
at line 58 of BMGame.php --- in particular, i don't see what in BMGame sets that, but it seems that BMInterface and various die skills still use it.  So how does this work?

cgolubi1:
Okay, for the most part, i'm going to take this refactor on faith as usual as long as it doesn't change the outward behavior.  But i wanted to go ahead and ask a few questions about the nuts and bolts (as above).

Test plan:
* testing against `loadGameData` by doing a lot of RandomAI regression tests is the obvious thing, and given the number of changes, i'd like to throw some thousands of games against this
* i'd like to play some new games, in particular it looks like the BMDieTwin value-setting code was modified, so i want to play a bunch of new RandomAI games with twin dice to look for internal errors
* look at the code some more for any places where BMInterface accesses player information *other* than via `loadGameData`, and devise some regression tests against those.
* bring up a database from an active site and poke around in the UI to see if anything looks visually off (e.g. counts of games waiting for action)

cgolubi1:
Suggestion: add a test or two explicitly covering the new BMGame getter functionality for *Array objects --- e.g. instantiate a game and a couple of players with buttons, then call `$game->getActiveDieArray` and make sure it has in there what you put in there.

That may well be covered by other tests, but explicitly testing the getters serves as documentation of a sort of how the new code behaves.

test/src/engine/BMGameTest.php
blackshadowshade: I guess this is just because I don't see any real value in maintaining these tests. If something is going to break, I'll see that anyway, and having this particular test break won't actually tell me more than something's broken.

src/engine/BMPlayer.php
blackshadowshade: I wonder if I'll actually break something if I enforce this to be an int, considering PHP's weak typing.  :)  I'll take a look.

blackshadowshade: Okay, enforcing to an int did actually break things, and the fix will help to address #712.

blackshadowshade:
I think that I've addressed all the issues mentioned on this pull request.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/996/ (if it passes)

cgolubi1:
Yeah, i'm fine with those changes.  Now to do some actual regression-testing per above.

cgolubi1:
Okay, i think i've kicked off a replay of 3000 games on an AWS instance - that will run through the day, and assuming it doesn't find anything, i'll be pretty happy about gameplay regressions.

cgolubi1:
There were no surprising errors in the 3000 games.

cgolubi1:
I didn't see a whole lot in BMInterface that i was interested in testing --- i created and joined an open game and poked around in the buttons tree a bit.

I played my twin die games, and saw nothing of interest.

I brought up a site with a prod database, and saw nothing of interest.

Conclusion: further testing is not actually adding value, this is probably fine.  Merging.

Issue #1763, opened 2015-11-03
horizontal line between Rejected/Completed games and New/Active games


cgolubi1:
In the Overview page, i'd like to see a visual separation between the tables which are just there for your information and have no action for you to take (Rejected and Completed) and the tables which are potentially waiting for your move (New and Active).

* One option is to reintroduce the horizontal line between those tables, which initially had that function.
* Another option might be to use coloring -- right now all the table header backgrounds are grey.  Maybe the table header backgrounds for "New" and "Active" could be a more active-looking color instead.

cgolubi1:
Screencap of the page with four undifferentiated tables:

![four_tables](https://cloud.githubusercontent.com/assets/2898207/10908061/daa8a864-81fc-11e5-96db-a1c306b18886.png)

irilyth:
I think both the HR and the color would be nice, the better to appeal to people with different visual styles (have a hard time seeing thin lines, have a hard time distinguishing colors).

cgolubi1:
Since we had the HR before and it is semantically meaningful to the player, i think losing it is a regression, and fixing that regression is a must-have for December and should be very easy.

If it's easy enough to do something with color as well or instead, great.

blackshadowshade:
For the record, I think that the horizontal line will be difficult, because we'll need to somehow work out whether we have any of the categories either above or below the line before deciding whether to generate the line. Colour is trivial in comparison.

cgolubi1:
> For the record, I think that the horizontal line will be difficult, because we'll need to somehow work out whether we have any of the categories either above or below the line before deciding whether to generate the line

You could, or you could move the HR into `Overview.pageAddGameTables()`, and just put it exactly where you want it.

blackshadowshade:
... meaning that it's always there, even if there's nothing above it?

Sure, I could do that, but I thought that that would look odd.

cgolubi1:
Hmm, yeah, good point --- i think the overall point, that the HR doesn't belong in `Overview.pageAddGameTable()`, is correct.  But it is the case that the Overview page has access to all of the relevant objects.  So it shouldn't be too hard to make a little function that just checks the lengths of the various arrays, and makes the line if A + B is nonzero and C + D is as well.

dwvanstone:
> In the Overview page, i'd like to see a visual separation between the tables which are just there for your information and have no action for you to take (Rejected and Completed) and the tables which are potentially waiting for your move (New and Active).

[...]

PR #1811, opened 2015-11-14
Refactored Api to ensure consistent handling of data of different game types

blackshadowshade:
Starts to address part of #1763.

This is a pure refactor, and shouldn't change any behaviour.

The aim of this pull request is to unify the duplicated codebase for different game types. It will simplify the process of changing the format of the New Games data from unsplit to split, which will then be achieved by simply changing parserargs.isSplit to true.

All further pull requests addressing #1763 will then be able to focus almost exclusively on Overview.js.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/953/ (if it passes)

### DIFF
--- a/src/ui/js/Api.js
+++ b/src/ui/js/Api.js
@@ -361,7 +361,7 @@ var Api = (function () {
   my.packageGameData = function(data, _, parserargs) {
     // the output format is one of the following:
     // - split into games awaiting the player and games awaiting the opponent
-    // - made entirely of completed games
+    // - made entirely of games not awaiting action
     if (parserargs.isSplit) {
       parserargs.target.games = {
         'awaitingPlayer': [],
--- a/src/ui/js/Api.js
+++ b/src/ui/js/Api.js
@@ -58,10 +58,18 @@ var Api = (function () {
   ////////////////////////////////////////////////////////////////////////
   // Generic routine for API POST used to parse data
 
-  my.apiParsePost = function(args, apikey, parser, callback, failcallback) {
-    my[apikey] = {
-      'load_status': 'failed',
-    };
+  my.apiParsePost = function(
+    args,
+    apikey,
+    parser,
+    callback,
+    failcallback,
+    parserargs
+  ) {
+    if (typeof my[apikey] === 'undefined') {
+      my[apikey] = {};
+    }
+    my[apikey].load_status = 'failed';
     args.automatedApiCall = my.automatedApiCall;
     $.post(
       Env.api_location,
@@ -75,7 +83,7 @@ var Api = (function () {
           };
           return failcallback();
         } else if (rs.status == 'ok') {
-          if (parser(rs.data, apikey)) {
+          if (parser(rs.data, apikey, parserargs)) {
             my[apikey].load_status = 'ok';
             return callback();
           } else {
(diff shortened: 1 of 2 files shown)


blackshadowshade:
Actually, I do have two questions here.

1. Should my.packageGameData() actually be part of the activity namespace?
2. Why is the presence of the corresponding unit test for my.packageGameData() not checked?

cgolubi1:
Okay, so i hadn't looked at this before, and my warning is that it's going to take me quite some time to understand how the refactor works and what was changed, before i can figure out how to verify that it hasn't changed anything important.

blackshadowshade:
So, just as a signpost for you, the basic refactor is to consolidate the four functions parseNewGamesData(), parseActiveGamesData(), parseCompletedGamesData(), and parseRejectedGamesData() into a common function packageGameData(). The rest is just to make this consolidation work.

src/ui/js/Api.js
```diff
@@ -279,150 +287,92 @@ var Api = (function () {
...
 
-  my.parseRejectedGamesData = function(data) {
-    my.rejected_games.games = [];
-    my.rejected_games.nGames = data.gameIdArray.length;
-    var i = 0;
-    while (i < my.rejected_games.nGames) {
+  ////////////////////////////////////////////////////////////////////////
+  // Generic parser and repackager of the data returned from the database
+
+  my.packageGameData = function(data, _, parserargs) {
```
cgolubi1: Hmm, what on earth is this _ arg?

src/ui/js/Api.js
```diff
@@ -279,150 +287,92 @@ var Api = (function () {
...
+    // the output format is one of the following:
+    // - split into games awaiting the player and games awaiting the opponent
+    // - made entirely of completed games
```
cgolubi1: Nit: i would say "- made entirely of games not awaiting action" - they could be completed or rejected (or broken, or some other "not awaiting" category we make up later)

cgolubi1:
Okay.  So, having finally read through it, yes, i like the idea of this refactor.
* I need to understand what the deal with the underscore argument is
* I want to just poke around all the functionality on the site with firebug enabled, and make sure nothing throws errors in the presence of this new optional parserargs argument (where "nothing" includes both the functions you modified and the functions you didn't modify).

cgolubi1:
Yeah, seems to work.  So what's up with the underscore?

blackshadowshade:
As far as I have read, the underscore is one convention for documenting that a required argument in JavaScript is unused. We could replace this with any alternative name.

src/ui/js/Api.js
blackshadowshade: Good point. I'll fix this when I can.

blackshadowshade:
Also, just reminding you that I had two questions, two weeks ago, and it would help my understanding to hear what your answers are, even if I don't change this particular pull request.

blackshadowshade:
Comment updated

cgolubi1:
Sorry about forgetting your questions:

1. Should my.packageGameData() actually be part of the activity namespace?

No.  The activity variables are primarily for caching information that the player provides via UI, so that if the UI has to give the player an error and reset a screen, it can retain those previous values.  I think Game.activity also does store some information about the game which is parsed from API variables, but iirc only specific pieces of interesting information wind up in *.activity --- a corpus of data obtained from the API should primarily live elsewhere.

2. Why is the presence of the corresponding unit test for my.packageGameData() not checked?

Because the unit test presence checker doesn't understand the "define all public variables and methods under my" business, and just finds functions by grepping.  So presence of unit tests under Api.js is not actually checked at all.  This is a bug.

cgolubi1:
> As far as I have read, the underscore is one convention for documenting that a required argument in JavaScript is unused. We could replace this with any alternative name.

Okay.  IMO it's more important to document what the variable is, than that it's unused, so i'm not a fan of this convention.  But i don't feel strongly about it and would prefer to accept your pull and move forward.

PR #1815, opened 2015-11-19
Reworked overview page to use only CLOSED and ACTIVE game tables

### LINKED ISSUE #1763: horizontal line between Rejected/Completed games and New/Active games

### LINKED ISSUE #1768: Change "Max wins" column to "Score // W/L/T (Max)"
The New and Rejected tables should both have "Score // W/L/T (Max)" rather than just "Max wins"; among other things, it makes it easy to tell at a glance what that column is, since "0/0/0 (3)" is instantly recognizable in a way that just "3" isn't.

### LINKED ISSUE #1782: "Game #" column in new-games and rejected-games tables should look like other tables
The cells in the "Game #" column of the new-games and rejected-games tables should look the same as in other tables, e.g. "View Game N", with a link to the game page.

### LINKED ISSUE #1795: Fix wrapping in Overview tables
Now that the description field is being truncated, I don't think we need to put hard breaks into the table headers anymore. Also, #1777 pointed out that we really shouldn't be letting the "[Accept] / [Reject]" links wrap.

I propose something like https://www.dropbox.com/s/5v0ivse23vngqcj/Overview.js?dl=0 (untested) to address these issues.

blackshadowshade:
Fixes #1763. Fixes #1768. Fixes #1782. Fixes #1795.

Dependent on #1811.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/962/

### DIFF
--- a/src/ui/js/Overview.js
+++ b/src/ui/js/Overview.js
@@ -255,7 +255,7 @@ Overview.pageAddGameTable = function(
     gamesource.reverse();
   }
 
-  var tableBody = Overview.tableBody(tableClass, sectionHeader, gameType);
+  var tableBody = Overview.tableBody(tableClass, sectionHeader);
   Overview.addTableRows(tableBody, gamesource, gameType);
 };
 
@@ -265,12 +265,19 @@ Overview.addTypeToGameSource = function(gamesource, gameType) {
   }
 }
 
-Overview.tableBody = function(tableClass, sectionHeader, gameType) {
+Overview.tableBody = function(tableClass, sectionHeader) {
   var tableBody = Overview.page.find('table.' + tableClass + ' tbody');
   if (tableBody.length > 0) {
     var spacerRow = $('<tr>', { 'class': 'spacer' });
     tableBody.append(spacerRow);
-    spacerRow.append($('<td>', { 'html': '&nbsp;', 'colspan': '7', }));
+    spacerRow.append($('<td>', {
+      'html': '&nbsp;',
+      'colspan': tableBody.closest('table')
+                          .children('thead')
+                          .children('tr:first')
+                          .children()
+                          .length,
+    }));
   } else {
     var tableDiv = $('<div>');
     tableDiv.append($('<h2>', {'text': sectionHeader, }));
@@ -286,11 +293,8 @@ Overview.tableBody = function(tableClass, sectionHeader, gameType) {
     headerRow.append($('<th>', {'text': 'Opponent', }));
     headerRow.append($('<th>', {'html': 'Score<br/>W/L/T&nbsp;(Max)', }));
     headerRow.append($('<th>', {'text': 'Description', }));
-    if (gameType == 'closed') {
-      headerRow.append($('<th>', {'text': 'Action', 'colspan': '2', }));
-    } else {
-      headerRow.append($('<th>', {'text': 'Inactivity', 'colspan': '2', }));
-    }
+    headerRow.append($('<th>', {'text': 'Inactivity', }));
+    headerRow.append($('<th>', {'text': 'Action', }));
     tableHead.append(headerRow);
     table.append(tableHead);
 
@@ -312,7 +316,6 @@ Overview.addTableRows = function(tableBody, gamesource, gameType) {
 
   for (var i = 0; i < gamesource.length; i++) {
     gameInfo = gamesource[i];
-    console.log(gameInfo);
     playerColor = gameInfo.playerColor;
     opponentColor = gameInfo.opponentColor;
 
@@ -337,7 +340,6 @@ Overview.addTableRows = function(tableBody, gamesource, gameType) {
     } else {
       gameLinkTd = $('<td>');
 
-  console.log(gameInfo);
       if (gameInfo.gameType == 'rejected') {
         gameLinkTd.append($('<a>', {
           'href': 'game.html?game=' + gameInfo.gameId,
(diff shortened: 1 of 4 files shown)


cgolubi1:
Okay, so, i'm going to go so far as to say that this pull is the blocker for having a December release, because this is the big thing that everyone wanted to see changed in reject game.  So let's focus like lasers on it, and i will secondarily *try* to focus like a laser on non-shadowshade-people's pulls (sorry, shadowshade, but you're already used to the catastrophic neglect i heap on the buttonmen dev community), but let's see if we can possibly get this done.

cgolubi1:
So, basically, the testing is in three parts:
* Does this appear to fix each of #1763, #1768, #1782, and #1795?
* Do i like the way the UI change looks?
* Do i like the way the code change looks?

cgolubi1:
Oh, let's see, shadowshade, if you're actually around and can tap your branch with your foot or whatever it is you have to do to update the diff to get rid of #1811, that'd be great.  If not, i'm not going to let it block me.

src/ui/js/Overview.js
```diff
@@ -213,25 +211,37 @@ Overview.pageAddGameTable = function(
...
-  if (gameType == 'finished') {
-    gamesource = Api.completed_games.games;
-    tableClass = 'finishedGames';
-  } else if (gameType == 'rejected') {
-    gamesource = Api.rejected_games.games;
-    tableClass = 'rejectedGames';
-  } else {
-    gamesource = Api.active_games.games[gameType];
+
+  switch(gameType) {
```
cgolubi1: FTR, it looks like we're not consistent about horizontal whitespace after switch --- i'd prefer we standardize on having some, because i like horizontal whitespace for everything that's not a function (so more like `switch (gameType) {`.

Obviously, i won't block the pull over this, just stating a preference for the future.

src/ui/js/Overview.js
```diff
@@ -213,25 +211,37 @@ Overview.pageAddGameTable = function(
...
+  case 'closed':
+    // closed games comprise completed games and rejected games
+    var gamesourceCompleted = Api.completed_games.games;
+    Overview.addTypeToGameSource(gamesourceCompleted, 'completed');
```
cgolubi1: Hmm... do we want to do this in Overview.js, or should `Api.packageGameData()` take care of it instead?  Probably either way is reasonable, it just feels a little hackish/verbose to do it here, and might be more natural in `packageGameData`

test/src/ui/js/test_Overview.js
```diff
@@ -184,16 +184,534 @@ test("test_Overview.pageAddGameTables", function(assert) {
...
 });
 
-test("test_Overview.pageAddGameTableNew", function(assert) {
+test("test_Overview.addTypeToGameSource", function(assert) {
+  var gamesource = [[], [], []];
+  var gameType = 'testType';
+  Overview.addTypeToGameSource(gamesource, gameType);
+  assert.equal(gamesource[0].gameType, 'testType', 'First element is assigned type');
+  assert.equal(gamesource[1].gameType, 'testType', 'Second element is assigned type');
+  assert.equal(gamesource[2].gameType, 'testType', 'Third element is assigned type');
```
cgolubi1: FTR, you should be able to do this whole thing with an `assert.deepEqual()`

cgolubi1:
Now that you've made the table wider and collapsed into fewer tables, IMO it would be better *not* to prevent the "W/L/T (Max)" text from wrapping.  With wrapping prevented, i get a horizontal scrollbar.  With wrapping allowed, everything fits horizontally with 3 lines of header --- that's more header, but since it's only twice instead of four times, it's better than having a scrollbar.

cgolubi1:
I have to put this down for a bit now --- status is that i've looked at the code and it looks okay modulo the comments i made (which are all effectively nits), i've looked at the visuals somewhat, and they look okay modulo the comment above about wrapping.  I still want to:
* load the dev site DB and play around with that, in particular make sure (1) there are no regressions to the UI for stale games, and (2) the sort ordering is correct for many games of different types
* verify that the four issues seem resolved

cgolubi1:
Yeah, i'm willing to see those four issues closed with this pull, and overall the pull looks good to me.

So the only thing i really want to see is removing the span from W/L/T (Max) that prevents it from wrapping --- i think i'm the right audience for that, because i really am at the borderline where letting that wrap prevents the need for horizontal scrolling, and it's obvious to me that given those two choices with the new UI, letting "(Max)" wrap is clearly superior.

I'm not doing a push to dev right now, so i'm going to leave that and see if shadowshade would like to make that change in this pull (and/or comment on any of the other nits).  If he hasn't commented when i have time to do a dev push, i'll go ahead and merge this as is at that time.

src/ui/js/Overview.js
blackshadowshade: Okay, that's fine.

test/src/ui/js/test_Overview.js
blackshadowshade: Good to know. I'll leave this as it is for the moment, though.

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/975/ (if it passes)

cgolubi1:
Jenkins did not pass, because you have a unit test which codes in the entire content of that table, including the span.

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/977/ (if it passes)

cgolubi1:
Thanks.  Merging.

PR #1818, opened 2015-11-20
Added confirmation popups for rejecting and cancelling games

### LINKED ISSUE #1777: Prompt for confirmation before rejecting a game
Re-creating a rejected game is rather annoying now, and will probably at least somewhat annoying in the future, so it'd be nice if it prompted for confirmation, like surrendering.

blackshadowshade:
Fixes #1777.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/967/

### DIFF
--- a/src/ui/js/Env.js
+++ b/src/ui/js/Env.js
@@ -30,7 +30,10 @@ if ('unit_test' in Env) {
   Env.ui_root = '../ui/';
 
   // We also want to mock the window and history objects in unit tests
-  Env.window = { location: {} };
+  Env.window = {
+    location: {},
+    confirm: function() {return true;},
+  };
   Env.history = {
     pushState: function(state, title, url) {
       Env.history.state = state;
--- a/src/ui/js/Overview.js
+++ b/src/ui/js/Overview.js
@@ -630,10 +630,10 @@ Overview.formAcceptGame = function(e) {
     Overview.showLoggedInPage);
 };
 
-Overview.formCancelGame = function(e, args) {
+Overview.formCancelGame = function(e, testArgs) {
   e.preventDefault();
 
-  if ((undefined === args) || !('isTest' in args) || !args.isTest) {
+  if ((undefined === testArgs) || !('isTest' in testArgs) || !testArgs.isTest) {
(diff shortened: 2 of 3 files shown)


blackshadowshade:
In this pull request, I added conditionals in the live code for test parameters. If we want to avoid that, we can use Sinon.JS instead to mock window.confirm() instead.

http://stackoverflow.com/questions/6608278/avoid-capture-verify-a-javascript-alert-when-testing-a-method-that-displays

blackshadowshade:
Jota suggests http://stackoverflow.com/questions/2274749/javascript-how-to-mock-confirm-method

cgolubi1:
So... i think the idea Jota pointed to of using a wrapper for window is the right one.  There's a fair amount of mocking of functions in the QUnit tests.

Another specific thing you should look at is Env.js --- that has a bunch of functionality which sets up e.g. Env.window.location to be the actual window.location on the live site, and a mock object when testing is happening.

So it seems to me that if you're going to setup mocking on a second window object (window.confirm), it would be very straightforward to do so using the same kind of method that is already used for Env.window.location.  That would be my recommendation, and should be very easy --- then we're keeping all of the test-centric functionality in one place, in Env.js.

blackshadowshade:
Okay, I've mocked Env.window.confirm, and that tidies things up nicely.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/973/ (if it passes)

blackshadowshade:
If I understand correctly, this pull request has been invalidated by #1832. Is that right? If so, it can be closed.

cgolubi1:
Oh, because the new button is only on the Overview page.  Yeah, that seems reasonable, but if i were you i wouldn't delete the branch yet --- that way if our flighty dev testers decide they'd prefer to bring back that column, we'll still have the code.  :>)

cgolubi1:
That said,  i do like the mocked confirm box --- maybe we should get that in the codebase so we'll have it later when we inevitably need it for something?

blackshadowshade:
I'm going to close this one, and we'll address the issue differently, i.e., directly from the game page.

Issue #1731, opened 2015-10-08
Make nicer UI pages for not-yet-accepted and rejected games


blackshadowshade:
In #1727, we solve the issue of rejecting games (#205), but the UI stubs shown to the player are not pretty. Ideally, we should show relevant information about the game.

jimmosk:
Moving this request here from https://github.com/buttonmen-dev/buttonmen/issues/1734:

Could there be a "Go to your next pending game (at least N)" link in the body of the "This game has been rejected" screen, same as there is on other game screens when you've done all you can on that game? (There's a Next Game link in the header, but some of us have gotten so used to having it be in the body that we never look up there.)

cgolubi1:
WRT jimmosk's request: as long as when we lay the pages out, we don't set `includeFooter = false`, i believe you'll get that link for free.

blackshadowshade:
The page for an unstarted new game should show the links to accept/reject (#1736).

blackshadowshade:
I'm going to link in #1312 here so that it doesn't get lost.

cgolubi1:
For the record, i would call this a prod-blocker if it were easier, because when you create a new game, you are given a link which goes to the game page, and the current behavior of that link is ugly and unhelpful.  I predict we will get a number of user complaints about that.

I think we can live without this for the benefit of getting the November release done, but it's a must-have for December.

irilyth:
I agree, I think this should be a prod-blocker. I think it suffers particularly, from a PR point of view, because players won't understand why we had to punt it: The overview page has these nice links that do just the right thing, why can't the game page also just have those links?

dwvanstone:
I notice that the New Games table does not have a link to the game. (Neither does "Your open games" on the Open Games page.) Perhaps a quick fix would be to have Rejected Games not have a link to the game for now until there is a nicer page.

dwvanstone:
When as I user I create a game, I see this message:  "Game 1912 created successfully. Go to game page". I think this is the link you're saying doesn't work well. Maybe the "Go to game page" should be removed for now as well until there is a nicer page. (Perhaps "Waiting for opponent to accept" would now be more appropriate?)

AdmiralJota:
I agree. I'm not sure we should even be encouraging people to think of the game as actually existing as a thing before anyone has joined it. From our point of view as programmers, it exists as a record in the DB. But from a player's point of view, it's still just a potential game that hasn't actually started yet. 

[...]

PR #1822, opened 2015-11-22
Added button table for new games

### LINKED ISSUE #1774: game page displayed in CHOOSE_JOIN_GAME does not show button recipes
See image --- the issue is that the API doesn't send dice until the game starts.  So possible fixes would include:
* Have the API send some dice
* Change the look of the page and make it look like the finished game page instead of the swing*-selection page, so there's a slot for the recipes
* Other

choose_join_game_no_recipe

blackshadowshade:
Fixes #1774.

### DIFF
--- a/src/ui/gui.css
+++ b/src/ui/gui.css
@@ -159,7 +159,7 @@ td.button_opponent {
     padding-bottom: 20px;
 }
 
-td.button_postgame {
+td.button_prepostgame {
     padding-top: 10px;
     padding-bottom: 10px;
 }
--- a/src/ui/js/Api.js
+++ b/src/ui/js/Api.js
@@ -27,29 +27,6 @@ var Api = (function () {
     'December',
   ];
 
-  // Game states must match those reported by the API
-  my.GAME_STATE_START_GAME = 'START_GAME';
-  my.GAME_STATE_APPLY_HANDICAPS = 'APPLY_HANDICAPS';
-  my.GAME_STATE_CHOOSE_JOIN_GAME = 'CHOOSE_JOIN_GAME';
-  my.GAME_STATE_SPECIFY_RECIPES = 'SPECIFY_RECIPES';
-  my.GAME_STATE_LOAD_DICE_INTO_BUTTONS = 'LOAD_DICE_INTO_BUTTONS';
-  my.GAME_STATE_ADD_AVAILABLE_DICE_TO_GAME = 'ADD_AVAILABLE_DICE_TO_GAME';
-  my.GAME_STATE_CHOOSE_AUXILIARY_DICE = 'CHOOSE_AUXILIARY_DICE';
-  my.GAME_STATE_CHOOSE_RESERVE_DICE = 'CHOOSE_RESERVE_DICE';
-  my.GAME_STATE_SPECIFY_DICE = 'SPECIFY_DICE';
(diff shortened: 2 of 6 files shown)


blackshadowshade:
Okay, I've got a problem here with JSHint:

Linting ../../src/ui/js/Api.js...ERROR
[L573:C32] W117: 'Game' is not defined.
    if ((Api.game.gameState == Game.GAME_STATE_CHOOSE_JOIN_GAME) ||
[L574:C32] W117: 'Game' is not defined.
        (Api.game.gameState == Game.GAME_STATE_REJECTED)) {

How do I tell JSHint that Api.js should also load Game.js? Or am I doing something fundamentally wrong here?

cgolubi1:
Hmm, without looking at your code at all: Api.js is a set of base methods to talk to and parse the API that all other classes use.  It can't *depend* on Game.js --- that's backwards.  If Game.js is calling a function in Api.js and that function needs to use Game.js's variables, that function should pass those variables.

cgolubi1:
Another idea: maybe the problem is that we've defined the game states as part of the game (putting them in Game.js), whereas really they are part of the API (and should be moved to Api.js).  I'd potentially buy that if it looked reasonable.

blackshadowshade:
So, I tried creating a GameState.js and having that as a required module, but that caused race conditions for the loading of Api.js.

Eventually, I've fallen back to putting the game states into Api.js.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/972/ (if it passes)

cgolubi1:
> Eventually, I've fallen back to putting the game states into Api.js

This seems unobjectionable to me, at least in principle.

cgolubi1:
We discussed this on IRC, and agreed that the problem shadowshade ran into with game states, is that the `playerWLTText()` function doesn't actually belong in Api.js in the first place --- it's a game state processing function, and should be in Game.js.  Shadowshade agreed to do that piece of cleanup (thanks).

cgolubi1:
I'm going to defer looking at the code any more until that fix is done, but i can go ahead and bring this up in a test site and start looking at it any time.

blackshadowshade:
I think that I've done the cleanup as discussed.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/979/ (if it passes)

cgolubi1:
Okay, now i see what this looks like.  Let me show two screencaps, comparing (1) the existing "game over" screen vs. (2) this new "new game" screen:
game_over_screen

new_game_screen

cgolubi1:
So, two things:
1. At the end of the game, the opponent's mat is inverted, so that the button info is below the image rather than above it, so the left and right are symmetric.  We made that change at player request, and it's implemented in `Game.buttonImageDisplay()`.  I think we want the same visual in the new game state.
2. To my eye, there's very little whitespace between the top info and the mat in the new game UI - i think just adding a blank line would fix this.

cgolubi1:
Otherwise this looks good, thanks for all the work on it.

blackshadowshade:
Those last changes have been made.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/983/ (if it passes)

cgolubi1:
Oh, good catch on fixing the page for the third party.  Taking a look at this now.

cgolubi1:
LGTM.  Merging.

blackshadowshade:
Just adding here that this partly addresses #1731.

PR #1826, opened 2015-11-28
1511 python api client caching

### LINKED ISSUE #1511: Add caching to the Python API client
It'd be cool if the Python API client had the (optional) ability to cache information about games that are completed, and which are thus very unlikely to change.

irilyth:
Fixes #1511.

This now uses a cache directory rather than a cache file. I think I'm fine with the duplication of the call to self.load_game_data(game), because trying to only have that in one place seems to me like it would require you to dip twice into the logic of "are we caching or not", and right now it's nice that all the "are we caching or not" stuff is under one top-level conditional.

I'm still using JSON for the cache, despite my current preference for YAML; if no one else cares, or if anyone likes YAML better, I'd be happy to change it.

This PR also includes a little script to bring up an interactive Python shell with game data loaded, which I created while I was debugging this; if it's undesirable, I can certainly remove it from the PR, I don't use it myself any more but someone might find it useful.


irilyth:
http://jenkins.buttonweavers.com:8080/job/buttonmen-irilyth/12/ (if it passes)

cgolubi1:
Okay, test plan:
* look at the code and have opinions
* try this out myself and see how using a cache file works for me
* try the interactive mode script and make sure it's usable by someone who isn't irilyth

tools/api-client/python/lib/bmutils.py
```diff
@@ -98,10 +100,40 @@ def wrap_create_game(self, pbutton, obutton='', opponent=''):
...
+        # get the game data from the cache
+        with open(cachefile) as cache_fh:
+          data = json.load(cache_fh)
+      # otherwise (the cache didn't already have a file for this game)
+      else:
+        # load the game
+        retval = self.load_game_data(game)
+        # if that didn't work, raise an exception
+        if not retval.status == 'ok':
+          raise ValueError, "Failed to call loadGameData, got: " + retval
```
cgolubi1: While you're touching this code, want to fix the bug in which this blows up because retval is an object, not a string?  I'm pretty sure the fix is just `s/retval/retval.message/` - you can change that here and in the non-cache case below.

cgolubi1:
I think it's a little weird that `interactive-game.py` has a hyphen and all the existing example utilities have underscores.  If you don't care at all and wanted to change it to an underscore for consistency, that'd be slick.  If you do care, i'm happy to agree to disagree and leave it as is.

tools/api-client/python/lib/bmutils.py
```diff
@@ -98,10 +100,40 @@ def wrap_create_game(self, pbutton, obutton='', opponent=''):
...
+        if data['gameState'] == "END_GAME":
+          # put the data for this game into the cache
+          with open(cachefile, 'w') as cache_fh:
+            json.dump(data, cache_fh, indent=1, sort_keys=True)
+    # otherwise (we aren't using a cache directory), load the game
+    else:
+      retval = self.load_game_data(game)
+      if not retval.status == 'ok':
+        raise ValueError, "Failed to call loadGameData, got: " + retval
+      data = retval.data
```
cgolubi1: Here's a suggestion for how to structure the logic of `wrap_load_game_data()` differently, to get rid of the repetition of the code block which actually gets the game:
```python
  data = None
  if self.cachedir:
    # Do all of your cache setup stuff and try to load the game from the cache file if it exists
  if data is None:
    # Actually load the data
    if self.cachedir:
      # do the cache write stuff
```

What do you think?  The repeated block is pretty short, so it's not too bad as is, but this seems like a potentially easy way to get around it.

cgolubi1: Ah, i see, rereading your PR, it sounds like you already thought of this and rejected it.  Well, i think i like it better, but don't feel strongly.  Obviously a subroutine would be another alternative.

cgolubi1:
Another potentially fraught question: would it be nice if the cache dir right-justified game numbers using a plausible number of zeroes so the files sorted correctly in a directory listing?  I guess if people aren't directly using the files in the directory, it doesn't really matter, but i imagine that people might.  (For my personal cache of game info, i use 6-digit numbers, because that seemed to work well on the old site.  I definitely agree it's more suspect to specify something like this in code.)

tools/api-client/python/interactive-mode.py
```diff
@@ -0,0 +1,26 @@
...
+bmconnection = bmutils.BMClientParser(os.path.expanduser(bmrc), site)
+if not bmconnection.verify_login():
+  print "Could not login"
+    
+# At this point you can do whatever you want. Here's how to load a game, and print its info in nice JSON.
+
+gamenumber = 3038
+
+game = bmconnection.wrap_load_game_data(gamenumber)
+json.dump(game, sys.stdout, indent=1, sort_keys=True)
```
cgolubi1: I know it's just an example, i really hate it when printed things don't end with newlines.  Why not use `print json.dumps(game, indent=1, sort_keys=True)` instead?

cgolubi1:
Okay, that's enough nits.  See what you think of all of those, it seems to work and looks fine otherwise.

And i do think the sample code is potentially useful, so let's keep it.

irilyth:
I left the logical structure alone for now, but made the other changes, which all seem good.

http://jenkins.buttonweavers.com:8080/job/buttonmen-irilyth/13/

cgolubi1:
Great, i'll take a look in the morning.

tools/api-client/python/lib/bmutils.py
```diff
@@ -98,10 +100,40 @@ def wrap_create_game(self, pbutton, obutton='', opponent=''):
...
-    if not retval.status == 'ok':
-      raise ValueError, "Failed to call loadGameData, got: " + retval
-    data = retval.data
+    # if we're using a cache directory
+    if self.cachedir:
+      # if the cache directory doesn't exist, create it
+      if not os.path.isdir(self.cachedir):
+        os.makedirs(self.cachedir)
+      # set the path to the cache file for this game
+      cachefile = self.cachedir + "/{0:010d}.json".format(game)
```
cgolubi1: Hmm.  This used to work if you passed game as either an int or a string, now it blows up if you pass a string.  Is that okay?  Maybe it's okay.  It blows up my scripts, but i can modify them.  Let me check to see if it blows up anything which we have committed to the repo.

cgolubi1: Indeed, sadly:
```python
$ ./game_info -c ~/games/buttonmen/dev-2012/python/.bmrc -s dev 1000
  File "./game_info", line 48, in <module>
    gamedata = b.wrap_load_game_data(args[0])
  File "lib/bmutils.py", line 109, in wrap_load_game_data
    cachefile = self.cachedir + "/{0:010d}.json".format(game)
ValueError: Unknown format code 'd' for object of type 'str'
```

I don't have a strong preference between having `wrap_load_game_data()` cast its argument to an int as its first action, vs declaring that that argument needs to be an int and updating the `game_info` example to do the cast, but you have to do one of those.

cgolubi1:
Error handling on retval.message looks a lot better:
```python
>>> game = bmconnection.wrap_load_game_data(10000)
  File "<stdin>", line 1, in <module>
  File "./lib/bmutils.py", line 121, in wrap_load_game_data
    raise ValueError, "Failed to call loadGameData, got: " + retval.message
ValueError: Failed to call loadGameData, got: Game 10000 does not exist.
>>> 
```

cgolubi1:
and the json.dump looks better, so it's just that one type-casting regression.

irilyth:
I think just wrapping 'game' in a call to int() should do the trick, and then I don't have to worry about what type it was before, or changing it to a different type that might break things, or whatever. Seems to work for your example anyway:

    +$ ./game_info -c ~/.bmrc -s dev 1000
    irilyth: Ringmaster:       f(6) f(6) f(8) f(8) (12) 
    rcw:     Stumbling Clowns: (8) t(8) (10) t(10) (X)  

Lemme know if that seems all right.

cgolubi1:
> Lemme know if that seems all right.

Absolutely.

cgolubi1:
This looks fixed to me.  Merging it.

PR #1828, opened 2015-11-30
Run bzip2 on backups while storing them

cgolubi1:
* Partially addresses problem that the backup disks on prod and (to a lesser extent) dev keep running out of space
* No jenkins because there are no changes to code that jenkins is aware of
* Tested manually on virtualbox with the dev database loaded:
```bash
sandbox,[~],13:32(0)$ ls -lart /srv/backup/
total 8
drwxr-xr-x 3 root root 4096 Nov 30 13:15 ../
drwxr-x--- 2 root adm  4096 Nov 30 13:32 ./
sandbox,[~],13:32(0)$ time sudo /usr/local/bin/backup_buttonmen_database

real    0m19.952s
user    0m18.453s
sys     0m1.084s
sandbox,[~],13:33(0)$ ls -lart /srv/backup/
total 2068
drwxr-xr-x 3 root root    4096 Nov 30 13:15 ../
drwxr-x--- 2 root adm     4096 Nov 30 13:32 ./
-rw-r--r-- 1 root root 2106650 Nov 30 13:33 buttonmen.20151130.sql.bz2
sandbox,[~],13:33(0)$ bzcat /srv/backup/buttonmen.20151130.sql.bz2 | mysql -u root buttonmen
sandbox,[~],13:33(0)$ sudo /usr/local/bin/backup_buttonmen_database 
sandbox,[~],13:37(0)$ ls -lart /srv/backup/
total 2068
drwxr-xr-x 3 root root    4096 Nov 30 13:15 ../
drwxr-x--- 2 root adm     4096 Nov 30 13:32 ./
-rw-r--r-- 1 root root 2106789 Nov 30 13:37 buttonmen.20151130.sql.bz2
sandbox,[~],13:37(0)$ 
```

That shows:
* the new backup scheme works
* the new backup scheme can be run a second time in a day and will overwrite the first backup (the current behavior)
* the resulting database can be loaded (then i went to the UI and verified that it looked fine)
* the approximate wall-clock time of running the new backups on a dev-sized database

### DIFF
--- a/deploy/vagrant/modules/buttonmen/templates/backup_database.erb
+++ b/deploy/vagrant/modules/buttonmen/templates/backup_database.erb
@@ -4,7 +4,7 @@ BACKUPDIR=/srv/backup
 DBNAME=buttonmen
 TODAY=`date +"%Y%m%d"`
 
-BACKUPFILE="${BACKUPDIR}/${DBNAME}.${TODAY}.sql.bz2"
+BACKUPFILE="${BACKUPDIR}/${DBNAME}.${TODAY}.sql"
 
 # Fail if BACKUPDIR isn't there
 if [ ! -d "$BACKUPDIR" ]; then
@@ -12,9 +12,18 @@ if [ ! -d "$BACKUPDIR" ]; then
   exit 1
 fi
 
-mysqldump -u root $DBNAME | bzip2 > $BACKUPFILE
+mysqldump -u root $DBNAME > $BACKUPFILE
 if [ ! -s "$BACKUPFILE" ]; then
   echo "after backup, backup file is missing or zero size"
+  exit 1
+fi
+
+# Now compress the backup to save disk space
+# This has to be done as a separate step to save site performance
+# while backup/compression are happening
+bzip2 -f $BACKUPFILE
+if [ "$?" != "0" ]; then
+  echo "Bzip2 of successful backup failed"
   exit 1
 fi
 
--- a/deploy/vagrant/modules/buttonmen/templates/backup_database.erb
+++ b/deploy/vagrant/modules/buttonmen/templates/backup_database.erb
@@ -4,7 +4,7 @@ BACKUPDIR=/srv/backup
 DBNAME=buttonmen
 TODAY=`date +"%Y%m%d"`
 
-BACKUPFILE="${BACKUPDIR}/${DBNAME}.${TODAY}.sql"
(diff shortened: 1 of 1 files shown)


cgolubi1:
Don't merge this - it won't work on prod.  1s

cgolubi1:
We can't put the bzip2 in a pipe with the mysqldump, because that keeps the entire database in memory while compressing it.  With the prod database on virtualbox, that makes UI access hang while the backup is running.  So we have to store an intermediate disk file.

cgolubi1:
Okay, updated - i tested the new version by loading the prod database onto virtualbox and running a backup while repeatedly reloading the overview page.  Site access may have been slightly slow, but it did succeed, and so did the backup:
```bash
sandbox,[~],14:03(0)$ time sudo /usr/local/bin/backup_buttonmen_database

real    1m55.262s
user    1m43.870s
sys     0m4.748s
sandbox,[~],14:05(0)$ ls -l /srv/backup/
total 9928
-rw-r--r-- 1 root root 10164503 Nov 30 14:03 buttonmen.20151130.sql.bz2
sandbox,[~],14:08(0)$ 
```

cgolubi1:
Incidentally: my thinking is i'd rather have this on prod this month than worry about the merge freeze, because me forgetting to clean the backup disk is a bigger threat to our backups than pushing bad code.  I'll definitely make sure it runs normally (out of cron) on dev at least once before the prod push.

But that's why we have a review system, so you decide what you think.

blackshadowshade:
How long does it take to do the compress step? Are we talking seconds or minutes here?

blackshadowshade:
I guess the second question is: when is the cron job scheduled to run?

blackshadowshade:
And my suggestion is that it might be an idea to 'nice' the compress step.

cgolubi1:
The compress step takes several minutes on prod.  It runs around midnight UTC.  I tried nice, and afaict it doesn't make a bit of difference (though since it doesn't make a bit of difference, i'm happy to put it back in if you'd prefer) --- the problem isn't CPU, the problem is memory.

I'm not too worried about this on prod as-is --- i don't think the site was actually slow, it's just that i didn't test.  backup-then-compress in a pipe dramatically hung the site, and that effect was not present in the current version.

blackshadowshade:
Okay, I'll approve this. If we see any obvious performance issues on dev, we can look more closely then.

PR #1833, opened 2015-12-03
Mocked Env.window.confirm()

blackshadowshade:
This extracts the mocking code from #1818, since we probably won't be accepting that pull request in its entirety.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/984/


cgolubi1:
In case i'm totally wrong about how everything works, i'm going to bring this up in virtualbox and confirm a surrender action to make sure we didn't break it, but i don't think i'm totally wrong about how everything works.

cgolubi1:
Yeah, it's fine.  Merging.

PR #1835, opened 2015-12-04
Added error message if a game decision becomes invalid

### LINKED ISSUE #1757: Canceling an accepted game gives incorrect "Loaded data" error.
Player 1 creates a game, then goes to the Overview page and sees the game listed in "New Games".

Player 2 accepts the game.

Player 1 hits "Cancel". 

The error message for player 1 is "Loaded data for game ##."

blackshadowshade:
Fixes #1757.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/987/

### DIFF
--- a/src/engine/BMInterface.php
+++ b/src/engine/BMInterface.php
@@ -434,7 +434,7 @@ class BMInterface {
         if (BMGameState::CHOOSE_JOIN_GAME != $game->gameState) {
             if (('reject' == $decision) &&
                 ($playerId == $game->playerIdArray[0])) {
-                $decision = 'cancel';
+                $decision = 'withdraw';
             }
             $this->set_message(
                 'Your decision to ' .
--- a/src/engine/BMInterface.php
+++ b/src/engine/BMInterface.php
@@ -432,6 +432,15 @@ class BMInterface {
         $game = $this->load_game($gameId);
 
         if (BMGameState::CHOOSE_JOIN_GAME != $game->gameState) {
+            if (('reject' == $decision) &&
+                ($playerId == $game->playerIdArray[0])) {
+                $decision = 'cancel';
+            }
+            $this->set_message(
+                'Your decision to ' .
+                $decision .
+                ' the game failed because the game has been updated in the meanwhile'
+            );
             return;
         }
 
--- a/src/engine/BMInterface.php
+++ b/src/engine/BMInterface.php
@@ -439,7 +439,8 @@ class BMInterface {
             $this->set_message(
                 'Your decision to ' .
                 $decision .
-                ' the game failed because the game has been updated in the meanwhile'
+                ' the game failed because the game has been updated ' .
+                'since you loaded the page'
             );
             return;
         }
--- a/test/src/api/responderTest.php
+++ b/test/src/api/responderTest.php
@@ -1433,6 +1433,16 @@ class responderTest extends PHPUnit_Framework_TestCase {
 
         $this->assertEquals($retdata, $dummydata,
             "Real and dummy game joining return values should both be true");
+
+       // If the creating player tries to reject the game after the target player
+        // has already accepted it, the API behavior should be reasonable
+        $_SESSION = $this->mock_test_user_login('responder004');
+        $args = array(
+            'type' => 'reactToNewGame',
+            'gameId' => $gameId,
+            'action' => 'reject',
+        );
+        $retval = $this->verify_api_failure($args, 'Your decision to cancel the game failed because the game has been updated since you loaded the page');
     }
 
     public function test_request_reactToNewGameReject() {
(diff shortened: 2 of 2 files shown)


src/engine/BMInterface.php
```diff
@@ -432,6 +432,15 @@ public function save_join_game_decision($playerId, $gameId, $decision) {
...
 
         if (BMGameState::CHOOSE_JOIN_GAME != $game->gameState) {
+            if (('reject' == $decision) &&
+                ($playerId == $game->playerIdArray[0])) {
+                $decision = 'cancel';
+            }
+            $this->set_message(
+                'Your decision to ' .
+                $decision .
+                ' the game failed because the game has been updated in the meanwhile'
```
cgolubi1: Turning around pulls fast means i get to be super picky without feeling guilty about it.  Yay!

I don't think "in the meanwhile" is grammatical, and it doesn't really tell the player what's going on - which "meanwhile"?  How about "...the game has been updated since you loaded the page"?  Maybe that's too inside-baseball.

cgolubi1:
Anyway, let me bring this up in a test site and see if it works --- i also *believe* that, contrary to my previous comment, we could actually add a test for this to one of the existing responderTests in three lines or so --- i'll see if i can propose such a test.

cgolubi1:
Okay, cool, this passes on the pre-update code, demonstrating the problem:
```git
$ git diff
index 61ca4e5..5213c3b 100644
@@ -1400,6 +1400,9 @@ class responderTest extends PHPUnit_Framework_TestCase {
         $this->assertEquals('ok', $dummyval['status'], 'Dummy load of new games should succeed');
     }
 
+    /**
+     * @group fulltest_deps
+     */
     public function test_request_reactToNewGameAccept() {
         $this->verify_login_required('reactToNewGame');
 
@@ -1433,6 +1436,16 @@ class responderTest extends PHPUnit_Framework_TestCase {
 
         $this->assertEquals($retdata, $dummydata,
             "Real and dummy game joining return values should both be true");
+
+       // If the creating player tries to reject the game after the target player
+        // has already accepted it, the API behavior should be reasonable
+        $_SESSION = $this->mock_test_user_login('responder004');
+        $args = array(
+            'type' => 'reactToNewGame',
+            'gameId' => $gameId,
+            'action' => 'reject',
+        );
+        $retval = $this->verify_api_failure($args, 'Loaded data for game ' . $gameId . '.');
     }
 
     public function test_request_reactToNewGameReject() {
```

cgolubi1:
So here's my actual proposed test diff (modulo any changes in wording in BMInterface of course:
```diff
$ git diff ../../test/src/api/responderTest.php 
diff --git a/test/src/api/responderTest.php b/test/src/api/responderTest.php
index 61ca4e5..08044de 100644
--- a/test/src/api/responderTest.php
+++ b/test/src/api/responderTest.php
@@ -1433,6 +1433,16 @@ class responderTest extends PHPUnit_Framework_TestCase {
 
         $this->assertEquals($retdata, $dummydata,
             "Real and dummy game joining return values should both be true");
+
+       // If the creating player tries to reject the game after the target player
+        // has already accepted it, the API behavior should be reasonable
+        $_SESSION = $this->mock_test_user_login('responder004');
+        $args = array(
+            'type' => 'reactToNewGame',
+            'gameId' => $gameId,
+            'action' => 'reject',
+        );
+        $retval = $this->verify_api_failure($args, 'Your decision to cancel the game failed because the game has been updated in the meanwhile');
     }
 
     public function test_request_reactToNewGameReject() {
```

cgolubi1:
Anyway, the behavior is right, so my requests are just:
* think about improving the grammar of the message (if you don't want to, i don't feel strongly)
* integrate a responderTest diff like the one proposed above to nail down the behavior

src/engine/BMInterface.php
AdmiralJota: "Meantime" might be the word to use here.

cgolubi1: Sure, "meantime" is more grammatical than "meanwhile", but i still think it makes players ask "in the meantime since what?", and we might as well tell them what actually happened instead.  But, again, i don't feel at all strongly - i probably would have gone ahead and merged this except that i'd like shadowshade to add the responderTest

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/991/ (if it passes)

blackshadowshade:
Okay, wait up on this. I'll change 'cancel' to 'withdraw', in line with the other changes I'm making.

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/992/ (if it passes)

cgolubi1:
Looks good.  We need to merge #1837 to finish standardizing on "Withdraw" - this will look slightly wrong until we do - but we're circling around that, so this is fine.

PR #1837, opened 2015-12-04
Added confirmation messages for Reject Game and Cancel Game links

### LINKED ISSUE #1767: When you cancel a game, include a link to the game, in case you mis-clicked
When you create a game, and then cancel it, the "Successfully rejected game" text at the top" should instead say "Successfully canceled Game N", with a link, so that if you clicked cancel by mistake, you could click the link, get the game page, and click something like "create a new game against with the same parameters", much like the "create a new game" follow-up buttons after a completed game. So this would depend on the rejected-game page having some more stuff on it, but once it did, insta-dismissing but linking to that seems good.

### LINKED ISSUE #1777: Prompt for confirmation before rejecting a game

blackshadowshade:
Fixes #1777. Fixes #1767.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/989/

### DIFF
--- a/src/ui/js/Game.js
+++ b/src/ui/js/Game.js
@@ -368,7 +368,7 @@ Game.actionShowRejectedGame = function() {
   Game.form = null;
 
   Game.page = $('<div>');
-  Game.pageAddGameHeader('This game has been rejected or cancelled');
+  Game.pageAddGameHeader('This game has been cancelled');
 
   var dieEndgameTable = $('<table>');
   var dieEndgameTr = $('<tr>');
@@ -1556,7 +1556,7 @@ Game.formCancelGame = function(e) {
   e.preventDefault();
 
   var doGameCancel = Env.window.confirm(
-    'Are you SURE you want to cancel this game?'
+    'Are you SURE you want to withdraw this game?'
   );
   if (!doGameCancel) {
     return;
@@ -1572,7 +1572,7 @@ Game.formCancelGame = function(e) {
     'gameId': $(this).attr('data-gameId'),
   };
   var messages = {
-    'ok': { 'type': 'fixed', 'text': 'Successfully cancelled game', },
+    'ok': { 'type': 'fixed', 'text': 'Successfully withdrew game', },
     'notok': { 'type': 'server' },
   };
   Api.apiFormPost(
@@ -1790,7 +1790,7 @@ Game.pageAddGameHeader = function(action_desc) {
       acceptRejectDiv.append(rejectLink);
     } else {
       var cancelLink = $('<a>', {
-        'text': '[Cancel Game]',
+        'text': '[Withdraw Game]',
         'href': '#',
         'data-gameId': Api.game.gameId,
       });
--- a/src/ui/js/Game.js
+++ b/src/ui/js/Game.js
@@ -1554,6 +1554,14 @@ Game.formPlayTurnInactive = function() {
 // "Form" for cancelling a game
 Game.formCancelGame = function(e) {
   e.preventDefault();
+
+  var doGameCancel = Env.window.confirm(
+    'Are you SURE you want to cancel this game?'
+  );
+  if (!doGameCancel) {
+    return;
+  }
+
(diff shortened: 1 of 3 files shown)


cgolubi1:
Sure, this isn't complicated, and thanks for garbage-collecting the Overview.js functions.

I'll bring it up in virtualbox and make sure it works, then merge it.

cgolubi1:
So, i don't wanna be picky, *but* i think this is probably not okay:
cancel_cancel_cancel

The user has just clicked a button which says "Cancel", and we're prompting them "Are you sure you want to cancel the game?" and we're showing them two buttons, "OK" which cancels the game, and "Cancel" which cancels the action to cancel the game - we can't use the same word to mean two opposite things in the same dialogue.

cgolubi1:
Not to beat on my hobbyhorse from 10 minutes ago, but i wonder if it's too late to get rid of the word "Cancel".  You could easily imagine this link saying "Reject your own game" (it's not like we don't have the horizontal space, now that this action no longer needs to be taken from the overview page), or something more terse that someone else comes up with, but i'll throw "Reject your own game" out there as a suggestion, or even "Reject the game you proposed"

Basically, my thinking is that getting rid of the word "Cancel" as a verb that someone can do to a new game will make us happier in the long-term, and in the short term, it would solve this particular problem.

cgolubi1:
Anyway, the functionality works:
* Cancelling a game succeeds
* Choosing not to cancel a game succeeds
* Rejecting a game succeeds
* Clicking Reject, clicking Cancel in the dialogue, and then accepting the game succeeds

So that's all good, let's just talk about the word Cancel.

irilyth:
Heh: I was just thinking that we should consolidate down to one term, and that "cancel" was much better than "reject ". I see your point about the Ok vs Cancel thing, though. Hmm.

-Josh (from my phone)

AdmiralJota:
Retract, Rescind, Revoke, Call Off, Withdraw...

irilyth:
One thing I like about "cancel" is that it feels like it applies equally well to the creator and the opponent. Things like "retract" and "revoke" doesn't really sound to me like what I'm doing if you propose a game and I don't accept it. (And vice versa "reject".)

blackshadowshade:
Okay, we're going with Reject/Withdraw for the active verbs, and Cancelled for past participle.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/992/ (if it passes)

cgolubi1:
Okay, great, let me:
* bring this up in virtualbox
* see if i can find the word "Cancel" in the UI to kick around any more

cgolubi1:
The front UI page still says "REJECTED".  Is that what we want?
front_ui_says_rejected

cgolubi1:
Looks fine otherwise.  My vote is that changing to CANCELLED on the front page would be more consistent with "two verbs and a noun".  It's also more correct in the case in which the proposing player withdraws their own game and the target player sees that game in the closed column, since that game was technically withdrawn not rejected.

cgolubi1:
Adding "dev-blocker" to remind myself that we should fully standardize on Withdraw/Reject/Cancelled before pushing to dev, because testers will be confused by an intermediate state.

irilyth:
> My vote is that changing to CANCELLED on the front page would be more consistent with "two verbs and a noun".

Sounds good to me, except that it turns out "cancelled" is actually an adjective, not a noun.

cgolubi1:
> Sounds good to me, except that it turns out "cancelled" is actually an adjective, not a noun.

I reject your narrow concepts of grammar.

blackshadowshade:
Okay, try this.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/995/ (if it passes)

cgolubi1:
Thanks.  Let's try this out.

PR #1838, opened 2015-12-04
Added Rejected / Cancelled to History page

### LINKED ISSUE #1770: Rejected and canceled games should show up in history
Rejected and canceled games (and, when we have this, forfeited games) should show up in the history page; possibly with a way to include them or not, if we think (or hear) that people don't like that, but it should be *possible* to find them there.

blackshadowshade:
Fixes #1770.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/990/

### DIFF
--- a/src/engine/BMInterface.php
+++ b/src/engine/BMInterface.php
@@ -1834,7 +1834,9 @@ class BMInterface {
             $whereParameters[':status_%%%'] = $searchFilters['status'];
         } else {
             // We'll only display games that have actually started
-            $where .= 'AND (s.name = "COMPLETE" OR s.name = "ACTIVE") ';
+            $where .= 'AND (s.name = "COMPLETE" ' .
+                      'OR s.name = "ACTIVE" ' .
+                      'OR s.name = "REJECTED") ';
         }
     }
 
--- a/src/api/ApiSpec.php
+++ b/src/api/ApiSpec.php
@@ -258,7 +258,7 @@ class ApiSpec {
                 ),
                 'status' => array(
                     'arg_type' => 'exactString',
-                    'values' => array('ACTIVE', 'COMPLETE'),
+                    'values' => array('ACTIVE', 'COMPLETE', 'REJECTED'),
                 ),
             ),
         ),
--- a/src/ui/js/History.js
+++ b/src/ui/js/History.js
@@ -75,9 +75,10 @@ History.searchParameterInfo = {
     'source': {
       'COMPLETE': 'Completed',
       'ACTIVE': 'In Progress',
+      'REJECTED': 'Rejected / Cancelled',
     },
     'dataType': 'string',
-    'toolTip': 'Search or sort by whether the game is completed or in progress',
+    'toolTip': 'Search or sort by the game status',
   },
   'sortColumn': {
     'inputType': 'hidden',
@@ -662,6 +663,11 @@ History.buildResultsTableBody = function() {
     if (game.status == 'COMPLETE') {
       gameRow.append($('<td>', {
         'text': 'Completed',
+        'style': 'font-weight: bold;'
+      }));
+    } else if (game.status == 'REJECTED') {
+      gameRow.append($('<td>', {
+        'text': 'Rejected / Cancelled',
         'style': 'font-weight: bold;'
       }));
     } else {
--- a/src/ui/js/History.js
+++ b/src/ui/js/History.js
@@ -70,12 +70,12 @@ History.searchParameterInfo = {
       '(for completed games, this is the player who won)',
   },
   'status': {
-    'text': 'Completed?',
+    'text': 'Game status',
     'inputType': 'select',
     'source': {
       'COMPLETE': 'Completed',
       'ACTIVE': 'In Progress',
-      'REJECTED': 'Rejected / Cancelled',
+      'REJECTED': 'Cancelled',
     },
     'dataType': 'string',
     'toolTip': 'Search or sort by the game status',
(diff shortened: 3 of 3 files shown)


cgolubi1:
Okay.  At a glance, i'm wondering if this is going to make the whole history page wider, because Rejected/Cancelled is much longer than Active or Completed.  How strongly do we feel about differentiating?

cgolubi1:
Anyway, there's nothing for me to do here except bring this up in virtualbox and take a look.

cgolubi1:
Okay, so here's what it looks like when i do a history search on myself (note that it's scrolled to the right, because the History page is too wide for my browser, but that's not really new):
history_search

cgolubi1:
The thing that i think is definitely wrong is boldfacing Rejected/Cancelled - that column seems to have "Completed" boldfaced and "In Progress" in italics - my suggestion would be to have no font formatting on "Rejected/Cancelled".

I don't want to get people used to the idea of thinking of Rejected and Cancelled as separate *game* states --- they're not, they're actions which we've given separate names, but they lead to the same game state of a game which was thrown away before it was started.  IMO encouraging people to think of a rejected game as different from a cancelled game takes up more horizontal space for the benefit of making people's expectations slightly more wrong.  :>)  So i'd rather we not go there more than needed, and thus would favor simply calling this state "Rejected".

Otherwise, this appears to work.

src/engine/BMInterface.php
```diff
@@ -1834,7 +1834,9 @@ protected function apply_all_filters($searchFilters, &$where, &$whereParameters)
             $whereParameters[':status_%%%'] = $searchFilters['status'];
         } else {
             // We'll only display games that have actually started
```
AdmiralJota: Does this comment still strictly apply?

AdmiralJota:
I think that either "Rejected" or "Cancelled" would work by itself. It's unfortunate that the word "Aborted" has other undesirable connotations, since I think it'd be exactly right otherwise. Other possibilities might be "Unaccepted", "Annulled", "Retracted", "Called Off", "Void", "Forsaken"...

As for formatting, I agree that it shouldn't be bolded or italicized. But it might be reasonable to make the color some shade of grey rather than black.

Also, as @irilyth mentioned in #1770, I think we should still have a way to show all the games that have actually been started (i.e., all active and completed games). I can see two approaches for that, both with pros and cons:

* Keep the blank, default option to ACTIVE and COMPLETED, and add an extra option at the end of the dropdown for "All (including Rejected)". On the one hand, I think the current behavior of the default being all games <i>that have actually started at some point</i> is probably what most people would expect and be looking for as the default. But on the other hand, it means that it's now not as obvious what the default option means without making inferences from the rest of the list.

* Have the blank, default option be all games (including rejected ones), and add an item to the list after Active and Completed that's something like "Active / Completed" or "All Started Games" or "This Is The Option You Probably Mean to Pick Unless You Really Want To See Rejected Games Mixed In With Everything Else For Some Reason". This has the upside that it's clear what every option means (provided you realize that rejected games are a thing that can appear in the history list at all), but the downside that it seems to make the dropdown options kinda more awkward. And it includes Rejected games in the default case, whereas I strongly suspect that wanting to include rejected games is going to be a special rare use case rather than the standard one.

* I suppose that, technically, there's the third option of "change the dropdown into some wacky multi-select control and let users check off individual statuses", but A) that seems like a much bigger change than we should be considering right now, and B) there are additional implementation implications and UI decisions that would need to follow from that. Although it might be something we may want to consider at some point in the future.

So, while we're here, adding Rejected games to the history list... why are we excluding challenges that haven't been accepted or declined yet? It seems like they both have about equal standing in the "is it really a game?" question -- arguably more standing, since they still show up in the active games list and can still be promoted to real games.

cgolubi1:
If we're suggesting replacement words entirely, i'd add Discarded and Abandoned as suggestions.

cgolubi1:
> Also, as @irilyth mentioned in #1770, I think we should still have a way to show all the games that have actually been started (i.e., all active and completed games)

This seems like a fine idea, but IMO this pull request would still be an improvement without it.

cgolubi1:
> So, while we're here, adding Rejected games to the history list... why are we excluding challenges that haven't been accepted or declined yet?

This is a fair question - we could probably add "New" games to the pulldown pretty easily using the same logic we already have.

irilyth:
> I don't want to get people used to the idea of thinking of Rejected and Cancelled as separate game states --- they're not, they're actions which we've given separate names, but they lead to the same game state of a game which was thrown away before it was started.

One way in which they're different from a player's point of view is that you might be looking for games that I created but that other people rejected, and not care about games that I created and then canceled.

I suppose there isn't an obvious way in the history interface to differentiate between games I created and the other player rejected, and games another player created and I rejected, is there? (There is in the DB, though, right? In the former, I'm "player" and in the latter I'm "opponent"?)

I wonder if we want to start thinking about the current history page as the "simple game search" page, and think about an "advanced game search" page where you could search by criteria that you can't currently enter on the current page (like creator, description, was this a followup game, etc).

blackshadowshade:
Right, I've changed the word to 'Cancelled' and grayed it out.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/994/ (if it passes)

cgolubi1:
Okay, let me check this out and see if it addresses my concerns - there's still Jota's request to be able to see New games in the search, but IMO we can/should make that a separate issue.

cgolubi1:
I opened #1839 and #1840 for things which are more about New games on the History page than about Cancelled games on the history page.  Now i am going to merge this.

PR #1845, opened 2015-12-09
Fixed bug with displaying opponent info on the game mat

blackshadowshade:
This pull request fixes the problem of the opponent's info not being visible on the game mat.

I am aware that there should be an appropriate unit test locking our expected behaviour into place, but I'd like to put this fix in first.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/997/

### DIFF
--- a/src/ui/js/Game.js
+++ b/src/ui/js/Game.js
@@ -2527,7 +2527,7 @@ Game.buttonImageDisplay = function(player) {
     'text': Api.game[player].button.recipe,
   });
 
-  if (player == 'opponent' && isPreOrPostGame) {
+  if (player == 'opponent' && !isPreOrPostGame) {
     buttonTd.append(playerName);
     buttonTd.append(buttonInfo);
     buttonTd.append(buttonRecipe);


cgolubi1:
For future reference - i have no memory of what the issue is that this pull is addressing, and the description here doesn't have a lot of context.  I'm sure it's obvious and awesome (and probably even something that i personally asked for), but if there's no issue to link to explaining the problem, i think the burden of explaining what this pull is for is higher.

cgolubi1:
Oh, i see.  Yeah, alright, let's fix that.

cgolubi1:
For those playing the home game:
for_those_playing_the_home_game

cgolubi1:
I'm going to bring this up in vagrant and:
* visually make sure it works in a game which is in progress
* visually make sure it works in a game which is *not* in progress (each of completed, cancelled, new)
* see how easy it would be to make a UI test to lock in all of that behavior

cgolubi1:
Without this fix:
* An active game is broken
* A completed game is fine
* A new game is fine
* A cancelled game is fine
* The situation is identical when those game types are viewed by a nonparticipant

This pull fixes the bug in the active game case, and does not regress the other three cases.

Merging - i agree with shadowshade that we can unit test later.

PR #1848, opened 2015-12-10
Added forum markup for button, set, wiki, issue

### LINKED ISSUE #1177: Add [button] and [set] tags to markup
Markup should support tags like ```[button="Kublai"]``` and ```[set="Vampyres"]```, e.g. in forum posts.

### LINKED ISSUE #1830: Add [wiki] tag to markup
It'd be cool if `[wiki=thing]` turned into a link to `http://buttonweavers.wikia.com/wiki/thing`.

### LINKED ISSUE #1841: Add [issue] tag to markup
It'd be cool if `[issue=thing]` turned into a link to `https://github.com/buttonmen-dev/buttonmen/issues/thing` .

blackshadowshade:
Fixes #1177. Fixes #1830. Fixes #1841.

With many thanks to @AdmiralJota for the fix.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1000/ (if it passes)

### DIFF
--- a/src/ui/js/Env.js
+++ b/src/ui/js/Env.js
@@ -340,7 +340,7 @@ Env.applyBbCodeToHtml = function(htmlToParse) {
       'isLink': true,
       'openingHtml':
           '<a class="chatWikiLink" ' +
-          'href="http://buttonweavers.wikia.com/wiki/###">',
+          'href="http://buttonweavers.wikia.com/wiki/###">Wiki: ',
       'closingHtml': '</a>',
       'escapeParameter': true,
     },
--- a/src/ui/js/Env.js
+++ b/src/ui/js/Env.js
@@ -434,7 +434,7 @@ Env.applyBbCodeToHtml = function(htmlToParse) {
         if (replacements[tagName].escapeParameter) {
           // We need to HTML decode the parameter before we URI encode it, 
           // and the easiest way is to pretend we're going to render it
-          var tempDiv = $("<div>");
+          var tempDiv = $('<div>');
           tempDiv.html(tagParameter);
           var decodedTagParameter = tempDiv.text();
           htmlOpeningTag = htmlOpeningTag.replace(
--- a/src/ui/js/Env.js
+++ b/src/ui/js/Env.js
@@ -319,6 +319,41 @@ Env.applyBbCodeToHtml = function(htmlToParse) {
       'closingHtml': '</a>',
       'escapeParameter': true,
     },
+    'button': {
+      'isAtomic': true,
+      'isLink': true,
+      'openingHtml':
+          '<a class="chatButtonLink" href="buttons.html?button=###">',
+      'closingHtml': '</a>',
+      'escapeParameter': true,
+    },
+    'set': {
+      'isAtomic': true,
+      'isLink': true,
+      'openingHtml':
+          '<a class="chatButtonSetLink" href="buttons.html?set=###">',
+      'closingHtml': '</a>',
+      'escapeParameter': true,
+    },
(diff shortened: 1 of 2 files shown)


AdmiralJota:
That's a good idea about adding "+ 'Issue '" for the text in the github links. I wonder if it might be good to do the same with "Wiki: " for the wiki links, since that's also an off-site link.

blackshadowshade:
Sounds reasonable. I'll add that now.

cgolubi1:
This looks pretty straightforward --- i'll just bring it up in virtualbox and see how it looks

cgolubi1:
I can't figure out how to link to Luna & Artemis.

* ```[button=Luna %26 Artemis]``` gives me http://localhost:8080/ui/buttons.html?button=Luna%20%2526%20Artemis which it doesn't like
* ```[button=Luna & Artemis]``` gives me http://localhost:8080/ui/buttons.html?button=Luna%20%26amp%3B%20Artemis which it doesn't like

The actual URL seems to be http://localhost:8080/ui/buttons.html?button=Luna%20%26%20Artemis (which is why i initially guessed ```[button=Luna %26 Artemis]]]```

I don't think the %20 is a problem, because ```[set=2005 Rare Promo]``` Just Works --- it's the ampersand that's the problem.

Or am i just missing how to do it?

cgolubi1:
I didn't have any problems linking to any other buttons, and i think i've covered all the characters that appear in button names.

AdmiralJota:
Ah, I see what's happening here. The contents of the forum post are HTML-encoded (for the purposes of displaying them on the web page) before the markup is even applied, so even though the user typed "[button=Luna & Artemis]", the Env.applyBbCodeToHtml() function is receiving "[button=Luna &amp; Artemis]". This is necessary, as we definitely want things to be HTML-encoded before we display them, and we can't do it after applying the markup (or else it would HTML-encode the tags generated by the markup).

I think the solution is to simply decode the text before putting it in the URL. And I believe the best way to do that is to update that function as per https://gist.github.com/AdmiralJota/338e2e58a6a56608f571

blackshadowshade:
This should work now. I've also updated the example shown on the left side of the forum.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1004/ (if it passes)

AdmiralJota:
Does "Luna & Artemis" fit in the help box? I would have expected it to be too long and to potentially wrap.

blackshadowshade:
The help for buttons goes over two lines, with the markup on one line and the resulting link on the next line. Personally, I think that it looks okay, but we'll see what Chaos thinks.

cgolubi1:
Hmm, yeah, i don't like the wrapping help text because the second line is justified with the first line, so it's confusing.  I think the help text box is slightly busy and hard to read as is, and the more consistent the examples can be, the better.

I also don't think there's a reason to use Luna & Artemis as the example --- people will try the obvious thing unless the obvious thing doesn't work, and if they can't figure out the obvious thing, they may or may not actually look at the help text.  I don't think you're going to prevent any significant number of questions by using that example, and it would be better to just use a short button name as an example.

cgolubi1:
Anyway, all the button, set, and wiki examples i had previously tried still work with the new code.  So i'm willing to merge this with a shorter example.

blackshadowshade:
Right, done. http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1005/ (if it passes)

cgolubi1:
Great!  Avis is everyone's favorite reference button.  Merging.

PR #1821, opened 2015-11-21
Made page titles consistent

dwvanstone:
Fixes #1806. Fixes #1125. Fixes #1684.

http://jenkins.buttonweavers.com:8080/job/buttonmen-dwvanstone/21/

For #1806:

All page titles have the subpage (or the identifying part of the subpage) in the front of the title. For instance, "Preferences - Button Men Online", For specific buttons, profiles, forum threads, etc.,  the specific name will be in the page title. (e.g., a button page: "Spite - Button Men Online"; a profile page "boilerbrown - Button Men Online")

Note that sometimes the HTML page was changed to have the simple title of "Button Men Online". (Such as buttons.html) In these cases the JS is expected to change the title. I found it less visually jarring to have the title change from "Button Men Online" to "Spite - Button Men Online" than from "Buttons - Button Men Online" to "Spite - Button Men Online". 

The overview page title now has the number of active games as part of the page title. ("(2) Button Men Online") The number in parentheses is set to be the sum of open games waiting your move plus the number of games waiting to be accepted by you. It does not include games that have finished and are waiting to be dismissed or games that you're waiting for others to accept. I think this matches the games that the "Redirect to waiting games when in Monitor mode" option would take you to.

For #1125:

The favicon will change to favicon_monitor.ico if the user is in monitor mode on the overview page. (The favicon_monitor.ico represents "I'm in monitor mode".) If the user has pinned the page (where only the favicon shows), they will notice the favicon change only if they have the redirect option on in the monitor mode.

If the redirect option is on in the monitor mode, the page title will be visibly different -- it will either go to the game page ("Game # ...") or to a forum thread page ("UBFC 64 - ..."). If the redirect option is off, the user will observe a difference in the title of the overview page when a game is ready to be played. ("(2) Button Men Online")

For #1684:

The page title now has the forum thread name. This won't help the user who has pinned the page (since only the favicon will appear). But this is at least an improvement for everyone else!

PR #1849, opened 2015-12-10
Added number of active games to overview

### LINKED ISSUE #1119: Add count to list of active games
Could your list of active games also show how many total games you're in? I think the old site did this.

blackshadowshade:
@AdmiralJota's solution that fixes #1119.

When #1821 has been accepted, this pull request will probably need to be deconflicted.

I am currently unsure whether the count of closed games is really necessary for consistency, since I don't think the number would every be really meaningful.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1001/ (if it passes)

### DIFF
--- a/src/ui/js/Overview.js
+++ b/src/ui/js/Overview.js
@@ -134,6 +134,22 @@ Overview.showPage = function() {
 
   // Actually lay out the page
   Login.arrangePage(Overview.page);
+
+  Overview.updateSectionHeaderCounts('activeGames');
+  Overview.updateSectionHeaderCounts('closedGames');
+};
+
+Overview.updateSectionHeaderCounts = function(tableClass) {
+  var header = Overview.page.find(
+    'div.' + tableClass + 'Holder h2.sectionHeader'
+  );
+  var tableRows = Overview.page.find(
+    'div.' + tableClass +
+    'Holder table.gameList tbody tr:visible:not(.spacer)'
+  );
+  var count = tableRows.size();
+
+  header.text(header.attr('data-text') + ' (' + count + ')');
 };
 
 Overview.executeMonitor = function() {
@@ -278,8 +294,12 @@ Overview.addTableStructure = function(tableClass, sectionHeader, showDismiss) {
 
   if (tableBody.length === 0) {
     // create table
-    var tableDiv = $('<div>');
-    tableDiv.append($('<h2>', {'text': sectionHeader, }));
+    var tableDiv = $('<div>', { 'class': tableClass + 'Holder', });
+    tableDiv.append($('<h2>', {
+      'text': sectionHeader,
+      'class': 'sectionHeader',
+      'data-text': sectionHeader,
+    }));
     var table = $('<table>', { 'class': 'gameList ' + tableClass, });
     tableDiv.append(table);
     Overview.page.append(tableDiv);
@@ -497,6 +517,7 @@ Overview.toggleStaleGame = function() {
     'Hide stale games' :
     'Show stale games'
   );
+  Overview.updateSectionHeaderCounts('activeGames');
 };
 
 Overview.pageAddIntroText = function() {
--- a/test/src/ui/js/test_Overview.js
+++ b/test/src/ui/js/test_Overview.js
@@ -175,6 +175,45 @@ test("test_Overview.showPage", function(assert) {
   });
 });
 
+test("test_Overview.updateSectionHeaderCounts", function(assert) {
+  stop();
+  Overview.getOverview(function() {
+    Overview.page = $('<div>');
+    Overview.pageAddGameTables();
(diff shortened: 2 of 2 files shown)


blackshadowshade:
This is currently missing a QUnit test for the new function Overview.updateSectionHeaderCounts(). I'll get to that when I can.

blackshadowshade:
I don't know how to write test_Overview.updateSectionHeaderCounts() correctly. For some reason, the call to Overview.updateSectionHeaderCounts('activeGames') isn't doing anything in the unit test.

Any idea how to fix this test?

cgolubi1:
Okay, so: this has been kicking around for awhile.  It needs to be deconflicted.  It does not have any code in it that Shadowshade cares about.  IIRC, i had some reservations about it, though i would be hard-pressed to remember what they are.

We think this was something Jota implemented in response to an issue irilyth opened.  Either of y'all want to weigh in on your attachment to our sorting this out, vs closing this PR and revisiting the original issue in the fullness of time?  Any responses are fine, but we don't want to think about it too hard if no one actually cares.  (But if you do care, please say so.)

AdmiralJota:
Am I correct in my reading that it's only the test file that's conflicted, not the code for the feature itself? Seeing as I didn't even write the test, I would certainly have no objections to test_Overview.js being rolled back and some other less conflict-y test getting written for it instead. 

Or for that matter, having the whole pull replaced with a new pull based off of https://gist.github.com/AdmiralJota/8afe049bd2fd46dc23c4, if that code is still salvageable (and if the test is the only thing that's conflicted, I don't see why it wouldn't be).

blackshadowshade:
So, this has been rebased into one commit and deconflicted.

blackshadowshade:
Note that I previously commented that one of the unit tests is broken. I'll need help to fix this unit test.

AdmiralJota:
I'm not spotting which is broken from the jenkins report. Is the test anything that I can help with without having a BM dev environment set up?

blackshadowshade:
Jenkins passes, the issue is more a logical one.

You'll note that in the test, the header should actually contain the number of games, but it doesn't, because the call to Overview.updateSectionHeaderCounts('activeGames') doesn't actually do anything in this context.

AdmiralJota:
Are we certain that active games are coming back from the server when we call Overview.getOverview()? (I.e., gamesource.length isn't 0 on line 265 of Overview.js?)

blackshadowshade:
So, I've just checked, and the values of gamesource.length (line 265 of Overview.js) are 129, 418, and 408.

The value of count (line 144 of Overview.js) is 0.

Any idea how I can debug the JQuery at line 144 to work out what's missing?

cgolubi1:
I stood up a dev site for this, see https://github.com/buttonmen-dev/buttonmen/wiki/Dev-site-index for IP.  I'm trying to debug getting Jota's IP able to load /test-ui, but i don't know if i'll get that done in the time i have or not.

cgolubi1:
FYI, here's the error the apache log is getting when the test fails:
```
[Fri May 05 19:55:23 2017] [error] [client X.X.X.X] DummyApiResponder tried to read nonexistent file 50.json in response to an API query for loadActivePlayers, referer: http://54.146.5.143/test-ui/
```

You probably have a UI test looking for dummy data that no responder test defines.

irilyth:
FWIW, the feature looks great to me.

cgolubi1:
My understanding here is that the feature itself is potentially ready to go (irilyth has signed off, no one has vetoed), but someone needs to get the UI unit tests passing, so the action item goes to anyone who has time to/is able to do that.

blackshadowshade:
So, I've finally worked out where the problem is here.

The issue is that the rows are present but aren't visible, so the selector tr:visible is obviously returning no rows. Time to work out why they're not visible.

blackshadowshade:
So, I think this now works. Let's see what Jenkins says.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1366/ (if it passes)

cgolubi1:
Tearing down the old dev site for this pull and standing up a new one.

cgolubi1:
irilyth, when you get a chance can you repoint the `temporary` CNAME to 54.145.71.191?

irilyth:
> irilyth, when you get a chance can you repoint the `temporary` CNAME to 54.145.71.191?

Done!

irilyth:
http://temporary.dev.buttonweavers.com/ for anyone looking for a link.

And, just to confirm, this feature still looks good to me. :^)

cgolubi1:
Okay, looks like this is ready to go now.  Merging it, and tearing down the dev site.

cgolubi1:
Done: irilyth, you can reclaim `temporary` at your leisure.

PR #1854, opened 2015-12-13
Reformatted count of pending games in title.

### LINKED ISSUE #1850: Reformat count of pending games in page header
In #1821, we added a header for the Overview or Monitor mode formatted like "(23 games waiting) - Button Men Online".  This is awkward looking, especially in a tab bar, where "(23) Button Men Online" would probably be much better.  Improve the situation.

dwvanstone:
Fixes #1850.

http://jenkins.buttonweavers.com:8080/job/buttonmen-dwvanstone/28/

I went with using the title "(N) Button Men Online" since that matches GitHub. 

An alternative would be to do what Google Mail does: "Overview (N) - Button Men Online." But I prefer having the title be as short as possible for this case.

### DIFF
--- a/src/ui/js/Overview.js
+++ b/src/ui/js/Overview.js
@@ -93,10 +93,8 @@ Overview.showPage = function() {
   var nGamesAwaitingAction = Api.new_games.nGamesAwaitingAction +
     Api.active_games.nGamesAwaitingAction;
   var gameCountText='';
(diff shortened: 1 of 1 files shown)


cgolubi1:
This looks simple enough --- i'll just bring it up in vagrant and check it out in the zero-games case and the more-than-zero games case.

cgolubi1:
Looks good.  Merging.

PR #1855, opened 2015-12-14
Changed W/L/T to –/–/– for rejected games on the history page

### LINKED ISSUE #1840: When showing New and Cancelled games in the History UI, use -/-/- for the W/L/T
What it says on the tin.

blackshadowshade:
Addresses part of #1840.

I am aware that the logic to detect new games is not really the best way, but until we explicitly add the ability to search for New Games in the history page (and perhaps add an extra game status type), the current implementation seems to be the easiest way to achieve the goal.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1011/

### DIFF
--- a/src/ui/js/History.js
+++ b/src/ui/js/History.js
@@ -668,8 +668,7 @@ History.buildResultsTableBody = function() {
 
 History.scoreCol = function(game) {
   var score;
-  if ((game.status == 'REJECTED') ||
-      (game.gameStart == game.lastMove)) {
+  if (game.status == 'REJECTED') {
     score = '–/–/–';
   } else {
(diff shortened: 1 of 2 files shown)


cgolubi1:
Hmm.  So IMO it would be better not to solve this issue yet for new games, than to use timestamps to guess which games are new.  That never ends well.

blackshadowshade:
Okay, logic for new games removed. 'Fixes' comment changed.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1012/ (if it passes)

cgolubi1:
Thanks.  This looks pretty reasonable to me, let me check it out in virtualbox.

cgolubi1:
Looks good.  Merging.

PR #1859, opened 2015-12-15
Set cookies to almost never expire

### LINKED ISSUE #1745: Stay logged in even when browser closes and reopens
It would be nice for the browser to remain logged in even when it is closed and reopened, so that users don't need to log in each time.

blackshadowshade:
Fixes #1745.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1015/


cgolubi1:
Can you link to the docs or a stackoverflow response that you're using to determine that this resolves the issue?  That is, i infer that your claim is that setting an explicit expiry date overrides some sort of end-of-session checking (as opposed to both the expiry date and the end-of-session checking being active).  Can you point to some docs which support your read of the situation?

blackshadowshade:
http://stackoverflow.com/questions/3290424/set-a-cookie-to-never-expire

AdmiralJota:
I would suggest refreshing the cookie to time() + one year in auth_session_exists() as well, after the cookie has been validated. That way, it'll only make you log in if you're away for a year (as opposed to making you log in again once a year every year, regardless of how often you're on the site).

cgolubi1:
Jota: i might be totally wrong about this, but my experience with RandomAI testing has been that for whatever reason, creating `$auth_key` is one of the more computationally-expensive things the site seems to do.

When i test with RandomAI, because i'm testing both sides of the match and haven't bothered to fudge the cookies, the bot has to login before every game action.  After some experimentation with, umm, i can't remember what it's called, the phpunit-compatible performance testing module, i started applying this diff whenever i want to run automated tests:
```diff
diff --git a/src/api/api_core.php b/src/api/api_core.php
index df1a143..cce6ee0 100644
--- a/src/api/api_core.php
+++ b/src/api/api_core.php
@@ -32,7 +32,8 @@ function login($username, $password) {
         $status = $result['status'];
 
         // check if the password is correct and if the account is in active status
-        if (($password_hashed == crypt($password, $password_hashed) && ($status == 'ACTIVE'))) {
+//        if (($password_hashed == crypt($password, $password_hashed) && ($status == 'ACTIVE'))) {
+        if ((($status == 'ACTIVE'))) {
 
             // if the user has too many active logins (allow 6), delete the oldest
             $sql = 'SELECT id FROM player_auth WHERE player_id = :id ORDER BY login_time';
@@ -46,7 +47,8 @@ function login($username, $password) {
             }
 
             // create authorisation key
-            $auth_key = crypt(substr(sha1(mt_rand()), 0, 10).$username);
+//            $auth_key = crypt(substr(sha1(mt_rand()), 0, 10).$username);
+            $auth_key = substr(sha1(mt_rand()), 0, 10).$username;
 
             // write authorisation key to database
             $sql = 'INSERT INTO player_auth (player_id, auth_key) VALUES (:id, :auth_key)';
```

That removes the calls to `crypt()`, and it makes tests run something like 30-50% faster.

I bring this up because, if we decided to rewrite the cookies every session whether we needed them or not, we'd presumably need to generate another auth key to go with them, and subject site users to that slowdown.  It'd be less noticeable for players than it is for RandomAI, but i still don't think it's a good idea.

cgolubi1:
I'd rather keep things simple and just accept that people will need to login once a year.  That's not so bad.  But if y'all prefer updating the cookies once a month (e.g. detecting their current age) or reusing auth keys (eww), we can talk about those things.  Marking this incomplete because of the discussion.

blackshadowshade:
I'm also fine with having users log in once a year as necessary.

cgolubi1:
> I'm also fine with having users log in once a year as necessary.

I think that's the simplest solution.  Let me see if Jota has any objections, and if not, i'll test this and make sure it works as advertised (i can quit and restart my browsers and stay logged in), and then merge it.

AdmiralJota:
Why would you need to generate a new auth_key? You already have the auth_key for their current cookie, right? All you need to do is to use that value and update the expiration date.

AdmiralJota:
E.g.: https://gist.github.com/AdmiralJota/545470d018cccae12743

I mean, having to log in once a year isn't the end of the world, but if we can easily avoid it, it seems like we might as well.

irilyth:
A contrary suggestion: If cookies never expire, then if someone steals your cookies, they can keep logging in as you forever, right?

My two cents would be to have cookies expire after thirty days, which seems reasonably common.

(The ticket that this is a fix for (#1745) was about closing and reopening your browser, not about expiration time... And my two cents is that your cookies should probably clear when you close your browser, but if we've collectively concluded that we're not sufficiently high-security to need that, that doesn't seem terrible to me.)

AdmiralJota:
>A contrary suggestion: If cookies never expire, then if someone steals your cookies, they can keep logging in as you forever, right?

If we're concerned about people stealing cookies, then we should give users a way to delete all their existing auth_keys from the database. And that's true whether cookies expire after a month, after a browser session, after a year or never. If we think that cookie-stealing is an issue, then there's no reason to say, "If you discover someone has stolen your cookies and is logging in as you, we're OK with letting them continue doing that for the rest of the month, <i>but no longer than that</i>!"

>My two cents would be to have cookies expire after thirty days, which seems reasonably common.

Eeeagh! Nooooooooooo!

Thirty days of <i>non-use</i>, I would be completely in support of. But not <i>every</i> thirty days, no matter how much you visit the site in that time...

irilyth:
I guess I don't entirely understand what's going on behind the scenes here, but the scenario I was thinking of is one where someone steals your cookies, and you *don't* discover it. If they expire in thirty days, at least the thief's ability to mess with you eventually runs out.

I don't really care about this all that much, though. My browser knows my password, so I don't care if I have to log in once a week, once a day, once every thirty days, whatever. I'm never going to log in from a friend's computer, or an library computer, or something, so no one's going to steal my cookies that way. I think that's the sort of scenario that auth cookie expiration is meant to deal with, though, when you're using a public computer and don't want the next user to be able to access your stuff (and don't necessarily want to have to remember to click logout, clear cookies, restart the browser, etc).

blackshadowshade:
So, where are we with this pull request? Do we let it through, and say that we'll allow forced expiration in the future? Or do we not let it through until we have the option of forced expiration? Or do we add some sort of checkbox to the login screen asking whether the computer should remember this login?

dwvanstone:
It seems the "remember me?" checkbox (default not checked?) is the nicest way to allow people to play on a public machine (and not have cookies) and on a prive machine (and have cookies). One year vs. one month doesn't matter as much to me.

blackshadowshade:
So, I'd like to alter this pull request so that we have a checkbox that allows someone to stay online (with a one-year cookie that is renewed if it's older than a month). Any objections to this plan of action? Or anything that I'm missing?

cgolubi1:
That's fine with me --- i'm not sure the checkbox is necessary (won't people logout on public machines if they care?  and won't logging out kill your cookies no matter what else we do?), but i don't object to it if others like it better that way.

AdmiralJota:
Sounds good to me.

irilyth:
> That's fine with me --- i'm not sure the checkbox is necessary (won't people logout on public machines if they care? and won't logging out kill your cookies no matter what else we do?), but i don't object to it if others like it better that way.

I think the scenario is that someone might forget to log out or clear their cookies or whatever, and we don't want to leave them vulnerable in that case. If someone clicks "keep me logged on this computer for a year", on a public computer, and *then* forgets to log out, there's not much we can do to help them. :^)

So, this all sounds fine to me too.

blackshadowshade:
I intend to implement #1865 too, so it will be possible to kill off other sessions, just not immediately.

blackshadowshade:
So, the front-end has been adapted to add the new checkbox (which required a bit of rearranging header elements). The rest of the changes propagate the checkbox setting to the backend, which now sets the expiry time of the cookies explicitly. Since we don't have access to the expiry time of cookies (see http://stackoverflow.com/questions/4203225/how-to-get-cookies-expire-time), I've added a new cookie which has the expiry time explicitly present.

Also present is Jota's suggestion, which is that we update the expiry time on each login (manual or automatic).

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1110/ (if it passes)

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1115/ (if it passes)

blackshadowshade:
This doesn't hit any unit tests, so a new Jenkins run shouldn't be necessary here.

src/api/api_core.php
```diff
@@ -54,13 +54,9 @@ function login($username, $password) {
...
 
-            // set authorisation cookie
-            setcookie('auth_userid', $result['id'], 0, '/', '', FALSE);
-            setcookie('auth_key', $auth_key, 0, '/', '', FALSE);
-            session_regenerate_id(TRUE);
-            $_SESSION['user_id'] = $result['id'];
-            $_SESSION['user_name'] = $result['name_ingame'];
-            $_SESSION['user_lastactive'] = time();
+            update_session($result['id'], $result['name_ingame']);
+            set_authorisation_cookies($result['id'], $auth_key, $doStayLoggedIn);
```
cgolubi1: Hmm, so in the previous code, we set the user's cookies first, and set the current session variables second.  Now the order is reversed.  Any particular reason?  I don't offhand see a problem necessarily (all of this state is stuff that belongs to the user that we're not validating or trying to store in a database), but i need to think about it a little more.

src/api/api_core.php
```diff
@@ -111,6 +104,26 @@ function auth_session_exists() {
...
+    if ($doStayLoggedIn) {
+        // an expiry time of one year in the future should be sufficient
+        $expiryTime = time() + 365 * 24 * 60 * 60;
+    } else {
+        $expiryTime = 0;
+    }
+
+    setcookie('auth_userid', $userId, $expiryTime, '/', '', FALSE);
+    setcookie('auth_key', $key, $expiryTime, '/', '', FALSE);
+    setcookie('expiry_time', $expiryTime, $expiryTime, '/', '', FALSE);
```
cgolubi1: IMO it would be less confusing if we named this cookie something like `stay_logged_in`, to make it clear what it's for, and that we're the ones controlling it.  As is, i had some confusion while reading the code, thinking maybe `expiry_time` was a magical cookie name, because of the reference to the `$expiryTime` variable on other cookies.

src/api/api_core.php
```diff
@@ -137,6 +150,7 @@ function logout() {
 
         setcookie('auth_key', '', time()-3600, '/');
         setcookie('auth_userid', '', time()-3600, '/');
+        setcookie('expiry_time', '', time()-3600, '/');
```
cgolubi1: Yeah, i definitely think that you should rename the `expiry_time` cookie and give it a value of true or false somehow (as strings would be fine) --- There are two completely distinct comparisons going on, right:
* comparisons involving the cookie named `expiry_time`, which we compare against a value we specify for the purpose of setting a user-defined boolean, and we can do that any way we like because it's our code
* comparisons involving the parameter `$expiryTime`, which nominally are in some fashion time-based, and are set by browser interpretations of the cookie protocol implementation.  This we're trying to reverse-engineer enough to make it do what we want

Naming our cookie `expiry_time` is confusing, because it conflates those two things, one which we have control over and one which we don't.  So we should name our cookie something else, and make its value a boolean (or string representation of a boolean, whatever works more straightforwardly) because that's what it is.

src/ui/gui.css
```diff
@@ -118,7 +123,6 @@ table.spaced {
 
 table.dice {
     padding: 0px;
-    spacing: 0px;
```
cgolubi1: What does this change do in the context of this pull?

cgolubi1:
So this change adds vertical whitespace to the header bar:

Before:

less_whitespace

After:

more_whitespace

The reason is that you end the `<p>` after "Create an account", which means that it ends in the middle of the header block, whereas previously "Create an account" was at the end of the header block, and thus was the right place for the paragraph to end.  Let me know if it's not obvious how to fix it, and i can help deep-dive more.

blackshadowshade:
I'm pretty sure that I added the extra whitespace on purpose. It only shows up when the person isn't logged in, so we're not consuming whitespace that we don't have, and it makes accidental clicking/tapping less likely when someone is trying to enter their password.

cgolubi1:
> I'm pretty sure that I added the extra whitespace on purpose. It only shows up when the person isn't logged in, so we're not consuming whitespace that we don't have, and it makes accidental clicking/tapping less likely when someone is trying to enter their password.

Okay.  I don't like it --- it moves the bar up and down when you login, which is visually distracting --- and it's not necessary for this pull.  We can put it aside for now, but we'll have to make a decision before accepting this.

cgolubi1:
Another visual glitch --- when you fail to login, there's no horizontal whitespace between the error message and the "Create an account" message:
no_horizontal_whitespace

src/ui/gui.css
blackshadowshade: I'm pretty sure I removed this because there is no CSS tag called "spacing". It's possible that NetBeans was giving me a warning to that effect, but I can't remember exactly.

cgolubi1:
Okay, testing the actual desired functionality.

Before this pull, in Safari:
* Login
* Quit the browser
* Start the browser again
* Browse to `http://localhost:8080/ui/`
After that, i am not logged in.

With this pull, in Safari:
* Login without checking the box
* Quit the browser
* Start the browser again
* Browse to `http://localhost:8080/ui/`
After that, i am not logged in.

With this pull, in Safari:
* Login with checking the box
* Quit the browser
* Start the browser again
* Browse to `http://localhost:8080/ui/`
After that, i am logged in.

So checking the box has a testable effect.  Good.

cgolubi1:
Looking at the treatment of player_auth entries in the database - so far, this looks good:


Before the change:
* login - player_auth entry is created
* logout - player_auth entry is deleted
* login - player_auth entry is created
* quit browser
* login - new player_auth entry is created
* logout - new player_auth entry is deleted (the old orphaned one remains)

After the change:
* login w/button - player_auth entry is created
* logout - player_auth entry is deleted
* login w/button - player_auth entry is created
* quit browser
* restart - logged in, player_auth entry still there
* logout - player_auth entry is deleted

After the change:
* login no button - player_auth entry is created
* logout - player_auth entry is deleted
* login no button - player_auth entry is created
* quit browser
* login - new player_auth entry is created
* logout - new player_auth entry is deleted (the old orphaned one remains)

blackshadowshade:
I don't consider the movement of the bar to be a problem, considering that 
(i) most of our players will be staying logged on, and that 
(ii) the contents of the header change considerably.
However, if you insist on maintaining the position of the bar, then I will remove the extra line.

src/ui/gui.css
cgolubi1: Not sure what to do about this --- it adds a bunch of visual testing of post-login behavior to this pull.  What's easier for you, if i add that testing time onto this pull, or if we do it separately?  I'm happy either way.

blackshadowshade: Up to you, let me know if you want it left in or taken out, I'm fine with both.

cgolubi1:
Okay, i'm out of time for today.  At a glance (looked at code, no testing yet), the new cookie name looks good --- thanks.

What's left:

* Redo my simple functionality test with the new cookie names
* Regression-test the new user creation workflow, both in terms of functionality and visuals
* Test with a single account logged in across two or three browsers --- various combinations of the box being checked or not, and restarting each browser
* Do the simple test with a browser other than Safari and make sure it works there
* Do a timed test of 100 random_ai games, and make sure there's no measurable slowdown in gameplay time (which i know from previous measurement is somewhat weirdly dominated by login time --- it doesn't look from the code like this pull request should change that in either direction, but let's double-check that it at least doesn't make it worse)
* Shadowshade to fix the horizontal whitespace on failed login
* Decide what to do about the new vertical whitespace --- i'm willing to have my position portrayed as "insisting", so let's go with that.  `:>)`
* Do some game mat testing in Firefox and Safari to make sure the apparently-nonexistent CSS `spacing` parameter is not secretly being used by browsers in a way we care about.

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1176/ (if it passes)

blackshadowshade:
Sorry, I missed your comment on the vertical whitespace. I'll address that now.

cgolubi1:
Before and after header HTML, for shadowshade's info.  In both cases, this is the inner contents of the `c_header` div:

Before:
```html
<div id="login_header">
  <p>Welcome to Button Men: You are not logged in. <form action="javascript:void(0);" id="login_action_form">Username: <input name="login_name" id="login_name" type="text"> Password: <input name="login_pass" id="login_pass" type="password"> <button id="login_action_button">Login</button><font> or <a href="create_user.html">Create an account</a></font></form></p>
</div>
<hr id="header_separator">
```

After:
```html
<div id="login_header">
  <p>Welcome to Button Men: You are not logged in. <a href="create_user.html">Create an account</a></p>
  <div class="login"><form action="javascript:void(0);" id="login_action_form">Username: <input name="login_name" id="login_name" type="text"> Password: <input name="login_pass" id="login_pass" type="password"> <button id="login_action_button">Login</button> <input name="login_checkbox" id="login_checkbox" type="checkbox">Keep me logged in</form></div>
</div>
<hr id="header_separator">
```

blackshadowshade:
So, I've fixed the CSS to get the correct padding around the login form div.

I've chosen to stick to the HTML that I wrote to avoid putting a <form> inside a <p>, while leaving the existing logic in Login.arrangeHeader() untouched. I have also removed the empty <form> that was being added to the logged-in page header <p>.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1177/ (if it passes)

cgolubi1:
> I've chosen to stick to the HTML that I wrote to avoid putting a <form> inside a <p>, while leaving the existing logic in Login.arrangeHeader() untouched. I have also removed the empty <form> that was being added to the logged-in page header <p>.

I need you to stop compulsively refactoring code that you are not able to test.  I don't know what to do about this, and i am pretty close to quitting this project over it, because it is very clear that my choices are (1) simply drop two hours of testing per line of code making a change i don't care about, and have the cost of *any* comment i make on a pull request being that you will make a couple of additional arbitrary LOC changes which cost me extra hours, (2) spend a lot of time having acrimonious fights about whether you should make changes i don't care about, (3) permanently be the bad guy who is preventing us from having UI improvements.

You tell me what i should do about this, given that we have agreed that our UI unit tests are not good enough for us to be able to safely refactor anything without extensive manual testing.

blackshadowshade:
I know that your expectations for UI performance greatly exceed my expectations. I guess that I feel that the bar that you are imposing on UI changes is unreasonably high.

Would you would be willing to forego your rigorous testing on UI changes, and simply test that the functionality as advertised actually works without breaking anything obvious, assuming that we have agreed on what the required functionality should be?

For the last change specifically, you asked that the dividing line not move. In my testing, I believe that I have achieved that goal. I agree that in doing so, I have also changed the code in a non-trivial way to make it XHTML compliant, because it wasn't previously, and that has always been important to me---as noted by the header on the top of all the html files that has been there since the beginning:

> This document should be written in polyglot HTML5/XHTML5, and therefore should run when served either as text/html or application/xhtml+xml.

I agree that my approach might cause unanticipated changes in behaviour, but I am been willing to live with the consequences of that, especially since we do have testers that we can ask to test UI changes. Sure, they'll possibly miss stuff, but probably not the crucial stuff that breaks gameplay. And it's not like one of the dev team (and I will put my hand up if no one else can) isn't going to be able to fix urgent bugs pretty quickly.

So, is there some way to make the testing less time-consuming and stressful for you, some way for you to be able to not take responsibility for undesirable UI changes that you might not pick up, while still having you do a coarse accept/reject of the pull requests?

At the end of the day, we're not talking about mission-critical systems here. It's a game, and sure, everyone wants to have a continually good experience, but we dealt with the old site with its many bugs and bonefishing and we still had fun as a community.

blackshadowshade:
I should mention that I feel differently about changes that could potentially corrupt the data in the database---there, we are talking about stuff that is fundamental to the site, and there, I agree completely that we need to be completely convinced that there's nothing nasty going on.

cgolubi1:
Okay.  So how do i unblock you here?  IMO the right answer is to actually implement the tests in `test_Login.js`.  Then we can see what visible effects your refactoring is having on the header's page elements.

So my suggestion is:
* I'll submit a pull for fleshing out those Login.js tests (with no change in code at all)
* Then you can merge that pull, and make the tests continue to pass after your refactor

That will give us something to look at which isn't so time-consuming for me, which can tell me what changes your refactor is introducing.

blackshadowshade:
Sounds good.

cgolubi1:
Marking this as incomplete for tracking, because after #2059 is merged, this will have to be updated to include those tests as noted above.

blackshadowshade:
I've merged in the QUnit tests and altered them to reflect the changes in the code. Let me know if you require additional QUnit tests.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1190/ (if it passes)

src/api/ApiResponder.php
```diff
@@ -565,7 +565,15 @@ protected function get_interface_response_editForumPost($interface, $args) {
 
     protected function get_interface_response_login($interface, $args) {
         assert(!is_array($interface));
-        $login_success = login($args['username'], $args['password']);
+
+        if (isset($args['doStayLoggedIn'])) {
+            $doStayLoggedIn = ('true' == $args['doStayLoggedIn']);
+        } else {
+            $doStayLoggedIn = FALSE;
+        }
```
cgolubi1: Nit: i'd be fine with the shorter:
```php
    $doStayLoggedIn = (isset($args['doStayLoggedIn']) && 'true' == $args['doStayLoggedIn']);
```

src/api/api_core.php
cgolubi1: Responding to my previous comment: the obvious answer here is to ask, since our goal is to keep the session and the cookies in sync (i think, anyway), is either of updating the session or setting the authorization cookies likely to fail a fraction of the time we care about at all?  If so, that one should happen first.

src/ui/js/Login.js
```diff
@@ -255,44 +261,55 @@ Login.stateLoggedOut = function(welcomeText) {
...
+  loginExisting.append('Username: ');
+  loginExisting.append($('<input>', {
     'type': 'text',
     'id': 'login_name',
     'name': 'login_name',
   }));
-  loginform.append(' Password: ');
-  loginform.append($('<input>', {
+  loginExisting.append(' ');
+  loginExisting.append('Password: ');
```
cgolubi1: Why split space from the password string?  Just curious.

test/src/ui/js/test_Login.js
```diff
@@ -271,29 +270,44 @@ test("test_Login.stateLoggedOut", function(assert) {
...
+  assert.deepEqual(formElements, expectedForm, "Login.formElements is set correctly after no login activity");
+
   Login.status_type = Login.STATUS_ACTION_SUCCEEDED;
   Login.stateLoggedOut('example welcome text');
   assert.equal(Login.logged_in, false, "Login.logged_in is set to false");
   assert.equal(Login.form, Login.formLogin, "Login.form is set correctly");
   msgProps = BMTestUtils.DOMNodePropArray(Login.message[0]);
-  expectedMessage[2][1] = [ "FONT", { "color": "green" }, [ "Logout succeeded - login again?" ] ];
+  expectedMessage[2][1] = [ "FONT", { "color": "green" }, [ "Logout succeeded - login again? " ] ];
   assert.deepEqual(msgProps, expectedMessage, "Login.message is set correctly after successful logout");
```
cgolubi1: So you want to also test the form in this case, now that you've split it from the message.

test/src/ui/js/test_Login.js
```diff
@@ -302,7 +316,7 @@ test("test_Login.stateLoggedOut", function(assert) {
   assert.equal(Login.form, Login.formLogin, "Login.form is set correctly");
   msgProps = BMTestUtils.DOMNodePropArray(Login.message[0]);
   expectedMessage[2][1] = [
-    "FONT", { "color": "red" }, [ "Login failed - username or password invalid, or email address has not been verified" ] ];
+    "FONT", { "color": "red" }, [ "Login failed - username or password invalid, or email address has not been verified. " ] ];
   assert.deepEqual(msgProps, expectedMessage, "Login.message is set correctly after failed login");
```
cgolubi1: Ditto

cgolubi1:
Okay, a spot check of a game mat shows no difference that i can detect based on the removal of "spacing" from table.dice, so let's go ahead and leave that change in.

cgolubi1:
Did a couple of timing tests of repeated-login-heavy RandomAI --- the new code definitely isn't faster, but it doesn't look like it's significantly slower.  Each of these data points is end-to-end timing of nine Avis vs Avis games:
```bash
# old code
real: 1m4.519s  user: 0m1.976s sys: 0m0.542s
real: 1m7.945s  user: 0m2.144s sys: 0m0.541s
real: 1m4.658s  user: 0m2.024s sys: 0m0.495s

# new code
real: 1m5.107s  user: 0m2.084s sys: 0m0.504s
real: 1m10.517s user: 0m2.280s sys: 0m0.536s
real: 1m6.312s  user: 0m2.162s sys: 0m0.510s
```

So i don't think those differences, which may be coincidence anyway, are anything to write home about.  That seems fine.

cgolubi1:
Okay, so, i commented on a couple of nits above, which you can take or leave.

And i do want you to add testing of `Login.formElements` in the two other logged-out cases (logout successful and login failed), to show that those work the same way in all three cases and guard against future regressions in that.

But otherwise this seems good to go.

cgolubi1:
One other thing: the green "Logout successful.  Login again?" text does not appear to actually show up in the header.  But i tested against the base code, and you didn't introduce this regression - it was already there.  So i'm going to open a separate issue for that.

blackshadowshade:
Right, I've addressed all the nits and added QUnit tests as requested.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1195/ (if it passes)

cgolubi1:
Thank you kindly.  Merging.

Issue #1802, opened 2015-11-11
Konstant dice that contribute negative values should also contribute negative fire values


blackshadowshade:
In #1717, Chaos found that konstant dice that are contributing negative values contribute *positive* fire values when another die with Fire is turned down. This is a problem that occurs with the buttons Hawaii and Arizona, as well as RandomBMMixed.

There is also a responder test in #1717.

cgolubi1:
I'm going to argue that this is a blocker for merging #1861 --- right now, there are no named buttons containing both Konstant and Fire, but #1861 will enable Arizona and Hawaii.

blackshadowshade:
I'm working on this at the moment. Just thought that I'd mention to @cgolubi1 that I've created a responder test for this (based on a previous responder test involving Hawaii), so you won't need to worry about that.

jl8e:
I wish you well. I know I tried to figure out how to fix it without a rewrite of the fire logic and got nowhere. Definitely a mistake on my part way back when.

jl8e:
Of course, right after posting that, my brain went "But what about...."

I have thoughts if you want them.

blackshadowshade:
Thoughts are always welcome, Julian!  👍 

I worked this out before, but didn't write it down, and so I've forgotten again. I'll work it through again. However, it's something to do with splitting the logic out in one of the functions related to help values, I believe.

jl8e:
There's probably several ways to approach it.

The way I was looking at was mostly in collect_firing_maxima. If, instead of a specific number, each die gave it a array of [point cost, current value, new value] entries, you could then test fire options by subtracting current_value and adding new_value.

(so, a normal die would return [ [1, 5, 6], [2, 5, 7], [3, 5, 8]], and a konstant would return [[1, 5, 6], [1, -5, -6], [2,5,7]...])

Since it's only skill attacks that care, there could be an additional method on each die to return the complex firing maxima info, and have skill attacks override the normal collect_firing_maxima to use it, which isolates all the changes to the attack code.

blackshadowshade:
My approach was going to be directly altering BMAttackSkill->help_bounds() to produce an array of possible help values, which would not necessarily be contiguous.

These help values would be calculated with the nuance that konstant can contribute either negative or positive, probably not using hooks in a first iteration. If I get the logic working, I can always farm the relevant bit out to the die skill using hooks later.

blackshadowshade:
[...]

PR #1861, opened 2015-12-15
Reactivated 50 States buttons

blackshadowshade:
Requires database update activate_50_states_01.sql.

I think that this will need extensive testing by the testers on the dev site before we merge this apparently trivial change, since the 50 states buttons are somewhat odd.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1016/


cgolubi1:
Yeah, why throw this over the fence now?  Is your goal in fact to have this go live in January?  If so, we'd better talk about how to get people to do dev site testing.  Here's my proposal if you really want to do this:
* I'll kick off a bunch of RandomAI games using 50 States buttons to try to provoke internal errors
* Any i find, we file bugs, and then disable those buttons again as part of this pull --- the bugs will be fixed and the buttons enabled later than January
* We push this to the dev site, and offer a quest for people to play N games (we can pick N) using these buttons
* Any logic bugs we find, we file bugs, and then disable buttons with those skill combinations again --- the bugs will be fixed and the buttons enabled later than January

If you feel strongly about this, we can do it; it's a good way to get people onto the dev site to look at other things.

But this plan as written strikes me as likely to lead to me spending all of my BM time in December reproducing bugs.  As you well know, when someone reports a bug, it is *not* fast to get a clean example of the bug.  I don't think you want me to spend all of my BM time in December reproducing 50 States bugs, and i'm not suggesting you do, but i do think it's historically accurate that i do about 90% of the work of reproducing logic bugs so they can be fixed, whether they are initially found via automated testing or on one of the sites.  I don't want to do that during the rest of December, and the above plan could lead to it.

One possible modification is, if any problems are found or reported, blindly disable *all* 50 States buttons in the offending match without any investigation at all, and trust that people won't abuse that for whatever reason.  I'd be okay with the above plan with that modification.

What did you have in mind?

blackshadowshade:
I thought that we had agreed on the pull requests that you would accept before we hit merge freeze, and that this wasn't one of them?

I put this pull request up so that the testers have something to do in January, while I'm travelling and you're busy with your API refactor.

cgolubi1:
Ahh, okay.  Yep, that makes sense.  Happy to test it in January.  I thought we agreed that i'd try to get to the pulls we agreed on before you went away, but would keeping working on the queue at my now-usual rate.  :>)

Anyway, sounds good --- i'll work on internal-error testing as time and interest permit, but not plan to merge this until after the January release.

irilyth:
If there were an easy way, even a hackish one-off way, to let testers create games between two random buttons from the set, that'd make it much easier to test.

cgolubi1:
> If there were an easy way, even a hackish one-off way, to let testers create games between two random buttons from the set, that'd make it much easier to test.

Sure, we can think about that in January too.

blackshadowshade:
This should be doable by altering BMInterface->resolve_random_button_selection() manually.

irilyth:
We could also potentially write a little Python script for it. Not all testers have the code checked out, probably, but I bet enough do.

cgolubi1:
I'm going to kick off some preliminary hands-off testing for this pull, namely:
* Do the usual database update-vs-rebuild smoke test
* Stand up an AWS instance and set it to playing games to look for internal errors

I'd like that second thing to run for a number of days before we put this on dev, so might as well start it now.

cgolubi1:
Looking at #1802, i'd also throw out the suggestion of scrapping Arizona and Hawaii for now and trying to proceed with the others --- i don't think we're close to the end of the issues with Konstant/Fire.  But that's just my gut instinct here.

cgolubi1:
Okay, i kicked off the testing.  Incidentally, i've already hit some weirdness in random_ai's ability to handle Arizona, and have disabled Arizona from my test set.  Presumably Hawaii will hit the same issue in short order.  I'm happy to revisit later, but until we can say we believe Fk works and understand *how* we think it works, IMO there's no point.

blackshadowshade:
I'm happy for us to disable Arizona and Hawaii. Let's see which other buttons cause us trouble.

cgolubi1:
Sounds good.  So far, RandomAI has played over 700 games without internal errors, with Arizona and Hawaii disabled.  I'll let it go for a few days and see what else develops.

blackshadowshade:
Okay, Arizona and Hawaii have been re-disabled.

cgolubi1:
You're not going to like this.  So Iowa has a Konstant die and a Doppelganger die, and thus in an Iowa vs Kansas match, managed to *acquire* a Fire die and wind up in the same situation as Arizona and Hawaii did...

I do note that pairing either Konstant or Doppelganger with other complicated skills is not actually all *that* common, and it's particularly uncommon for Konstant to be paired with other skills ~~because it's so useless~~.  Doppelganger is sometimes paired with other skills --- e.g. there are buttons which have Doppelganger and T&S, so if that's scary, then there we are.  But i don't immediately see any other buttons which pair D with known-problematic other skills (except m (Lust), and we've certainly had our share of problems with that in the past).  And by-and-large buttons which have multiple D dice don't have any *other* skills, so they can't reproduce a skill combination unless that skill combination already exists on another button.

 So i'm inclined to say you should disable Iowa too, because until we solve more of the Konstant bugs, a button which can pair Konstant with any other skill in existence is likely to be dangerous.

cgolubi1:
Shadowshade, if you're online in the next couple of days and get a chance to weigh in on disabling Iowa, i might be able to push this to dev once the merge freeze is over.

blackshadowshade:
Sounds good, I'm happy to disable Iowa. I'll get on that now.

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1037/ (if it passes)

cgolubi1:
Alright, at this point, RandomAI has played 21000 games without internal errors, except for Arizona, Hawaii, and Iowa, so i'm going to shut down the RandomAI tests, do another DB build-vs-update test, and then declare this ready to merge.

cgolubi1:
My suggestion for testing is going to be to clobber random button selection on dev in the most crude way possible:
```diff
$ git diff src/engine/BMInterface.php 
diff --git a/src/engine/BMInterface.php b/src/engine/BMInterface.php
index 7bef6f5..0e33d53 100644
--- a/src/engine/BMInterface.php
+++ b/src/engine/BMInterface.php
@@ -1067,7 +1067,7 @@ class BMInterface {
             }
 
             if (empty($allButtonData)) {
-                $allButtonData = $this->get_button_data(NULL, NULL, TRUE);
+                $allButtonData = $this->get_button_data(NULL, '50 States', TRUE);
                 $nButtons = count($allButtonData);
             }
 
```

With this change, every time you create a game and ask for a random button, you get one from 50 States.  This optimizes for not spending a lot of time thinking about how to do this more cleverly.  :>)  I don't mind doing it a different way, but would prefer alternate suggestions come with sample code.  :>)  (In particular, "add a second 'Random 50 States Button' option to the pulldown" is *not* a trivial suggestion --- it could be done, but it would be nontrivial integration and debugging work, let's spend our time on something with more long-term value, please.)

cgolubi1:
No one voiced an objection to my big-hammer approach to having dev testers access random 50 states buttons by clobbering "Random Button".  So that's what i'm going to do when i push to dev.

PR #1869, opened 2015-12-17
Fixed regression when using Echo with any other button that has no recipe

### LINKED ISSUE #1868: infinite loop bug when creating game between Echo or Zero and RandomBM
infinite_loop_while_advancing

Seems to hold for Echo or Zero vs. any type of RandomBM that i tried.

blackshadowshade:
Fixes #1868.

This is a regression that crept in with pull request #1809.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1019/

### DIFF
--- a/src/engine/BMGame.php
+++ b/src/engine/BMGame.php
@@ -508,7 +508,7 @@ class BMGame {
 
     protected function update_game_state_specify_recipes() {
         foreach ($this->playerArray as $player) {
-            if (empty($player->button)) {
+            if (empty($player->button->recipe)) {
                 return;
             }
         }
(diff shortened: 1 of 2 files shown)


cgolubi1:
Okay, so, i need to:
* Verify that this works for me
* Play back a bunch of games (arguably less important)
* Play a bunch of new games (more important)

I think we need to go ahead and merge this sooner rather than later, so i'll focus on new games, since that's where we want to make sure starting games works --- but i also want to catch regressions in e.g. the number of dice that get rolled at the beginning of games, so step 2 is somewhat important as well.

cgolubi1:
Works by hand:
success

cgolubi1:
I didn't see any unexpected errors in 200 new games and 500 replay games, so i guess that's good enough.  Merging.

PR #1871, opened 2015-12-18
added loadNewGames, reactToNewGame to bmutils, bmapi python clientent tooling

danlangford:
just playing with the api i noticed the API functions `loadNewGames` and `reactToNewGame` were not yet implemented in the python tooling and they were ones i wanted to play with. i went on to also add them to the monitor but in such a specific fashion to my needs i decided not to push up those changes.

let me know if i messed up the target branches or whatever this is my first pull request into you guys 

jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-danlangford/3/ (if it passes)

### DIFF
--- a/tools/api-client/python/lib/bmutils.py
+++ b/tools/api-client/python/lib/bmutils.py
@@ -87,6 +87,18 @@ class BMClientParser(bmapi.BMClient):
       raise ValueError, "Failed to call loadActiveGames, got: " + retval
     return self._wrap_game_list_data(retval.data)
 
+  def wrap_load_new_games(self):
+    retval = self.load_new_games()
+    if not retval.status == 'ok':
+      raise ValueError, "Failed to call loadNewGames, got: " + retval
+    return self._wrap_game_list_data(retval.data)
+
+  def wrap_react_to_new_game(self, game, accept):
+    retval = self.react_to_new_game(game, 'accept' if accept else 'reject')
+    if not retval.status == 'ok':
+      raise ValueError, "Failed to call reactToNewGame, got: " + retval
+    return retval.data
+
   def wrap_load_completed_games(self):
     retval = self.load_completed_games()
     if not retval.status == 'ok':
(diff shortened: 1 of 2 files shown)


tools/api-client/python/lib/bmutils.py
```diff
@@ -87,6 +87,18 @@ def wrap_load_active_games(self):
       raise ValueError, "Failed to call loadActiveGames, got: " + retval
     return self._wrap_game_list_data(retval.data)
 
+  def wrap_load_new_games(self):
+    retval = self.load_new_games()
+    if not retval.status == 'ok':
+      raise ValueError, "Failed to call loadNewGames, got: " + retval
```
cgolubi1: This business is an old bug --- i'll try to fix it across the board, but in the meantime, fix it in new code --- you can't print retval like this because it's an object, you instead want to print `retval.message`

cgolubi1:
Otherwise, this seems okay at a glance --- i'll try out the new commands locally and see if they work for me.

cgolubi1:
Oh, you asked about branches --- i don't object to merging from a branch named "dev" as long as it stays clean, but, in general, we use branch names which describe what the code does --- the typical pattern is issuenum_some_description (e.g. `1775_randomBM_info` is a branch i'm testing right now) but `some_description` is also valid if it's not in response to an issue number.

cgolubi1:
We should also set you up on jenkins (http://jenkins.buttonweavers.com:8080/) so we can verify that your code passes the test suite before being merged, but i don't insist for this pull because the python testing jenkins does is pretty lightweight, and i can look for regressions myself in virtualbox.

https://github.com/buttonmen-dev/buttonmen/wiki/Developer-guide:-getting-started mentions jenkins.

danlangford:
ok i now print `retval.message` and i updated all of the other old bug `retval`s in that same file

cgolubi1:
I created a job for your buttonmen tree on jenkins, so i can get a build of this ftr.  Since you're just changing the python client, jenkins isn't super relevant, but it doesn't hurt.  I'll add a link, and merge this, once that passes.  Thanks for fixing all the other messages.

cgolubi1:
I added the jenkins build link (it's still in progress) to the description --- if you do more pulls, we'll get you setup with a login account to jenkins so you can kick off builds yourself.

cgolubi1:
Build succeeded, no surprise there.  Merging.

danlangford:
+1 awesome . Ya I will request a Jenkins account once I really become an
active dev on the primary code base

PR #1873, opened 2015-12-19
Changed criteria for chat editing to allow editing at start of round

### LINKED ISSUE #1517: backend and UI disagree about chat editing, leading to user-visible badness
Here's the situation:
- The most recent thing that happened was the server decided who has initiative
- Your opponent is the one who gets initiative
- It's your opponent's turn to attack
- You submit some chat
- You want to chat some more and/or edit your chat

In this case, the UI thinks that the chat you just submitted was not the most recent thing to happen (because the decision about who gets initiative is the most recent thing in the action log), so it shows you a blank box for "add/edit chat".  However, the backend thinks that your recent chat is editable, so you can't insert a new chat.  So when the UI tries to submit a new chat, the backend rejects it with "You can't add a new chat message now".

IMO, the backend is correct, and the most recent chat should be editable in this situation.

I tripped over this situation in http://www.buttonweavers.com/ui/game.html?game=3745, which i don't really want to screencap because it has live chat in it.  If it's totally opaque what i'm talking about, i can reproduce this pretty easily.

blackshadowshade:
Fixes #1517.

It is possible the change here has consequences (possibly good!) for chat editing at the end of game, or between rounds. Although this is currently not possible because we don't show the appropriate button on the UI, we *could*, if we wanted to.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1020/ (if it passes)

### DIFF
--- a/src/engine/BMInterface.php
+++ b/src/engine/BMInterface.php
@@ -2973,10 +2973,21 @@ class BMInterface {
         if (count($chatLogEntries) == 0) {
             return FALSE;
         }
-        // Only the most recent chat entry can be modified --- was
-        // it made by the active player?
-        if ((FALSE === $currentPlayerIdx) ||
-            ($playerNameArray[$currentPlayerIdx] != $chatLogEntries[0]['player']) ||
+
+        // only a player in this game can modify the last chat message
+        if (FALSE === $currentPlayerIdx) {
+            return FALSE;
+        }
+
+        // only the player who chatted last can modify the last chat message
+        if ($playerNameArray[$currentPlayerIdx] != $chatLogEntries[0]['player']) {
+            return FALSE;
+        }
+
+        // only the player who was last active can modify the last chat message,
+        // unless the game is in a state where the last activity in the game was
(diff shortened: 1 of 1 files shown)


cgolubi1:
Okay.  Your change as written appears to address the bug as written, and passes jenkins.  I'll bring this up in virtualbox and make sure it works for me.  I think i'd also like to use a modified RandomAI to collect and play back a few games with chat (typically RandomAI doesn't chat), so i can see the regressions in gameChatEditable and make sure they make sense to me.

cgolubi1:
Seems good.  I can edit chat while the initiative action is the last thing that happened.  I can't edit chat if the other player has taken an action since i loaded the page, whether or not the other player chats.

cgolubi1:
Hmm, i should probably be more diligent about this, but getting RandomAI to support game chat is a slightly annoying refactor that i'm not up for right now.  So i'm just going to declare victory on the basis of the by-hand testing, and merge this.

PR #1875, opened 2015-12-22
Refactored part of BMInterface into BMInterfaceHistory

blackshadowshade:
Continues to address part of #974.

This is a pure refactor and should not change any behaviour whatsoever.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1027/


cgolubi1:
I may start testing this if there's time, but i'm not going to consider merging it until January, because i'm not super confident that i can catch all regressions.

test/src/engine/BMInterfaceHistoryTest.php
```diff
@@ -0,0 +1,14 @@
...
+class BMInterfaceHistoryTest extends BMInterfaceTestAbstract {
+
+    protected function init() {
+        $this->object = new BMInterfacePlayer(TRUE);
+    }
+
+    public function test_dummy() {
+        // currently, there are no tests of history functionality
+    }
+}
```
cgolubi1: This file needs to end with a newline.

cgolubi1:
Okay.  So, basically, what you did here is move a bunch of functions from BMInterface to BMInterfaceHistory.  Given that, what i need to know is:
* Does that functionality still work to do searches in its new location (which i have to test by hand, because we don't yet have API-level tests (or any other) for game history?
* Did you make any changes as part of the move, and do i like them?  (I can figure this out using stupid shell tricks --- github won't tell me.)
* Did you move anything out of BMInterface that any *other* functionality needs?  (This is arguably the trickiest, but i can poke around and try to figure out what i think, anyway.)

cgolubi1:
Well, other than the public `search_game_history()`, which is invoked by ApiResponder, none of the 18 functions moved is referenced anywhere except within BMInterfaceHistory itself.  That's a good sign that this is a piece of fairly isolated functionality.

cgolubi1:
Yeah, i dunno, as far as i can tell, this just works and is totally simple.  I know i said i wasn't going to merge it until January, but IMO the risk of doing so is small, and the benefit is we don't have to keep the pull floating around, and shadowshade can keep building other refactors on top of it.  So i'm going to go ahead and merge it.

Issue #1858, opened 2015-12-15
all API tests should work the way the loadGameData test works


cgolubi1:
The loadGameData API tests work by having responderTest run live tests against the backend and cache JSON response files, which are then passed by dummy_responder.php to the UI when UI testing is called for.

All other API tests should work the same way, to keep the test response data which is being used by the frontend in sync with what backend tests show the API is currently providing.

PR #1879, opened 2015-12-27
More BMPlayer refactoring

### LINKED ISSUE #1608: BMGame has too many properties

blackshadowshade:
Fixes #1608.

This is the second half of the BMPlayer refactor, and it completes the work that I have envisaged. As usual, the refactor is intended to cause no player-visible changes. However, there are quite a few changes in the way that the code initialises various properties, so this will need rigorous testing.

The major changes to the BMGameTest code are caused because the code now requires the game to be accepted by both players before it will progress past the early game states. Previously, if the game happened to have a nonexistent BMGame->hasPlayerAcceptedGameArray, it would also progress.

The only possible outstanding work that I can see is to update all the tests that depended on BMGame array properties. However, the convenience accessors that BMGame provides of arrays of BMPlayer properties all seem to work correctly, and so all the unit tests are working fine.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1031/ (if it passes)


deploy/amp/database_initialise.php
```diff
@@ -5,12 +5,12 @@
 <body>
     <?php
 
-        recreate_database('buttonmen');
-        recreate_database('buttonmen_test');
-
-        $conn = db_connect('buttonmen');
-        run_all_sql_files($conn);
+// james: uncomment the following lines to also initialise the interactive database
```
cgolubi1: Nit: it's obvious to me what you mean by "interactive database", but i'm not sure it would be obvious to someone just starting out development, who's the target audience for this comment.  I'd say something more like "to (re)initialize the primary 'buttonmen' database", or something.  Don't feel strongly about the wording --- i mostly think "interactive" is confusing because "interactive database" could also mean "interactive PHP interface to the database".

src/engine/BMGame.php
```diff
@@ -437,17 +342,17 @@ protected function update_game_state_start_game() {
         $this->nRecentPasses = 0;
         $this->autopassArray = array_fill(0, $this->nPlayers, FALSE);
         $this->fireOvershootingArray = array_fill(0, $this->nPlayers, FALSE);
-        $this->gameScoreArrayArray = array_fill(0, $this->nPlayers, array(0, 0, 0));
```
cgolubi1: Nit: it seems like this all zeroes array was simply a bug which was harmlessly overwritten by `do_next_step_apply_handicaps()` to have the correct keys.  Maybe that means that one or the other of these functions doesn't *need* to zero-fill the array?

src/engine/BMGame.php
```diff
@@ -476,10 +380,13 @@ protected function do_next_step_choose_join_game() {
...
-            is_array($this->hasPlayerAcceptedGameArray) &&
-            in_array(FALSE, $this->hasPlayerAcceptedGameArray)) {
-            foreach ($this->hasPlayerAcceptedGameArray as $playerIdx => $hasAccepted) {
+        $hasPlayerAcceptedGameArray = array_fill(0, $this->nPlayers, NULL);
+        foreach ($this->playerArray as $playerIdx => $player) {
+            $hasPlayerAcceptedGameArray[$playerIdx] = $player->hasPlayerAcceptedGame;
+        }
+
+        if (in_array(FALSE, $hasPlayerAcceptedGameArray)) {
+            foreach ($hasPlayerAcceptedGameArray as $playerIdx => $hasAccepted) {
```
cgolubi1: So, one argument i could make here is: in the new refactored world, any time you're accessing `$playerIdx` inside BMGame, it's worth thinking about why, because why can't you just act on each `$player` object without having to futz with indices?

And in this case, i think that question is legit --- in this case, the only reason you're pulling `$this->playerArray` with indices is so you can construct your own temporary array.  But i'm not sure you need that --- IMO it would be tidier to separate the "find out if any player has not accepted" part from the "set each player's waitingOnAction" part.  Because the latter part you can do by setting:
```php
$player->waitingOnAction = !$player->hasPlayerAcceptedGame
```

There's no need to reference a BMGame temporary variable for that.

cgolubi1: Worth noting: i'm being intentionally picky here to see if you agree with my thinking about this.

src/engine/BMGame.php
```diff
@@ -623,23 +526,20 @@ protected function load_swing_values_from_previous_round() {
...
+                continue;
+            }
 
-                foreach ($optionValueArray as $dieIdx => $optionValue) {
-                    $die = $this->playerArray[$playerIdx]
-                                ->activeDieArray[$dieIndicesWithoutReserve[$dieIdx]];
-                    if (!($die instanceof BMDieOption)) {
-                        throw new LogicException('Die must be an option die.');
-                    }
+            $dieIndicesWithoutReserve = $this->die_indices_without_reserve($playerIdx);
```
cgolubi1: Again, here, if `die_indices_without_reserve()` takes a mandatory player argument and acts just on the dice of that player, why not move it to `BMPlayer`?

src/engine/BMGame.php
```diff
@@ -939,8 +821,10 @@ protected function roll_active_dice_needing_values() {
 
     protected function update_game_state_specify_dice() {
         if (!$this->isWaitingOnAnyAction()) {
-            $this->prevSwingValueArrayArray = NULL;
-            $this->prevOptValueArrayArray = NULL;
+            foreach ($this->playerArray as $player) {
+                $player->prevSwingValueArray = array();
+                $player->prevOptValueArray = array();
+            }
```
cgolubi1: Why the change from NULL to array()?  Does this fix a bug?

src/engine/BMGame.php
```diff
@@ -2798,10 +2655,14 @@ public function __set($property, $value) {
         }
 
         $subProperty = substr($property, 0, strlen($property) - 5);
-        $isSubPropertyArray = ('Array' == substr($subProperty, -5));
+        $isDieArray = ('DieArray' == substr($subProperty, -8));
```
cgolubi1: This is starting to look a little magical.  Please add a comment to this block explaining what these strings are such that substringing them leads to anything good.  :>)

src/engine/BMGame.php
```diff
@@ -3266,27 +3081,26 @@ protected function get_playerDataArray($requestingPlayerIdx) {
...
                 'capturedDieArray'    => $this->get_capturedDieArray($playerIdx),
                 'outOfPlayDieArray'   => $this->get_outOfPlayDieArray($playerIdx),
                 'swingRequestArray'   => $this->get_swingRequestArray($playerIdx, $requestingPlayerIdx),
-                'optRequestArray'     => $this->get_optRequestArray($playerIdx),
-                'prevSwingValueArray' => $this->get_prevSwingValueArray($playerIdx),
-                'prevOptValueArray'   => $this->get_prevOptValueArray($playerIdx),
+                'optRequestArray'     => $player->optRequestArray,
+                'prevSwingValueArray' => $player->prevSwingValueArray,
+                'prevOptValueArray'   => $player->prevOptValueArray,
                 'waitingOnAction'     => $this->get_waitingOnActionArray($playerIdx, $requestingPlayerIdx),
```
cgolubi1: So my question above applies here: why are all these accessors still in BMGame rather than moving to BMPlayer?

src/engine/BMGame.php
```diff
@@ -4084,7 +3851,7 @@ protected function setBMPlayerProps($property, array $value, $subclass = '') {
     /**
      * Set all players' waitingOnAction statuses to FALSE
      */
-    protected function setAllToNotWaiting() {
+    public function setAllToNotWaiting() {
```
cgolubi1: Arguably this is a symptom that there's way too much game logic in BMInterface.  My first cut at this would be to say that anywhere BMInterface is directly calling `$game->log_action()`, that's a symptom that a BMGame function is needed to handle that case.  (At a glance, that seems right --- BMGame calls log_action() 12 times, while BMInterface calls it 5 times.  That matches my instinct that the 12 times are correct and the 5 times are wrong.)

On the one hand, that's orthogonal to this pull, so you should create a new issue and not handle this here.  On the other hand, you're making a function public which shouldn't be public, in this pull, because of that issue.  So i'm torn about whether it should be fixed now.

src/engine/BMInterface.php
```diff
@@ -1074,7 +1050,7 @@ protected function resolve_random_button_selection(BMGame &$game) {
             $randIdx = bm_rand(0, $nButtons - 1);
             $buttonId = $allButtonData[$randIdx]['buttonId'];
 
-            $this->choose_button($game, $buttonId, $buttonIdx);
+            $this->choose_button($game, $buttonId, $playerIdx);
```
cgolubi1: Pass `$player` instead of `$playerIdx`, and have the database insert use `$player->position`.  Then you don't need to reference `$playerIdx` here.

cgolubi1:
Okay, i'm out of time for today.  I don't see any reason *not* to be picky about this change --- it's a refactor and is going to need a lot of testing, so it might as well be the best refactor it can be since i'm already doing the testing.  Anyway, i'll pick up reading through the code tomorrow in `BMInterface::does_need_random_button_selection()`, and once i'm done with that, i'll figure out how i'm going to test it.

At a glance, the big automated testing difficulty is going to be that this change affects pre-created game state for which we don't have a lot of regression tests.  It's *probably* not right to twiddle my thumbs on this until i've both finished #1858 and gone on to add more API regression tests of pre-game states and hooked up RandomAI to be able to use those arbitrarily, because that's going to be awhile.  So i may wind up testing by hand, but i'd much rather use RandomAI if i could.  I'll continue to ponder.

src/engine/BMGame.php
blackshadowshade: Actually, I think you have a good point here. If we're having to zero-fill, then we should actually be shifting these also into BMPlayer.

src/engine/BMGame.php
blackshadowshade: On a cursory glance, I agree with you. It wasn't my primary intent in this refactor to tidy up all the references to $playerIdx, but I'm happy to do so where it's obvious how that should work.

src/engine/BMGame.php
blackshadowshade: Good idea.

src/engine/BMGame.php
blackshadowshade: Yeah, this is in keeping with the change to initialisation philosophy, where now, all the arrays accessed via the convenience accessors should always exist. That gets around the problem that we had right at the beginning that we always had to check if the input was an array before doing a foreach.

src/engine/BMGame.php
blackshadowshade: But it *is* magical!  XD

src/engine/BMGame.php
blackshadowshade: If you're talking about the get_... accessors, the ones that need to stay in BMGame are those where BMGame actively lies about what's happening. I'll take a closer look, though, to see if any of these are simply passing information directly across.

src/engine/BMGame.php
blackshadowshade: I don't know that I agree that this function shouldn't be public, since it would be possible to replace the function call with the function contents, and that wouldn't break any of PHP's (broken) encapsulation. However, I'll take a closer look at it.

blackshadowshade:
I'm fine with you being picky about the refactor, especially since it won't be merged until the next release.

I'll be less online in the next week or so, so I'm not ignoring you if I don't respond immediately.  :)

src/engine/BMGame.php
cgolubi1: Yeah, that makes sense as a strategy (places where you see `$playerIdx` and/or zero-filling arrays are targets for refactoring) and a tactic (if it's easy to see what to do, do it, if not, wait for a future refactor).

src/engine/BMGame.php
cgolubi1: Huh, okay, i had somehow gotten it into my head that the issue was that foreach can't iterate over a zero-length array.  If the issue is just that foreach can't treat NULL as an array, that's substantially less terrible of PHP.  :>)  Anyway, makes sense.

src/engine/BMGame.php
cgolubi1: Yeah, hmm, right, because you don't want BMPlayer to be in charge of the lying.  In principle you might break those into a BMPlayer accessor which does the accessing, and a BMGame routine which takes the clean value given by BMPlayer and filters it based on requesting player and game state.  But i don't feel strongly.

You should add a code comment here explaining this distinction, though.

src/engine/BMGame.php
cgolubi1: Yeah, but this isn't about PHP's choices, it's about our choices --- my argument is that logically it should be BMGame, not BMInterface*, which makes the decision to set a player's state to waiting or not-waiting.  I mean, again, that's not necessarily part of this refactor, but that's why it seems like a regression to me to make the function public.

cgolubi1:
> I'm fine with you being picky about the refactor, especially since it won't be merged until the next release.
>
> I'll be less online in the next week or so, so I'm not ignoring you if I don't respond immediately. :)

Sounds good.  Mostly, i see this kind of refactor as something which has the potential to make the code a lot cleaner, so i think it's worth being picky about.

src/engine/BMInterface.php
```diff
@@ -1086,27 +1056,31 @@ protected function resolve_random_button_selection(BMGame &$game) {
     }
 
     protected function does_need_random_button_selection(BMGame $game) {
+        $hasRandomlyChosenButtons = FALSE;
+        $hasUnresolvedNames = FALSE;
+        foreach ($game->playerArray as $player) {
+            // do not resolve random names unless all players have joined the game
```
cgolubi1: Nit: "do not resolve random *button* names".  ("What's a random player name?" i asked myself on first read.)

src/engine/BMInterface.php
```diff
@@ -1385,18 +1337,19 @@ protected function save_player_with_initiative($game) {
                      'AND player_id = :player_id;';
             $statement = self::$conn->prepare($query);
             $statement->execute(array(':game_id' => $game->gameId,
-                                      ':player_id' => $game->playerIdArray[$game->playerWithInitiativeIdx]));
+                                      ':player_id' => $game->playerArray[$game->playerWithInitiativeIdx]
+                                                           ->playerId));
```
cgolubi1: Nit: i think two-line variable accesses like this are a little suspect.  Would our linewrapping allow us to format this block as something like:
```php
$statement->execute(
    array(':game_id' => $game->gameId,
          ':player_id' => $game->playerArray[$game->playerWithInitiativeIdx]->playerId));
```

test/src/engine/BMGameTest.php
```diff
@@ -669,12 +665,13 @@ public function test_do_next_step_specify_dice() {
...
         $button1->load($recipe1);
         $button2->load($recipe2);
-        $this->object->activeDieArrayArray = array($button1->dieArray,
-                                                   $button2->dieArray);
-        $this->object->swingRequestArrayArray = array(array(), array());
-        $this->object->gameScoreArrayArray = array_fill(0, $this->object->nPlayers, array(0, 0, 0));
+        $player = $this->object->playerArray[0];
+        $player->activeDieArray = $button1->dieArray;
+        $player = $this->object->playerArray[1];
+        $player->activeDieArray = $button2->dieArray;
```
cgolubi1: This is pretty nit-level for a test, but make these `$player1` and `$player2` to match `$button1` and `$button2`.

cgolubi1:
I assume you know this, but it looks like you didn't run jenkins on the most recent version of this.  So you'll need to do that at some point, obviously.  I'm inclined to kick that off now myself, because if something clearly broke, i don't want to own detecting it.

cgolubi1:
Okay, no, i'm totally confused --- i kicked off another jenkins run, http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1031/ and it's *also* running against `17e33769e726ca35a5c797c6cb7b81e2c9f1e85f`... which according to github is not the most recent version of this pull.

cgolubi1:
Okay, but, seriously folks: this has been fun and all, but how am i going to test this?
* For logic within active games, i might as well trust the tools, meaning:
** At least 3000 new random games against the final codebase to look for internal errors
** At least 3000 replayed random games against the final codebase to look for regressions
* Otherwise, i'm going to have to test everything players can do, one way or another, because any of BMInterface's searches which involve games or game attributes *could* have bitrotted.  Any of these tests which i eventually have a way to automate or can claim are already automated, great, but if not i should check by hand.  This means:
** Looking at the overview: with various numbers of games in various states
** Creating games in all valid combinations of autoaccept or not, buttons specified or not, opponents specified or not
** Setting preferences, especially automated and game-related preferences
** Using the game history to do various searches
** Looking at who's online
** Posting to the forum (should be orthogonal, but it's pretty easy to do a stub forum check)

src/engine/BMGame.php
blackshadowshade: Before, we had to do *two* checks:
- is it an array?
- is the array empty?

The shift from NULL to array() gets around the first check.

I'm not going to excuse PHP for being terrible!

src/engine/BMGame.php
blackshadowshade: Okay, let's make that another issue then: that BMInterface ideally shouldn't directly call $game->log_action().

cgolubi1: Sounds good --- created #1880

cgolubi1:
So, FYI, Jenkins was giving you checkstyle warnings for the most recent run --- http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1031/checkstyleResult/

You made another change since then, but i'm going to go ahead and mark this incomplete until you're happy that you have a version that you think passes Jenkins.  (If you can't run Jenkins from where you are, don't worry about it --- let me know and i can click the button, but i want to hear that you think it's ready to pass.)  I don't think there's any rush on this.

blackshadowshade:
Okay, I think I've addressed all your comments. Let me know if I've missed anything.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1032/ (if it passes)

cgolubi1:
Okay, what are the issues i noticed before?

* Fix comment in deploy/amp/database_initialise.php: done
* Move more zero-filled arrays from src/engine/BMGame.php to BMPlayer as in update_game_state_start_game(): done
* Strip reference to $playerIdx from do_next_step_choose_join_game(): done
* Move die_indices_without_reserve() to BMPlayer: done
* Add comment about subproperty array substrings: done
* Tidy up or at least add a comment on the player accessors where BMGame lies about the state: done except for `$this->get_buttonInfo()`, which i'll look into in a bit
* Make BMInterface::choose_button() take a player rather than a playerIdx: done
* Fix comment in does_need_random_button_selection(): done
* Fix whitespace in save_player_with_initiative: done
* Fix variable names in test test_do_next_step_specify_dice: done

Okay, so that's everything i asked for before.

cgolubi1:
I've kicked off the plan of collecting 3000 new random games, using an AWS instance.

cgolubi1:
Turns out i don't have any very recent games to replay, so i've also kicked off collecting those.  With any luck, i should either have 3000 of each or an interesting failure by tomorrow morning.

cgolubi1:
RandomAI collected 3500 games and is still chugging away, so i think that's a sufficiently good sign that i'm not going to easily turn up internal errors in the new codebase.

cgolubi1:
Creating games in all valid combinations of autoaccept or not, buttons specified or not, opponents specified or not:
* With requesting player autoaccept on:
    * With opponent autoaccept on:
        * both buttons specified
        * PB specified, OB random
        * PB random, OB specified
        * PB random, OB random
    * With opponent autoaccept off:
        * both buttons specified
        * PB specified, OB random
        * PB random, OB specified
        * PB random, OB random
    * With opponent not specified, picked up by an opponent with autoaccept on:
        * both buttons specified
        * PB specified, OB random
        * PB specified, OB not specified (and opponent specifies button)
        * PB specified, OB not specified (and opponent picks random)
        * PB random, OB specified
        * PB random, OB random
        * PB random, OB not specified (and opponent specifies button)
        * PB random, OB not specified (and opponent picks random)
    * With opponent not specified, picked up by an opponent with autoaccept off:
        * both buttons specified
        * PB specified, OB random
        * PB specified, OB not specified (and opponent specifies button)
        * PB specified, OB not specified (and opponent picks random)
        * PB random, OB specified
        * PB random, OB random
        * PB random, OB not specified (and opponent specifies button)
        * PB random, OB not specified (and opponent picks random)

Did all those, found no issues.  In the process, i tested the monitor with and without redirect to next game enabled.  I'll go ahead and repeat the whole set with c1's autoaccept turned off (which should make no difference at all) just to check.

cgolubi1:
Okay, all tests passed, i have run out of excuses not to merge this, let's see what happens.

PR #1890, opened 2016-01-04
Fix regression by always using the generic entry point to Forum.js

### LINKED ISSUE #1888: TypeError: state is null bug in Forum
See discussion in #1886 for details - the symptoms are something like this:
* Go to the Overview page
* Click on "(New post)" in the header - that takes me to http://localhost:8080/ui/forum.html#!threadId=1&postId=3793 and scrolls to the bottom of the post, as expected
* Click on "Mark thread as read" - that takes me to the top of the page of the Miscellaneous Chatting forum, as expected
* Click on "(New post)" in the header - that puts http://localhost:8080/ui/forum.html#!threadId=210&postId=3794 in my URL bar, and it takes me to the bottom of that post

And the underlying bug is "TypeError: state is null", i.e. (i'm pretty sure) something in that sequence of actions is causing Forum.js's state variable to be null at a time when something else expects it to not be null.

cgolubi1:
* Fixes #1888
* Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/420/

### DIFF
--- a/src/ui/js/Forum.js
+++ b/src/ui/js/Forum.js
@@ -57,7 +57,7 @@ Forum.SCROLL_ANIMATION_MILLISECONDS = 200;
 // These functions are part of the main action flow to load the page
 
 Forum.showLoggedInPage = function() {
-  $(window).bind('popstate', Forum.showPage);
+  $(window).bind('popstate', Forum.showLoggedInPage);
 
   var state = {
     'boardId': Env.getParameterByName('boardId'),


cgolubi1:
Manually verified that i can now click through new posts on each of Firefox, Safari, and Chrome, on my mac.

cgolubi1:
I'm going to wait to make sure Jenkins passes and give Jota a couple of hours to pop up and have an opinion about why this isn't an acceptable fix (in particular, if it's right, why wasn't it done this way in the first place?  If that was just a bug, no big deal, i'll merge it.  But what if i'm missing an intentional design choice?).

But if i don't hear from anyone by 22:00 EST that i shouldn't merge this, i'm going to merge it and load it on dev, so that we can start the clock on having dev testers look at it and hopefully do the prod release Tuesday or so.

AdmiralJota:
Unfortunately, I believe this does introduce a regression. Forum.showLoggedInPage() should only be called once, when Forum.html is initially loaded.

If you load the forum, then navigate around for a while between threads and boards, you can see in your browser's developer tools (in the network section) that the page makes one call to api/responder for each click. 

However, if you start using the "Back" and "Forward" buttons in your browser, you can see that the first time you do so it just makes one call to api/responder, but the second time it makes two calls, and the third time it makes four calls, and so on.

This is because changing Forum.showLoggedInPage() so that it binds the popstate event to Forum.showLoggedInPage() means that every time a popstate event occurs, it's binding it an additional time. When you use the Back and Forward buttons, the popstate event is triggered, which binds a second call to Forum.showLoggedInPage() to the event. The second time you click Back, both of those bindings are triggered, and each of them binds <i>another</i> call, so the page becomes geometrically slower as you navigate through your browser history.

PR #1887, opened 2016-01-03
Revert "pull changes from dev to production on 2016-01-03"

cgolubi1:
Reverts buttonmen-dev/buttonmen#1886

See discussion in #1886.

PR #1886, opened 2016-01-03
pull changes from dev to production on 2016-01-03

cgolubi1:
This update brings these pulls, already on dev, to production:
* New functionality:
    * #1762: add Insult die skill
    * #1702: add a page footer with copyright and help information
    * #1848: add forum markup to link to buttons, buttonsets, wiki pages, and github issues
* Bugfixes and improvements to existing functionality:
    * #1811, #1815, #1832: rework the overview UI to divide different game states (pending, active, completed, cancelled) more clearly
    * #1822, #1817, #1847: improve game display page for new and cancelled games
    * #1838, #1855: allow history searching on cancelled games
    * #1835, #1837: require confirmation for rejecting and cancelling games, other minor API fixes for reject/cancel
    * #1821, #1854: make page titles consistent, and put number of active games in overview page title
    * #1799, #1856: improve display of in-game skill help 
    * #1739, #1819: fix logic and wording bugs involving trip
    * #1717: fix logic bugs involving fire
    * #1862: fix UI bugs involving error codepaths
    * #1873: fix bug affecting chat editing at the beginning of the round
    * #1826, #1871: improve API coverage of the python API client, and add caching of completed games
* Refactoring and other invisible functionality:
    * #1793, #1809, #1833, #1869, #1874, #1875, #1876: code refactoring, documentation, modifications to unused code, and modifications to tests
    * #1828: system administration fixes

Dev pulls which collectively cover the same changes in this prod pull:
* #1800: 2015-11-09
* #1827: 2015-11-30
* #1836: 2015-12-04
* #1844: 2015-12-08
* #1847: 2015-12-10
* #1853: 2015-12-13
* #1857: 2015-12-15
* #1863: 2015-12-16
* #1872: 2015-12-18
* #1878: 2015-12-25
* #1881: 2015-12-31

Database updates: none

Once this pull is merged, i will load the changes on www.buttonweavers.com as part of #700.

Issue #700, opened 2014-03-12
push new features out to www.buttonweavers.com


cgolubi1:
Periodically update www.buttonweavers.com with new features which have been tested in dev.

Generic procedures for use in deployments:

Use a checked-out copy of buttonmen-master to find a sort order for database updates:

``` bash
cd deploy/database/updates/
UPDATES="<space-separated list of DB updates>"
git log --reverse --first-parent --oneline $UPDATES
UPDATES="<space-separated list of DB updates, now in the order reported by that output>"
```

Database update plan:

``` bash
# in the correct git working directory
git pull origin production
cd deploy/vagrant
vagrant provision

# on the sandbox or production VM
UPDATES="<space-separated ordered list of DB updates, as determined above>"


sudo -u mysql mkdir /tmp/pre /tmp/post
ls /tmp/pre /tmp/post
sudo -u mysql mysqldump -u root --skip-comments -T /tmp/pre buttonmen
cd /buttonmen/deploy/database/updates
for update in $UPDATES; do
  echo "APPLYING: $update"
  cat $update | mysql -u root buttonmen
done

sudo -u mysql mysqldump -u root --skip-comments -T /tmp/post buttonmen
```

How to diff database dumps:

``` bash
diff -r --brief /tmp/pre /tmp/post
```

PR #1895, opened 2016-01-05
Revert "Revert "pull changes from dev to production on 2016-01-03""

cgolubi1:
Reverts buttonmen-dev/buttonmen#1887

See discussion in #1886 

This is part of the chain of edits tracked in #700


cgolubi1:
So, basically, what i'm going to do is:
* bring up a test VM with the production branch as it appears right now
* revert the revert (in this pull)
* create a new pull to merge in the two changes since 2016-01-03
* load all of that on the test VM and look around
* push all of that to prod

PR #1899, opened 2016-01-06
If Forum.showPage() is invoked with no usable state information, regenerate it from the URL

### LINKED ISSUE #1888: TypeError: state is null bug in Forum

cgolubi1:
* Fixes #1888
* Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/421/

### DIFF
--- a/src/ui/js/Forum.js
+++ b/src/ui/js/Forum.js
@@ -57,7 +57,7 @@ Forum.SCROLL_ANIMATION_MILLISECONDS = 200;
 // These functions are part of the main action flow to load the page
 
 Forum.showLoggedInPage = function() {
-  $(window).bind('popstate', Forum.showLoggedInPage);
+  $(window).bind('popstate', Forum.showPage);
 
   var state = {
     'boardId': Env.getParameterByName('boardId'),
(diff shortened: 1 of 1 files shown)


cgolubi1:
This seems to work for me as well as my previous fix did.  Jota, is it what you had in mind?  If so, let's merge it.

AdmiralJota:
Yup, this looks exactly like what we talked about.

PR #1900, opened 2016-01-06
pull changes from master to dev on 2016-01-06

cgolubi1:
This update brings these approved pulls to the dev branch:
* User-visible:
    * #1899: fix forum navigation regression in a less performance-affecting way
* Ideally not user-visible: none

Database updates: none

Once this pull is merged, i will load the changes on dev.buttonweavers.com as part of #217.

PR #1901, opened 2016-01-06
pull changes from dev to production on 2016-01-06

cgolubi1:
This update brings these pulls, already on dev, to production:
* New functionality: none
* Bugfixes and improvements to existing functionality:
    * #1900: fix forum navigation regression in a less performance-affecting way
* Refactoring and other invisible functionality: none

Dev pulls which collectively cover the same changes in this prod pull:
* #1900: 2016-01-06

Database updates: none

This update is a follow-on to the prod release which initially started with #1886.


cgolubi1:
Just merging and loading, since this is still basically a one-liner.

PR #1907, opened 2016-01-07
Fix regression which entirely broke user verification page

cgolubi1:
What it says on the tin.  If you try to verify a user right now on dev or prod, you'll get a blank page.

Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/422/ (if it passes)

### DIFF
--- a/src/ui/js/Verify.js
+++ b/src/ui/js/Verify.js
@@ -50,10 +50,13 @@ Verify.getVerifyParams = function(callbackfunc) {
 };
 
 Verify.showStatePage = function() {
+  Verify.page = $('<div>');
 
   // If there is a message from a current or previous invocation of this
   // page, display it now
   Env.showStatusMessage();
+
+  Login.arrangePage(Verify.page);
 };
 
 Verify.setVerifyUserSuccessMessage = function(message) {


cgolubi1:
Also:
```
leocrotta,[~/src/git/cgolubi1/buttonmen-fix_verify_user_regression/src/ui/js],22:34(2)$ grep -L Login.arrangePage *.js
Api.js
Config.js
Env.js
Loader.js
leocrotta,[~/src/git/cgolubi1/buttonmen-fix_verify_user_regression/src/ui/js],22:34(0)$ 
```

None of those are pages, so that implies to first order this is the only thing which is exactly that 100% broken for exactly that reason.  I don't think i need to tell anyone who's been paying attention that this was caused by #1702

cgolubi1:
If Jenkins fails, i'll follow up on this, but i'm merging it now and starting to push it to dev.  Every player who gets a blank page on account verification is a player who's never coming back.

cgolubi1:
So, Jenkins failed, but i'm pretty sure it's because some useless test expects the wrong useless thing, not because the new code is broken.  Looking now.

blackshadowshade:
Can you fill me in on why this was broken by #1702? I don't exactly understand what's going on.

blackshadowshade:
Ah, I think I understand now. The Env message has nowhere to go because the page structure hasn't been set up (i.e., #c_body, and then Env.setupEnvStub() in Login.arrangePage).

blackshadowshade:
I'd suggest that we further streamline the function to remove the Env.showStatusMessage() call, since this is dealt with by Login.arrangePage():

    Verify.showStatePage = function() {
        Verify.page = $('<div>');
    
        Login.arrangePage(Verify.page)
    };

I've tested this locally on my MAMP stack, and the Env message still shows, as expected.

PR #1911, opened 2016-01-07
Strip out the Verify.showStatePage() test, since it was obviously useless

cgolubi1:
This is a followup to #1907 with the goal of making jenkins pass.  My assertion is that the previous test was useless --- it was testing the exact function that let us show a blank page without noticing it.  IMO this will all come out in the wash in #1858 (which is what i was working on when i found this problem in the first place, natch), so debugging this interim test is not a good use of time, and we should just make Jenkins happy until we can fix it for real.

* Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/423/ (assuming it passes)

### DIFF
--- a/test/src/ui/js/test_Verify.js
+++ b/test/src/ui/js/test_Verify.js
@@ -90,14 +90,7 @@ test("test_Verify.getVerifyParams", function(assert) {
 });
 
 test("test_Verify.showStatePage", function(assert) {
-  Env.setupEnvStub();
-  Env.message = {
-    'type': 'error',
-    'text': 'test error',
-  };
-  Verify.showStatePage();
-  var item = document.getElementById('env_message');
-  assert.ok(item.innerHTML.match('test error'), "env message is set by this function");
+// FIXME: put a test here that actually meaningfully tests the code
 });
 
 test("test_Verify.setVerifyUserSuccessMessage", function(assert) {


blackshadowshade:
I'm happy to strip out this test. My question is whether we need Verify.showStatePage() at all any more, since it's actually just Login.arrangePage($.'<div>')  (see my comment on #1907).

I'm happy to leave it there for readability purposes, but we could also just add a comment to the code, and thus lose the raw function and the test function directly.

cgolubi1:
I guess i'd file all of this under "optimize on your own time":
* We had a major regression which could have rapidly scared off potential new players.  So my first priority was to fix that as soon as i discovered it, using time i hadn't planned to spend on BM.
* My fix broke this test, basically because i didn't run the tests because i knew they hadn't done anything useful, so i assumed they were missing instead of wrong.  My second priority, which i'm still on the hook for, is to unbreak the build so people can make future changes.

IMO, optimization of this code is a distant third priority, and is not something i feel at all obligated to drop everything to do.  I can work on it this weekend, but i don't think the build should remain broken until this weekend.  So, if stripping the worthless test is unacceptable as a short-term fix to unbreak the build, i'd prefer you submit your own pull that fixes it in a way you find acceptable.

cgolubi1:
I opened an issue for fixing this function better, and i'm hoping that will be sufficient that you'll be willing to merge this pull.  Again, i think unbreaking the build is currently the top priority, so i hope you'll be willing to either accept the 3-second workaround or put the next BM time you have into unbreaking the build.

blackshadowshade:
I'll merge this based on Jenkins. When I have a bit of time online, I'll get rid of both Verify.showStatePage and the test for it, unless you have any objections (and it was for that reason that I made the comment that I did here, to find out if you really wanted to keep the function for purposes of readability).

PR #1914, opened 2016-01-07
Removed Verify.showStatePage

### LINKED ISSUE #1912: Fix Verify.showStatePage()
The function `Verify.showStatePage()` is not very useful in its current form, and is not tested.  Do something about this.  See comments in #1907 and #1911

blackshadowshade:
Fixes #1912.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1041/

### DIFF
--- a/src/ui/js/Verify.js
+++ b/src/ui/js/Verify.js
@@ -7,7 +7,7 @@ Verify.bodyDivId = 'verify_page';
 Verify.showLoggedInPage = function() {
   // Find the verification parameters and submit them to the server,
   // then display the result
-  Verify.getVerifyParams(Verify.showStatePage);
+  Verify.getVerifyParams(Login.arrangePage);
 };
 
 // Verify basically behaves roughly the same way regardless of whether or not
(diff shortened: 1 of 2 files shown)


cgolubi1:
Hmm, okay.  Can you call `Login.arrangePage()` without an argument?  I thought we had to define a div for this to work.  Let me try it.

cgolubi1:
It seems to work in Firefox.  I have to admit that the inconsistency makes me nervous --- every other page which calls `Login.arrangePage()` uses an argument except for a single "you shouldn't logically be able to reach this code" exception within Login.js itself.  And the `Login.arrangeBody()` code doesn't do anything in the logic to handle the `page` argument being undefined.

It definitely has the correct behavior in Firefox, of generating an empty div.  But i think:
* If you want to be able to call `Login.arrangePage()` with no `page` argument, you need to modify `Login.arrangeBody()` to wrap the use of the `page` argument with "if page"
* My druthers would be to just be consistent and use a stub function (and i do intend to add a reasonable test for the stub function, as part of #1858).

blackshadowshade:
In that case, should I just close the pull request then?

cgolubi1:
IMO let's close this out, since it's a small refactor and we don't have consensus on it.  Okay with you?

blackshadowshade:
Fine, closing this one then.

Issue #1394, opened 2014-11-06
Allow swing die changes after three draws in a row


irilyth:
If three rounds in a row end in a draw, both players should be allowed to change their swing dice. (I've helpfully forgotten if this was an official rule, or a house rule from the old site, or what, but I think it's a good rule and that we should do it, unless someone comes up with a compelling reason not to.)

PR #1920, opened 2016-01-13
Added logic to cause game to end in a tie after 10 consecutive drawn rounds

### LINKED ISSUE #1603: round number can only go up to 256, but the backend doesn't stop you from having more than 256 rounds
I just realised that RandomAI has been playing Dead Guy vs. Skomp for the past 3 days, and once it got up to 256 rounds, it appears that every round is number 256.  I'll include a grep of the round-ending lines below to demonstrate.

We could change the DB settings, but that only defers the problem for a game in an infinite loop.  In fact i don't think human players need more than 256 rounds --- the right answer is presumably for the backend to eventually say "you know what, you guys are done here" and end the game.

blackshadowshade:
Fixes #1603.

This will probably need an appropriate responder test.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1047/

### DIFF
--- a/src/engine/BMGame.php
+++ b/src/engine/BMGame.php
@@ -29,6 +29,7 @@
  * @property-read array $defenderAttackDieArray  Array of defender's dice used in attack
  * @property      array $auxiliaryDieDecisionArrayArray Array storing player decisions about auxiliary dice
  * @property-read int   $nRecentPasses           Number of consecutive passes
+ * @property-read int   $nRecentDraws            Number of consecutive draws
  * @property-read array $roundScoreArray         Current points score in this round
  * @property-read array $gameScoreArrayArray     Number of games W/L/D for all players
  * @property      int   $maxWins                 The game ends when a player has this many wins
@@ -197,6 +198,13 @@ class BMGame {
      * @var int
      */
     protected $nRecentPasses;
+
+    /**
+     * Number of consecutive draws
+     *
+     * @var int
+     */
+    protected $nRecentDraws;
 
     /**
      * Current points score in this round
@@ -1755,6 +1763,10 @@ class BMGame {
     }
 
     protected function do_next_step_end_round() {
+        if (!isset($this->nRecentDraws)) {
+            $this->nRecentDraws = 0;
+        }
+
         $roundScoreArray = $this->get__roundScoreArray();
         if (isset($this->forceRoundResult)) {
             foreach ($this->playerArray as $playerIdx => $player) {
@@ -1771,6 +1783,7 @@ class BMGame {
         }
 
         if ($isDraw) {
+            $this->nRecentDraws++;
             for ($playerIdx = 0; $playerIdx < $this->nPlayers; $playerIdx++) {
                 $this->gameScoreArrayArray[$playerIdx]['D']++;
                 // james: currently there is no code for three draws in a row
@@ -1784,6 +1797,7 @@ class BMGame {
                 )
             );
         } else {
+            $this->nRecentDraws = 0;
             if (isset($this->forceRoundResult)) {
                 $winnerIdx = array_search(TRUE, $this->forceRoundResult);
                 $forceRoundResult = $this->forceRoundResult;
@@ -1830,6 +1844,12 @@ class BMGame {
                 $this->gameState = BMGameState::END_GAME;
                 return;
             }
+        }
+
+        // 10 draws in a row cause the game to be declared as a tie
+        if ($this->nRecentDraws >= 10) {
+            $this->gameState = BMGameState::END_GAME;
+            return;
         }
     }
 
--- a/src/engine/BMInterface.php
+++ b/src/engine/BMInterface.php
@@ -191,6 +191,7 @@ class BMInterface {
                      '     n_players, '.
                      '     n_target_wins, '.
                      '     n_recent_passes, '.
+                     '     n_recent_draws, '.
                      '     creator_id, '.
                      '     start_time, '.
                      '     description, '.
@@ -200,6 +201,7 @@ class BMInterface {
                      '     :n_players, '.
                      '     :n_target_wins, '.
                      '     :n_recent_passes, '.
+                     '     :n_recent_draws, '.
                      '     :creator_id, '.
                      '     FROM_UNIXTIME(:start_time), '.
                      '     :description, '.
@@ -209,6 +211,7 @@ class BMInterface {
                                       ':n_players'     => count($playerIdArray),
                                       ':n_target_wins' => $maxWins,
                                       ':n_recent_passes' => 0,
+                                      ':n_recent_draws' => 0,
                                       ':creator_id'    => $playerIdArray[0],
                                       ':start_time' => time(),
                                       ':description' => $description,
@@ -695,6 +698,7 @@ class BMInterface {
         $game->maxWins   = $row['n_target_wins'];
         $game->turnNumberInRound = $row['turn_number_in_round'];
         $game->nRecentPasses = $row['n_recent_passes'];
+        $game->nRecentDraws = $row['n_recent_draws'];
         $game->description = $row['description'];
         if ($row['previous_game_id'] == NULL) {
             $game->previousGameId = NULL;
@@ -1137,8 +1141,8 @@ class BMInterface {
                  '    game_state = :game_state,'.
                  '    round_number = :round_number,'.
                  '    turn_number_in_round = :turn_number_in_round,'.
-        //:n_recent_draws
                  '    n_recent_passes = :n_recent_passes,'.
+                 '    n_recent_draws = :n_recent_draws,'.
                  '    current_player_id = :current_player_id '.
         //:last_winner_id
         //:tournament_id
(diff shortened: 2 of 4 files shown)


AdmiralJota:
How do the History results handle ties?

blackshadowshade:
Good point, Jota. The display is correct except for the colouring. I'll fix that now.

AdmiralJota:
What about the W/L/I totals at the bottom?

blackshadowshade:
Tied games don't count as any of W/L/I, so they don't affect the totals at all.

I checked the colouring, and that is actually also correct. The only thing that is currently incorrect is the application of the filter "Round Score", which needs to be tweaked to count completed games with no winner as tied.

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1048/ (if it passes)

Screenshots to follow (to allow discussion about whether we should also be counting ties in the totals summary).

blackshadowshade:
Screenshot 1: no filter for Round Score
screen1

Screenshot 2: player A winning
screen2

Screenshot 3: player B winning
screen3

Screenshot 4: tie
screen4


I think that it might be desirable to have a W/L/T/I summary, instead of the current W/L/I. What do you all think?

dwvanstone:
What does "I" stand for? Incomplete? I'm very confused and I guess I never noticed the summary below.

Why doesn't that column simply have W/L/T for sums of Win/Lose/Ties? If "I" is incomplete that belongs in the Game Status summary. 

Or is "I" inconclusive, i.e., kind of like a stalemate in chess? (10 consecutive draws?)

blackshadowshade:
The 'I' stands for incomplete, according to the mouseover text. Take a look at the history page yourself, mouse over, then tell us what you reckon.  :)

dwvanstone:
Thanks - I didn't know there was mouseover text there. (The die skills on a game page have a blue circle containing an "i" to indicate there's some mouseover text, so that's the tip-off I was looking for.)

It still seems to me that the "I" column belongs with Game Status, but I can open that up as a separate bug.

What's confusing about this total is that it doesn't actually say who won more games. 

For instance, if I care about matches between me and a fictitious player "Harold", it's possible that I've won 10 games with a record 3/2/0 and lost 5 games with a record 0/3/0 and never had a draw.

So my stats against Harold, in my mind, should be W/L/D = 10/5/0. But here the stats would be W/L/T = 30/35/0, which doesn't seem right. I can open this up as a separate bug too (if it's not there already).

I know I'm asking for this total to be all things to all people here, so I'd say the simplest is to just add up all of the values. I'd suggest simply "W/L/T" on the bottom (simply summing the values), and it's fine if you want to add incomplete to be consistent with what we had ("W/L/T/I").

By the way, I really like how flexible the history page is overall. It's really, really, really nifty. There's a lot of flexibility in a small amount of space. Well done, folks!

blackshadowshade:
I'm waiting for @AdmiralJota to join the discussion, since the History page was his baby, from memory. I'm sure that we'll have a couple of issues arising from this discussion. What I'm looking for at the moment, though, is an agreement on an acceptable endpoint for this pull request.

dwvanstone:
> What I'm looking for at the moment, though, is an agreement on an acceptable endpoint for this pull request.

I figured as such -- that's why I encourage your suggestion of "W/L/T/I". That way for those people (like me) who didn't notice the "I" they at least will find the "T" column.

irilyth:
I don't think we should merge this until we fix #1394, because otherwise you could get a situation where you get ten draws in a row not because the game is deadlocked, but just due to unlucky swing die choices.

dwvanstone:
I'm sorry for all of my misunderstanding today. I found this very confusing and I was wrong in my post earlier. (I'll correct it.)

Rounds can end in a win, loss, or a tie. 
Completed games can end in a win, loss, or tie [new in this fix].

I can't wrap my head around a game having a "tie". That just seems like the wrong word. I strongly suggest think it should be "draw". (One player may be winning when there are 10 consecutive tied rounds.)

irilyth:
There was some more talk about this on #1603, the ticket that this pull is intended to fix. One particular thing is that I don't think the "one player is winning when there are ten consecutive tied rounds" situation should ever arise -- and if it somehow does, I agree that this is more like a stalemate than like a draw.

I think #1603 was only about a situation where a game *can't* end, and "N consecutive tied rounds" was a way to identify games that can never end.

And, I still think (as I said in #1603) that if it's human players playing the game, there should be some way for them to terminate the game whenever they feel like it, and that having the backend try to guess whether they want the game to end is fundamentally the wrong idea. Human players *never* need the backend to tell them "ok, you've played enough, time for this game to be over".

irilyth:
In fact, I like this suggestion from #1603:

> 
I'm not actually sure what should happen in a game that neither player can win. Do we even have the idea of a game that ends in a draw, at this point? The old site had a way to delete "stale" games, such that they didn't count as a win or a loss for either player, and I could imagine using that here, letting the players delete a game if they both agree. Possibly related to the often-desired functionality of letting a player decline to start a game they're challenged to in the first place.

If a game turns out to be unwinnable, how about marking it as canceled, exactly as if it had been withdrawn or rejected?

dwvanstone:
Just so I can do this correctly in the future, is it more helpful to have discussion here in the pull request or in the bug itself?

AdmiralJota:
> So my stats against Harold, in my mind, should be W/L/D = 10/5/0. But here the stats would be W/L/T = 30/35/0, which doesn't seem right. I can open this up as a separate bug too (if it's not there already).

Could you give an example of a search you've done where it behaved that way? When I just did a history search, the results looked normal to me:

bm1

In any case, to be honest, I think I'd prefer it if a BM game never ended in a tie/draw. It would keep one of my personal assumptions intact. I'd be more in favor of either cancelling a game that was determined to be unwinnable or else arbitrarily choosing a winner (preferably the latter). If there have been any rounds that weren't tied, pick the player who won more of them. Otherwise, coin flip. We already do something similar for choosing initiative, right?

dwvanstone:
> So my stats against Harold, in my mind, should be W/L/D = 10/5/0. But here the stats would be W/L/T = 30/35/0, which doesn't seem right. I can open this up as a separate bug too (if it's not there already).

I take back this case. I made an assumption as to what the statistics on the bottom were, and I was wrong.

blackshadowshade:
I disagree with @irilyth that this pull request should be dependent on #1394.

However, if it is decided that #1394 needs to be addressed first, then I am pretty sure that the infrastructure that I introduce in this pull request will be needed for #1394 also, which means that I'll eventually be making this pull request more complicated, and offering a solution for both #1920 and #1394.

irilyth:
> I disagree with @irilyth that this pull request should be dependent on #1394.

Why?

blackshadowshade:
I disagree because I'm pretty sure that we have never hit the case with 10 consecutive ties yet. Plus, all current games will be defaulting to 0 consecutive ties, so players would need to have an *extra* 10 consecutive ties before the tied game state occurs.

Basically, the pull request shouldn't actually change anything for *any* real player for quite a while. Holding up the addition of the infrastructure (and complicating the pull request) because of an extremely rare occurrence seems to me to be a rather excessive reaction.

blackshadowshade:
I can't actually check my assertion about 10 ties directly on the database, since my current IP address is not whitelisted, so it's possible that I'm wrong.

irilyth:
What infrastructure are we trying to add here? I thought #1603 was a fairly straightforward problem where the number of rounds could go on longer than the number of rounds that the infrastructure supported, and that it'd be a trivial fix to say "oops, database is out of rounds, game over".

Arguing about whether games should conceptually ever end in a tie or not, weird cases where players might want games to go on for lots of rounds or not, etc, all seems like a distraction from that.

blackshadowshade:
I see #1603 as a symptom of a bigger problem, rather than a problem that needs to be solved directly. The bigger problem here is that degenerate games have no meaningful endpoint currently.

Even taking into account the suggestion that we should be supporting games with 256 rounds, what should then happen when the game tries to go to round 257? Should the game then end as a tie?

We're not really avoiding the main philosophical questions by saying that we should simply ending games after 256 rounds, because the History and Overview pages would still theoretically need to know how to deal with them.

The infrastructure that needs to be added is *in the case* that a game ever ends without an obvious winner.

cgolubi1:
So, i actually care about the bug in #1603.  I didn't file it as a symptom of a larger problem, i filed it because it makes my testing utility get stuck, and more to the point because i feel from a technical debt POV that our codebase should be consistent with itself --- we shouldn't let a game get into a state in which it's claiming to advance rounds but isn't actually, that's a bug.  Obviously, the bug-filer's intent doesn't trump everything or anything, but i do want to get that out there.

I don't think we need to limit the number of rounds human players can play.  Human players can surrender if they care.  I'd be perfectly happy with a solution in which:
* We made the limit 256 rounds, and the backend and the DB agreed on that limit
* After 256 rounds, the game wound up in the existing cancelled state

cgolubi1:
I should also note: i was going to let y'all argue this one out, but in fact i really dislike the idea of adding a lot of logic and complexity (tracking of recent draws, a new game state, new UI elements on the history page, etc) to support a use case that we expect human players to really truly seriously almost never hit.  So when i'm able to test this, my testing is going to reflect the fact that i think adding that complexity is a bad idea.

blackshadowshade:
For the record, there's no new game state.  :)

As for tracking of recent draws, we're going to need that for #1394, so that complexity will be coming, whether or not we introduce it to solve this issue.

blackshadowshade:
Since it appears that the approach in this pull request isn't going to be acceptable, I'm closing this pull request.

PR #1921, opened 2016-01-13
Fixed description of option die with one side

### LINKED ISSUE #1882: singular/plural bug with option dice
When bobby5150's `(1/30)` die is set to have 1 side, the API shares the description as `Option Die (with 1 sides)` - it should be "1 side".  I assume this is a bug with option dice.

blackshadowshade:
Fixes #1882.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1046/


cgolubi1:
Not much to talk about here.  Merging, and thanks for the fix.

PR #1923, opened 2016-01-15
Changed error message for server connection issues

### LINKED ISSUE #1315: show a different UI error when a POST gets no response from the server at all
Right now, Api.apiParsePost() and Api.apiFormPost() have a generic fail() block that just reports an internal error calling the function.

However, if the POST doesn't get a response from the server at all, it's more likely that the player has a local network issue (or the server is down) than that anything is getting an internal error.  Detect when the error is that no response at all was received from the server, and show the player a different error, suggesting a possible network problem, in that case.

blackshadowshade:
Fixes #1315.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1050/

### DIFF
--- a/src/ui/js/Api.js
+++ b/src/ui/js/Api.js
@@ -220,7 +220,6 @@ var Api = (function () {
 
   ////////////////////////////////////////////////////////////////////////
   // Load and parse a list of buttons
-
   my.getButtonData = function(buttonName, callbackfunc) {
     my.apiParsePost(
       {
--- a/src/ui/js/Api.js
+++ b/src/ui/js/Api.js
@@ -105,14 +105,14 @@ var Api = (function () {
       }
     ).fail(
       function(XMLHttpRequest) {
-        // when the client fails to connect to the server, then
-        // the request finishes (readyState = 4), but the response
-        // is 404: Page not found
-        if ((404 == XMLHttpRequest.status) &&
-            (4 == XMLHttpRequest.readyState)) {
+        // when the client fails to connect to the server at all, then
+        // the request is not initialised (readyState = 0) and there is
+        // no response
+        if ((0 === XMLHttpRequest.status) &&
+            (0 === XMLHttpRequest.readyState)) {
           Env.message = {
             'type': 'error',
-            'text': 'Error when attempting to connect to server'
+            'text': 'Could not connect to Button Men server—are you online?'
           };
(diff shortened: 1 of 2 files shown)


src/ui/js/Api.js
```diff
@@ -104,11 +104,22 @@ var Api = (function () {
...
-        };
+      function(XMLHttpRequest) {
+        // when the client fails to connect to the server, then
+        // the request finishes (readyState = 4), but the response
+        // is 404: Page not found
+        if ((404 == XMLHttpRequest.status) &&
+            (4 == XMLHttpRequest.readyState)) {
+          Env.message = {
+            'type': 'error',
+            'text': 'Error when attempting to connect to server'
```
cgolubi1: I would make this more explicit, something like --- "Could not connect to Button Men server -- are you online?"  (or maybe "reach" or "find" or "talk to" instead of "connect to").

cgolubi1:
Anyway, other than quibbles about wording, this seems like it should work.  To test it, i'll bring up a site and try both breaking the ability to talk to the server, and doing something to trigger a legitimate internal error, to make sure both cases still look right to me.

cgolubi1:
Hmm, actually: now that i think about this, you triggered a 404 by breaking the URL of the backend.  But in that case, you're reaching the buttonmen server and it's giving you a 404 because you broke the responder URL.  That's not actually what's going on in the case in which the network connection is *missing*.  In that case, there won't be a 404.

Let me see if i can reproduce that case easily on a local server, 1s.

cgolubi1:
Okay, here we go --- i think the right test is:
* Create a game and load it so it's your turn
* Shut down apache (or whatever your webserver is)
* Click "Beat People UP!"

That should reproduce the actual bug, and when i do that, i get "Internal error while calling loadGameData" with or without your fix.

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1055/ (if it passes)

cgolubi1:
This seems promising.  Let me check it out.

cgolubi1:
Okay, this seems good.  The fix is pretty targeted towards that particular test i suggested, and we may find that that test isn't a good match for what is actually happening when people's network connections flake.  But it's my best guess, and this works in that case.  So let's try it.

PR #1929, opened 2016-01-18
Hid swingRequestArray for game states later than SPECIFY_DICE

### LINKED ISSUE #1289: Make the API-visible form of swingRequestArray only non-NULL at swing setting time
This arises from the comments on #1073:

"... the dual interpretation of swingRequestArray:
1. as a mirror of the internal swingRequestArrayArray
2. as a publicly available component of the API, used during swing setting

Since swingRequestArray is currently used only by the UI at swing setting time, it really should be empty at other times. Thus, it ceases to be a mirror of swingRequestArrayArray.

Renaming one or both will help to separate the API response from the internal representation."

blackshadowshade:
Addresses part of #1289.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1053/

### DIFF
--- a/src/engine/BMGame.php
+++ b/src/engine/BMGame.php
@@ -3706,6 +3706,11 @@ class BMGame {
      * @return array
      */
     protected function get_swingRequestArray($playerIdx, $requestingPlayerIdx) {
+        // do not show swing requests unless this information is actually necessary
+        if ($this->gameState > BMGameState::SPECIFY_DICE) {
+            return array();
+        }
+
         $swingRequestArray = array();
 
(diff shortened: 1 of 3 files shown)


cgolubi1:
There's not much to do for this except to replay some games and look at the regressions by hand to see if they make sense.  It will probably be time-prohibitive to test too many, so i'll just have to pick some that seem representative and hope for the best.

cgolubi1:
I looked at a few, and they seem okay.  I'm going to go ahead and merge this, since i don't have a better idea of how to test.

Issue #1927, opened 2016-01-17
Mood Option bug


irilyth:
In http://dev.buttonweavers.com/ui/game.html?game=2117, I made 15 consecutive attacks with a Mood Option die, and it didn't change size once. Seems likely to be a bug.

dwvanstone:
I believe Mood dice only work with Swing dice, not with Option dice. I'd say this is a bug with the Connecticut button. (The States, in general, have dice that may have worked as intended on the DHS site but won't work on the current site, such as shadow stealth on the Illinois button.)

blackshadowshade:
This is not a bug, as far as I can tell. "Mood" is actually "Mood Swing", which doesn't take effect with non-swing dice.

irilyth:
I suppose it is technically "Mood Swing", and Option isn't the same as Swing. I think we should support this, but if no else thinks we should, I'm not writin' code so I guess we're not. :^ )

Should we leave the skill on the relevant die on Connecticut? I feel like it's ok for buttons to have useless skills (e.g. GorgorBey); we might want to include an interaction note that Mood doesn't effect variable dice other than Swing dice, in that case.

cgolubi1:
Hmm.  IMO, it's correct for Mood to affect option dice, because i know that on DHS *Turbo* affected option dice (e.g. Alice (`(6) (8) (8) (10) (1/30)!`) or Notorious (`G(2/12)! g(4/16)! G(8/20)! g(10/24)! rH(7) rH(11) rh(13) rh(17)`) or theMachine (`dp(1) B(1/30)! tD(4) q(7,11) rcz(20)`), and IMO Mood and Turbo should operate on the same set of dice as each other.

Shadowshade: is your claim that our implementation is different from the DHS one?  Or that Turbo and Mood should behave differently from each other?  (There were no Mood Option dice on DHS.)  Or something else?

Regardless: i don't think it's okay to release a button with a Mood Option die that doesn't Mood --- it's just going to be too confusing to players.  If we don't want to support Mood Option, we should outlaw it, and change Connecticut's recipe.

AdmiralJota:
I vote for A) supporting Mood Option and B) disabling Connecticut temporarily until we do support it.

blackshadowshade:
My comment was based on the fact that Mood Swing and Turbo Swing are officially sanctioned skills on the Cheapass website (http://www.cheapass.com/bpu/butnrulz.html), while Turbo Option was a DHS extension of Turbo.

[...]

PR #1931, opened 2016-01-22
Disabled Connecticut so that mood option doesn't get misinterpreted

blackshadowshade:
Addresses part of #1927.

Requires database update 01927_mood_option_01.sql.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1054/


cgolubi1:
Okay, i just need to do a database update-vs-reload test, and then this will be good to go.

PR #1916, opened 2016-01-10
pull changes from master to dev on 2016-01-10

cgolubi1:
This update brings these approved pulls to the dev branch:
* User-visible:
    * #1885: fix description of weak die skill
    * #1898: make forum text area larger
    * #1861: activate most of the "50 States" buttons
* Ideally not user-visible:
    * #1911: patch around a previous test regression

Database updates:
```
activate_50_states_01.sql
```

Once this pull is merged, i will load the changes on dev.buttonweavers.com as part of #1915.

Special actions: apply this patch to the codebase for the dev site so people can test random 50 States buttons:
```diff
$ cat ~/50_states.patch 
diff --git a/src/engine/BMInterface.php b/src/engine/BMInterface.php
index 7bef6f5..0e33d53 100644
--- a/src/engine/BMInterface.php
+++ b/src/engine/BMInterface.php
@@ -1067,7 +1067,7 @@ class BMInterface {
             }

             if (empty($allButtonData)) {
-                $allButtonData = $this->get_button_data(NULL, NULL, TRUE);
+                $allButtonData = $this->get_button_data(NULL, '50 States', TRUE);
                 $nButtons = count($allButtonData);
             }
```

PR #1932, opened 2016-01-22
pull changes from master to dev on 2016-01-22

cgolubi1:
* User-visible:
    * #1921: fix description of option dice with one side
* Ideally not user-visible:
    * #1879: refactor of part of BMGame into BMPlayer

Database updates: none

Once this pull is merged, i will load the changes on dev.buttonweavers.com as part of #1915.

Special actions: keep the patch active from #1916 so people can keep testing 50 states buttons.


cgolubi1:
No database change, so i'm just going to merge this and load it on dev.

PR #1933, opened 2016-01-22
Add a few missing skills and interactions to responderTest's hardcoded list

cgolubi1:
No effect on existing tests, but these skill definitions are correct, and not having them sometimes causes false positives in regression tests i run with RandomAI.

Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/426/


blackshadowshade:
Looks fine to me, and Jenkins passes, so I'll accept this.

PR #1938, opened 2016-01-30
Added PHPDoc comments to BMGame and BMPlayer

blackshadowshade:
Continues to address part of #1130. The rest of the required PHPDoc comments are in BMInterface*, ApiSpec, or ApiResponder.

I've also taken the opportunity to make the PHPDoc types consistent, using bool and int instead of boolean and integer.

There is one small code change, namely the change from BMGame->get_waitingOnActionArray(...) to BMGame->isWaitingOnAction(...), since I realised that the return value is a boolean, not an array.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1065/ (if it passes)

### DIFF
--- a/src/api/ApiResponder.php
+++ b/src/api/ApiResponder.php
@@ -33,7 +33,7 @@ class ApiResponder {
      *   don't start a session
      *
      * @param ApiSpec $spec
-     * @param boolean $isTest
+     * @param bool $isTest
      */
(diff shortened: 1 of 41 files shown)


blackshadowshade:
So, I've added PHPDoc comments to a number of BMInterface* classes too.

cgolubi1:
Seems fine.  Merging.

PR #1942, opened 2016-02-06
BMInterfaceGameAction refactor, second try

blackshadowshade:
Continues to address part of #974.

This is meant to be a pure refactor, with no change in user experience at all.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1070/


blackshadowshade:
Deconflicted.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1075/ (if it passes)

cgolubi1:
Hmm, can you merge buttonmen-dev into this?  I'm having trouble finding a good way to see your diff, which github won't show me because it's too big.  The `git diff --name-only origin/master bss/$branch` trick is showing a lot of unrelated recent changes, presumably because origin/master has them, but your branch doesn't.

cgolubi1:
So... first high-level question: why is this `gameAction`?  Why isn't it `game`, like `player` or `forum` that you created earlier?  Is it `gameAction` because there's going to be `gameSomethingElse` later, or what?

blackshadowshade:
So, I'm following Jota's initial separation, as defined in #974.

cgolubi1:
Okay, shadowshade and i discussed this, and my current understanding is that the refactor work left to do (which is inspired by, but not identical to, Jota's separation defined in #974), is:
* all game interface changes made in response to player API calls go into `BMInterfaceGameAction`
* all functions which load/save game-related state from the DB go into `BMInterfaceGameDatabase` (or something)

This pull is the first step of that.

cgolubi1:
Okay, so step 1 is trying to figure out what's actually in this diff.  I verified that everything git views as "added" to BMInterface.php is simply something that was moved around, and that those functions (`find_editable_chat_timestamp()` and `log_game_chat()`, for my reference) are indeed unmodified.  The horizontal whitespace on the function comment for `find_editable_chat_timestamp()` got munged so it doesn't line up with the function now, so if there are other fixes, might as well make that line up, but obviously i don't care in isolation.

cgolubi1:
These are what i see as the steps of testing:
* ~~Understand the goal of this refactor, so i can look at it against the goal~~
* Make sure all code is only moved, not edited, and if any code is edited in addition to being moved i understand why.
* Make sure each function that was moved seems like it more-or-less belongs in the new BMInterfaceGameAction
* Replay old games
* Play new games

cgolubi1:
Here are the function names and permissions which are now in BMInterfaceGameAction:
```
protected add_player_to_new_game
protected chat_is_insertable
protected db_delete_chat
protected db_insert_chat
protected db_update_chat
protected insert_new_game
protected is_action_current
protected retrieve_button_ids
protected retrieve_player_autoaccept
protected set_option_values
protected set_random_button_flags
protected set_swing_values
protected validate_game_info
protected validate_previous_game_players
public adjust_fire
public create_game
public dismiss_game
public join_open_game
public react_to_auxiliary
public react_to_initiative
public react_to_reserve
public save_join_game_decision
public select_button
public submit_chat
public submit_die_values
public submit_turn
```

Here are the function names and permissions which stayed in BMInterface:
```
protected assemble_button_data
protected build_game_log_query_restrictions
protected choose_button
protected clear_option_values_from_database
protected clear_swing_values_from_database
protected count_noun
protected db_insert_die
protected delete_dice_marked_as_deleted
protected determine_game_colors
protected does_need_random_button_selection
protected execute_button_data_query
protected find_editable_chat_timestamp
protected get_all_games
protected get_button_id_from_name
protected get_button_recipe_from_name
protected get_button_tags
protected get_buttonset_id_from_name
protected get_config
protected get_currentPlayerId
protected get_friendly_time_span
protected get_game_status
protected get_player_name_from_id
protected get_player_name_mapping
protected load_button
protected load_die_attributes
protected load_game
protected load_game_action_log
protected load_game_attributes
protected load_game_chat_log
protected load_game_parameters
protected load_hasPlayerDismissedGame
protected load_lastActionTime
protected load_message_from_game_actions
protected load_option_values_from_last_round
protected load_option_values_from_this_round
protected load_player_attributes
protected load_player_colors
protected load_swing_values_from_last_round
protected load_swing_values_from_this_round
protected log_game_actions
protected log_game_chat
protected mark_existing_dice_as_deleted
protected read_game_list_from_db_results
protected recreate_optRequestArrayArray
protected regenerate_essential_die_flags
protected resolve_random_button_selection
protected save_action_log
protected save_active_dice
protected save_basic_game_parameters
protected save_button_recipes
protected save_captured_dice
protected save_chat_log
protected save_dice
protected save_game
protected save_option_values_from_last_round
protected save_option_values_from_this_round
protected save_out_of_play_dice
protected save_player_game_decisions
protected save_player_with_initiative
protected save_players_awaiting_action
protected save_random_button_choice
protected save_round_scores
protected save_swing_values_from_last_round
protected save_swing_values_from_this_round
protected set_logEntryLimit
protected set_message
protected set_option_max
protected set_swing_max
protected set_twin_max
protected validate_and_set_homepage
protected validate_url
public __construct
public __get
public __set
public cast
public count_pending_games
public forum
public gameAction
public get_active_players
public get_all_active_games
public get_all_cancelled_games
public get_all_completed_games
public get_all_new_games
public get_all_open_games
public get_button_data
public get_button_set_data
public get_next_pending_game
public get_player_id_from_name
public get_player_names_like
public history
public load_api_game_data
public player
```

cgolubi1:
So, a few things here don't make sense to me:

I would have expected BMInterfaceGameAction to contain the public functions which relate to games, per our previous discussion about the split.  Given that:
* Why wasn't `load_api_game_data()` moved?  That's a public function which relates to games.
* Why were all these various protected functions moved?

The choice of which chat-related functions to move seems super arbitrary to me.  You moved:
```
protected chat_is_insertable
protected db_delete_chat
protected db_insert_chat
protected db_update_chat
```
and didn't move:
```
protected find_editable_chat_timestamp
protected load_game_chat_log
protected log_game_chat
protected save_chat_log
```
That seems weird to me because:
* `chat_is_insertable` and `find_editable_chat_timestamp` seem very similar in what type of function they are (data-backed logic checks which return information to the player about whether it's valid to mutate chat based on the game state).  It's weird to have them in different files.
* The `db_*` functions you moved literally do database queries.  So why should they be here rather than in the BMInterfaceGameDatabase class you're going to create?

Tell more more about why you chose to move the functions you did and leave the functions you did.  The downside of the class casting trick is that (as i understand it) it makes it easy to make a lot of calls back and forth between BMInterface and BMInterface* functions without realising you're doing so.  I *think* with this split you're going to wind up making a lot of calls back and forth between BMInterfaceGameAction and BMInterfaceGameDatabase in both directions.  That seems wrong.

Is this split more coherent than it seems at a glance?  If so, tell me more about how.

cgolubi1:
Interesting, okay - i realised that of course i can sanity-check the includes by looking to see if any function moved to BMInterfaceGameAction is called from BMInterface.  And i was right about one path --- `db_insert_chat()` is called from BMInterface.  That seems to be the only one.

blackshadowshade:
So, I've taken a look at this again, since I don't remember exactly what I was doing way back when I did the refactor. I'm pretty sure my main motivation for moving the protected functions was to reduce the need for casting within both BMInterface and BMInterfaceGameAction.

For your specific example, I went back to the commit called "Moved submit_chat() from BMInterface to BMInterfaceGameAction" and took a look. Here's what I'm seeing:

The function submit_chat() calls:
- load_game_chat_log(), also needed for generic game loading
- load_game_action_log(), also needed for generic game loading
- find_editable_chat_timestamp(), also needed for generic game loading
- db_update_chat(), only called by submit_chat()
- db_delete_chat(), only called by submit_chat()
- chat_is_insertable(), only called by submit_chat()
- db_insert_chat(), also needed during generic game saving

Thus, three of the functions can be shifted without a second thought, while the last one is borderline. Based on the naming, I shifted it along with the other two db_*_chat functions.

Pretty much all of the decisions about shifting protected functions were similar to this. If they were exclusively required in BMInterfaceGameAction, then I shifted them without a second thought. If they were borderline, it looks like I left them in BMInterface, except for db_insert_chat().

cgolubi1:
Hmm, okay, thanks for the context.

I guess i think we need an understanding of what the *intention* of the difference between BMInterfaceGameAction and the future BMInterfaceGameDatabase (or whatever) will be.  If there's not a clear difference between the two, then why not just make BMInterfaceGame and move all game-related stuff into it?

At any rate, my *assumption* is that we'll wind up with some possible code path trees:
* BMInterface -> BMInterfaceForum
* BMInterface -> BMInterfaceHistory
* BMInterface -> BMInterfacePlayer
* BMInterface -> BMInterfaceGameAction -> BMInterfaceGameDatabase
and that while it's fine for a function to "call right" (a function in a class to the left of an arrow can call a function in a class to the right of the arrow), it doesn't make sense for a function to "call left".

Given that, you can't move db_insert_chat into BMInterfaceGameAction if log_game_chat (which calls it) is going to wind up in BMInterfaceGameDatabase - because that would be a "call left".

So i guess i'm looking for one of two options:
A. Come up with (and write down) a definition of BMInterfaceGameAction and BMInterfaceGameX (i've been assuming "Database", but if something else makes more sense, that's fine --- you haven't implemented it yet, so don't feel tied to that name) which we can use to clearly figure out what functions should be in what class.
B. Decide there is no rigorous distinction with the code as it currently is, so we only need BMInterfaceGame (in which case i'm tentatively okay with renaming BMInterfaceGameAction to BMInterfaceGame and approving this intermediate pull, and you can move the rest of everything from BMInterface to BMInterfaceGame in a future pull).

blackshadowshade:
@AdmiralJota, do you have any thoughts on this, considering that you did the initial work dividing up BMInterface into logical blocks?

blackshadowshade:
On my part, I don't mind actually consolidating the original suggestion a little more, so that we don't end up with a class BMInterfaceGameDisplay that only has four functions. However, I see some value in separating out game-related user actions from everything else that happens behind the scenes.

If we decide to put everything into a big BMInterfaceGame class, I'm also fine with that. I don't feel strongly about the final choice.

cgolubi1:
Yeah, so i may be being prejudiced by my vague high-level sense of what's going on in the codebase here.  But that vague high-level sense is that the game functions that exist in BMInterface (before this refactor) break down into:
* Player-visible actions (public functions)
* Functions that actually read or write some game state from the database
* "Game state change" logic functions that really ought to be refactored into BMGame
* Stuff that deals with action logs or chat (maybe some of which should be refactored into BMGameAction, not sure)

Anyway, the point is: my bias is that the problem with game stuff in BMInterface is that a lot of it doesn't belong in BMInterface at all.  And i think if we moved that stuff into BMGame, we'd find that the combination of the first two items (player visible functions and database read-write functions) is not actually too big to fit in a single BMInterfaceGame class.

And conceptually it makes sense *to me* that a BMInterfaceGame class, representing the interface to a game, would contain both the public functions and the protected helpers.  Conversely, if we had a really clean setup where everything in BMInterfaceGame* was either a public function or a database reading function, it would be reasonable either to put both types of things in BMInterfaceGame or to have a BMInterfaceGame and a BMInterfaceGameDatabase.

I guess what i dislike is having multiple classes where it *isn't* obvious what class a particular function should live in.  So that says to me, what i think i said before, which is that if we can't articulate a clean definition of what the BMInterfaceGame* subclasses should be and what should live in them, then we should just make a unitary BMInterfaceGame and call that good enough for now.

AdmiralJota:
>@AdmiralJota, do you have any thoughts on this, considering that you did the initial work dividing up BMInterface into logical blocks?

To be honest, my recollections from back then are a bit vague. But if @cgolubi1's premises are true (that a lot of the BMInterfaceGame methods would fit better in BMGame and that what remains would be small enough to make up a single BMInterfaceGame class), then that seems like a reasonable approach to me.

That does raise a follow-up question: if we move all of that code into BMGame, how big will that make BMGame? Will we want to refactor it next?

blackshadowshade:
I don't understand which parts of BMInterface belong in BMGame. To my mind, BMGame is well defined.

blackshadowshade:
Okay, after chatting with Chaos, the easiest way forward at the moment is to refactor most of BMInterface into a catch-all BMInterfaceGame class. We'll consider the current pull request to be a partial attempt at this, to be completed later.

blackshadowshade:
Right, I think I've done the name change. We'll see what Jenkins thinks.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1123/ (if it passes)

cgolubi1:
New steps of testing:
* Make sure all code is only moved, not edited, and if any code is edited in addition to being moved i understand why
* Make sure the set of moved functions looks like it should continue to work, and devise a test for any specific thing i'm not sure will still work
* Make sure the set of moved functions contains only things which belong in BMInterfaceGame (note: some things which belong in BMInterfaceGame may not yet have been moved, and that's okay for this pull)
* Replay old games
* Play new games

cgolubi1:
Okay, so the only code differences i see are a couple of additional casts, in `add_player_to_new_game()` and `retrieve_player_autoaccept()`.

cgolubi1:
After discussion with shadowshade, i opened #1987 for the problem introduced by this partial refactor.  I've finished replay and new game testing, and did a quick smoke test to make sure that editing of game chat (which uses functions which have been made messier by this refactor) still works.  Merging.

PR #1943, opened 2016-02-06
Added more RandomBM* buttons, and added explicit sort order for buttons

### LINKED ISSUE #779: Play with a randomly-generated button

blackshadowshade:
Fixes #779.

Requires database update 00779_RandomBM_06.sql

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1074/


cgolubi1:
Sort orders should be multiples of 10 the way the buttonset sort orders are, so that we have room between buttons for insertions.  Basically there's no reason not to, and it might simplify a change later.

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1079/ (if it passes)

cgolubi1:
Okay, so, test plan here:
* The usual database update-vs-rebuild test
* Look for regressions caused by the sort order business - just look for and manually go through frontend calls which return lists of buttons, and look out for internal errors or other problems (and also make sure the sort order actually works in the create game UI)
* i *think* this is enough of a noop for RandomBMMixed that i can use replay testing.  If i can't, figure out why not and make sure it's okay
* ditto RandomBMFixed
* try out a bunch of automated games with the new RandomBM types and look for problems, especially look for things that blow up with RandomBMPentaskill

src/engine/BMBtnSkillRandomBMPentaskill.php
```diff
@@ -0,0 +1,61 @@
+<?php
+/**
+ * BMBtnSkillRandomBMPentaskill: Code specific to RandomBMPentaskill
+ *
+ * @author: james
+ */
+
+/**
+ * This class currently supports the special skills of RandomBMPentaskill, which has
+ * 4 normal dice and a swing die, with 5 skills appearing a total of 13 times on various dice.
```
cgolubi1: Hmm.  It's not instantly obvious to me why 5 skills would naturally appear a total of 13 times, nor how many times i should expect each of the five skills to appear.  I can source-dive and figure out (but it didn't jump out at me and i don't want to spend my test time on this rather than actually testing), but i bet players will be curious what to expect too, so you should probably flesh out the button special description to give some more detail.

irilyth: That's what the button description says, or should anyway, based on what I came up with on https://github.com/buttonmen-dev/buttonmen/issues/779#issuecomment-78077615 way back when.

blackshadowshade: The actual balance of "how many instances of skills" may well need to be tweaked or balanced based on tester feedback anyway.

cgolubi1:
The flavor text doesn't match between data.button.sql and 00779_RandomBM_06.sql - there's a "with" in data.button.sql which is an "and" in the update file.

cgolubi1:
Re the discussion about number of skills appearing on each die: so does that mean the choice of how many times each skill is represented is also supposed to be random, e.g. for penta-skill, it could be 5, 5, 1, 1, 1, or 3, 3, 3, 2, 2, or anything in between?  By "is supposed to be", i mean:
* irilyth, is that what you had in mind when you wrote the description?
* shadowshade, is that you think you implemented?

And sorry if this is too nitpicky --- to be clear, it's not that i necessarily think there's a wrong answer, it's that i want to verify that the code does what it's supposed to, and knowing what we think it's supposed to do makes that easier.

blackshadowshade:
My understanding is that zero skills on a die is also okay, so 5, 5, 3, 0, 0 would be the extreme.

blackshadowshade:
I've fixed the descriptions.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1135/ (if it passes)

cgolubi1:
> My understanding is that zero skills on a die is also okay, so 5, 5, 3, 0, 0 would be the extreme.

Other way, i'm not asking about number of skills on a die, but rather number of dice across which each skill is uses.

blackshadowshade:
For Pentaskill, each skill must exist at least once, and in total, there must be 13 instances of skills in total. These are distributed randomly across all dice. So yes, there could two skills that each occur 5 times, and the others occur once only, or there could be a more even distribution, so that the frequency of the skills is 3, 3, 3, 2, 2.

irilyth:
> Re the discussion about number of skills appearing on each die: so does that mean the choice of how many times each skill is represented is also supposed to be random, e.g. for penta-skill, the number of dice on which each skill appears could be distributed 5, 5, 1, 1, 1, or 3, 3, 3, 2, 2, or anything in between?

Hmm, good question. I had in mind that it would work like RandomBM Mixed does now, but indeed, that has each skill appearing on the same number of dice (but some dice having different numbers of skills). Hmm.

My instinct is that it would be better if it were balanced -- maybe not so much in the already totally crazybeans world of Pentaskill, but Duoskill or Triskill, for example, I think it would be weird if one skill appeared on four or five dice and the other skill(s) appeared on only one. Hmm.

Would that be hard to do?

blackshadowshade:
No, it wouldn't be difficult. However, we'd have to rework the descriptions to make them more precise.

cgolubi1:
Arguably, it's going to be pretty close to balanced most of the time, because you're effectively flipping N-2 coins.  So for DuoSkill, what, you'll get a 4/1 split 25% of the time?  And for things bigger than DuoSkill, it should be less often.  That doesn't seem so bad, and is has the advantage of making these new random buttons more different from RandomBM{Fixed,Mixed}.  So i'm happy with it the way Shadowshade implemented it, but it's irilyth's request, not mine.

blackshadowshade:
I would suggest that we get this out to testing, and then have a look at what works and what doesn't.

irilyth:
That sounds good to me.

cgolubi1:
Alright, RandomAI played 5500 RandomBMPentaSkill games without a hitch, then blew up.  I'm looking into why, but my guess is that it's because there's a trip/size-change bug on RandomAI's end which i still haven't fixed, and that it's not a problem with the new button.

cgolubi1:
Hmm, so, i think this is a server-side bug.  Can someone else take a look and see if you agree or disagree?  I think Konstant dice don't reroll, and therefore the fact that the target die is also Maximum Mighty, and therefore if it did reroll it would then be showing a 12 is not germane.

Assuming it is a bug, what i'm not sure about is how many of the five skills actually have to be present on the target die to trigger the bug --- this is the kind of thing that i would have assumed was a RandomAI bug and not investigated until i did the latest round of RandomAI bugfixes.  And i'm still not 100% confident in RandomAI's trip logic, but in this case it did attempt the attack, and unless i'm missing something i agree that the attack should have been valid:

trip_rejected

cgolubi1:
Worth noting: even if i'm right about that bug, it's not a blocker for accepting RandomBMPentaSkill --- i'll file another issue for it and turn that game into a responderTest, but i don't want to bother until someone else agrees with me that it's a bug.

dwvanstone:
By just reading the descriptions, there is ambiguity in the rules of skill interaction:

- Trip: A Trip Attack is illegal if it has no chance of capturing.  
- Max: Maximum dice always roll their maximum value.

Makes sense.

- Konstant:  If another skill causes a Konstant die to reroll (e.g., Chance, Trip, Ornery), it continues to show the same value.

The intention here is clear - konstant dice are to show the same value. 

- Mighty: When a Mighty Die rerolls for any reason, it first grows from its current size to the next larger size

And so, by these rules, if a konstant Mighty die is tripped by an opponent it should grow in size but keep the value it had before. However, when a konstant Max Mighty die is tripped by an opponent there is a conflict in the rules as to what should happen. 

In general, konstant plus any of the skills that change the number of sides when rerolled seems to be in conflict in the rules as written. I would propose that if a konstant die changes size it should reroll, as if the die were coming into the playing area for the first time.

irilyth:
DVS, here's where I disagree:

> Mighty: When a Mighty Die rerolls for any reason, it first grows from its current size to the next larger size

This only triggers if the die rerolls; and Konstant dice never reroll, so Mighty is never triggered.

If something causes the Konstant die to change size *other* than a skill that triggers on reroll, then the die would be rolled when the new (different-sized) die came into play. But first the die has to change size, and I don't think the Konstant die is changing size in this case.

dwvanstone:
However:

> Konstant: If another skill causes a Konstant die to reroll (e.g., Chance, Trip, Ornery), it continues to show the same value.

If what you say is correct, that Konstant dice never reroll, then this sentence should be rewritten to clarify this. It reads to me that it does reroll but simply shows the same value each time it is rerolled.

Either way is fine by me, I'm just trying to interpret the help text.

irilyth:
You're right, that is confusing -- I was reading it as "would cause a Konstant die to reroll", but in fact that's not what it says.

I propose changing it to say that, and "it does not reroll, but continues etc". (If I'm right about the "Konstant dice never reroll" ruling, but I think I am.

blackshadowshade:
I'm pretty sure that trip dice force a konstant die to reroll, and thus this triggers the mighty skill. The mighty skill then causes the die to grow, meaning that the die needs to be rerolled on entry into the game, which means that the die value ends up being the new maximum of 12. All in all, the original trip die can't now reach the new maximum, thus the trip attack is invalid.

irilyth:
> I propose changing it to say that, and "it does not reroll, but continues etc". (If I'm right about the "Konstant dice never reroll" ruling, but I think I am.

We were talking about this on #1953, so I'm going to take this there.

irilyth:
> Worth noting: even if i'm right about that bug, it's not a blocker for accepting RandomBMPentaSkill --- i'll file another issue for it and turn that game into a responderTest, but i don't want to bother until someone else agrees with me that it's a bug.

I agree with this, and think that we should forge ahead with getting these out to the dev site, so we can test 'em. I don't think the behavior that Chaos found is a bug with the RandomBM implementations, it's just a potential bug with skill interactions in general (which is perhaps more likely to show up with Pentaskill, but RandomBM Mixed could end up with a Mighty Maximum Konstant die, and hit the same issue).

Anyone feel otherwise?

blackshadowshade:
I agree that we should be okay to go to testing on the dev site with the new skills.

cgolubi1:
BTW, i went to check RandomAI's progress, and it had gotten stuck again, on another trip attack with the same issue, but different symptoms:
trip_vs_fkm

Here, again, there's certainly a discrepancy between the "does not reroll" text you would see in the action log, and the fact that the backend (presumably) expects Maximum to take effect.  Anyway, that's not this issue, just pointing it out because it's another size-changing impact, and i'll modify my proposed list of skill interactions in #1953 to take this into account.

cgolubi1:
Anyway.  Onward.  Merge, ho!

PR #1947, opened 2016-02-08
Cause degenerate games to move to the CANCELLED game state after 200 rounds

### LINKED ISSUE #779: Play with a randomly-generated button

### LINKED ISSUE #1603: round number can only go up to 256, but the backend doesn't stop you from having more than 256 rounds

### LINKED ISSUE #1926: How to deal with degenerate games?
(I'm trying to pull this conversation out of #1920 and #1603 so we can find it later.)

Since games are not infinite [in terms of number of rounds] the site needs to end a game (even if the players may want to play a degenerate game forever). Also, players may jointly decide that a game is not worth playing and jointly decide to end the game.

I've seen several proposals:

1. Allow a game to end in a draw 
1a Consider a game a draw after N consecutive tied rounds or after N rounds
1b Allow players to agree to a draw 

2. Cancel or delete a game 
2a Consider a game canceled/deleted after N consecutive tied rounds or after N rounds
2b Allow players to jointly decide to cancel/delete a game

3. Use a coin flip to determine the winner
3a After N consecutive tied rounds or after N rounds determine winner with coin flip
3b Allow players to agree to end the game with a coin flip

My favorite of these is the third one because I think it will make tournaments easier to implement later.

blackshadowshade:
Fixes #1603. Fixes #1926.

I've not explicitly tested this code, so I presume that Chaos will run the responder tests against it and note whether degenerate games now stop appropriately.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1080/ (if it passes)


cgolubi1:
Yeah, i think given that Jenkins passed, all i have to do is run this against some Envy vs. Dead Guy games and make sure the backend and UI can deal with what happens without blowing up.

cgolubi1:
Hmm, this doesn't appear to work.  I couldn't tell you why not, but my Envy vs. Dead Guy game cheerfully blasted through round 200 and is currently at round 248 and still going.

cgolubi1:
I'm suspicious about the fact that the variable `$this->roundNumber` does not appear anywhere else in BMGame.php.  It's declared, but is it actually used or set, or do you need to use `$this->get__roundNumber()` the way the other code does?  Arguably, maybe if `$this->roundNumber` isn't ever set, you should pull it out of the code to avoid confusion.

blackshadowshade:
You're absolutely right about using get__roundNumber().

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1101/ (if it passes)

cgolubi1:
Okay, let's try this again.

cgolubi1:
Works.  The UI isn't super friendly, because when it shows a game in CANCELLED, it doesn't show W/L/T, and it doesn't give you an option to ask for more game history.  I think we should address those complaints the instant a pair of human players wind up in this situation and ask us to!

PR #1948, opened 2016-02-12
Show original recipe on the end-of-game screen

### LINKED ISSUE #1939: Game page should show correct recipe even after reserve dice are used
http://buttonweavers.com/ui/game.html?game=9945 is completed, but it's showing my opponent's button recipe as "(6) (10) (20) (20) (6) (10) r(12) r(20)". The correct recipe is "(6) (10) (20) (20) r(6) r(10) r(12) r(20)"; the page is apparently being influenced by the fact that Oaktree had added two reserve dice during the game.

blackshadowshade:
Fixes #1939.

Requires database update 01939_original_recipe.sql

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1086/


blackshadowshade:
I'm aware that this doesn't retrospectively fix the display for old games, but it does work for RandomBM* games. If we want to retrospectively fix old games, I'm thinking that we need to run some rather clever SQL, searching for buttons with reserve and adding the original recipe to the original_recipe column.

cgolubi1:
My two cents is that retroactively fixing old games is not very important for this issue, but weigh in if anyone disagrees.

blackshadowshade:
Just thinking aloud, we'd probably need something like

UPDATE game_player_map g 
SET original_recipe = (SELECT b.recipe FROM button b WHERE b.id = g.button_id)
WHERE recipe LIKE '%r%';

irilyth:
This may be a dumb question, but why limit it to buttons with Reserve dice? What would happen if you did that update on a button without Reserve dice?

blackshadowshade:
Still working on fixing the query. I'd need to customise it to also include Auxiliary, and exclude buttons that have no fixed recipe.

blackshadowshade:
If an original_recipe field is empty, then the game code reports the current recipe. This is necessary for all those cases where the button recipe is initially undefined, like Echo or RandomBM*.

cgolubi1:
Hmm.  So, looking for API regressions would be a pain and would require extra tooling, because the beginning of each test will change, as evidenced by the responderTest.php diff --- also note that we're going to have to carefully watch for other responderTest additions before this is merged, e.g. if we add a test 45 for #1949 before merging this, you'll have to update this (and if we add this first, we'll have to update test 45 before merging it).

What i really care about is:
* Does this look correct to me for new games in various states (not yet accepted, in-progress-completed) of various types (no relevant recipe changes, RandomBM-type, Echo-type, has-reserve-type)?  I'll test this by manually playing new games of those various types.
* What will the impact of this change be on existing games in various states (not yet accepted, in-progress, completed) of various types (no relevant recipe changes, RandomBM-type, Echo-type, has-reserve-type)?  I'll test this by loading a prod database backup on virtualbox, and viewing/advancing those games. 
* Will this cause any internal errors in any circumstances where the originalRecipe either can't be determined, or is looked for and isn't set yet?  I'll test this by playing a lot of new games, probably using AWS so i can run a large number of games over a day or two.

blackshadowshade:
(I'll need to resolve the changes to responderTest with the newly merged tests, so that this passes again)

blackshadowshade:
Right, that should be correctly merged. Let's see if Jenkins agrees.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1089/ (if it passes)

blackshadowshade:
Oops, I realised that I had pushed some of the stuff I was doing for Turbo Swing on this branch. I've reverted that commit, and I'll get a clean Jenkins run.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1092/ (if it passes)

cgolubi1:
Okay.  So one thing of note is that, if you have the following sequence of events:
* Create a game where player A has a reserve die
* Player A loses a round
* Player A adds a reserve die
* This update gets applied
* Player A loses a second round
* Player A adds a second reserve die
* The game ends

then on the game end screen player A's recipe will show up as e.g. `(4) (6) (8) (12) r(8) r(8) (10) r(12)` --- that is, the reserve die which was added before the update will show up as part of the original recipe, and the second reserve die won't.

This is no skin off my nose, and i don't think it's worth any effort to fix; just flagging it for due diligence.

cgolubi1:
Hmm, i do think that what you're doing here leads to some rather inconsistent database state between different cases, though.  Here's some games where the button is Avis, which has a fixed recipe that does not change during the game (the first game was completed before the database updated, the second game was in progress during the database update, and the third game was created but still open during the database update).  You can see that original_recipe is never set:
```sql
mysql> select * from game_player_view where button_name='Avis' and player_id=1;
+---------+-----------+-----------+-----------------+------------+----------+--------------------+--------------------+--------------+---------------+----------------+----------+------------------+---------------------+--------------------+------------------+---------------------+-------------+----------+-------------------+-------------+---------------+
| game_id | player_id | button_id | original_recipe | alt_recipe | position | did_win_initiative | is_awaiting_action | n_rounds_won | n_rounds_lost | n_rounds_drawn | handicap | is_player_hidden | last_action_time    | was_game_dismissed | is_button_random | has_player_accepted | player_name | autopass | fire_overshooting | button_name | n_target_wins |
+---------+-----------+-----------+-----------------+------------+----------+--------------------+--------------------+--------------+---------------+----------------+----------+------------------+---------------------+--------------------+------------------+---------------------+-------------+----------+-------------------+-------------+---------------+
|       1 |         1 |       256 | NULL            | NULL       |        0 |                  0 |                  0 |            0 |             1 |              0 |        0 |                0 | 2016-02-28 21:35:28 |                  0 |                0 |                   1 | c1          |        1 |                 0 | Avis        |             1 |
|       2 |         1 |       256 | NULL            | NULL       |        0 |                  1 |                  0 |            1 |             0 |              0 |        0 |                0 | 2016-02-28 21:49:39 |                  0 |                0 |                   1 | c1          |        1 |                 0 | Avis        |             1 |
|       3 |         1 |       256 | NULL            | NULL       |        0 |                  1 |                  0 |            0 |             1 |              0 |        0 |                0 | 2016-02-28 21:52:37 |                  0 |                0 |                   1 | c1          |        1 |                 0 | Avis        |             1 |
+---------+-----------+-----------+-----------------+------------+----------+--------------------+--------------------+--------------+---------------+----------------+----------+------------------+---------------------+--------------------+------------------+---------------------+-------------+----------+-------------------+-------------+---------------+
3 rows in set (0.00 sec)
```

By contrast, here's some games with RandomBMMixed, which has a randomly chosen button which is nevertheless not updated during the game (the distribution of game states is as before):
```sql
mysql> select * from game_player_view where button_name='RandomBMMixed';
+---------+-----------+-----------+-------------------------------+-------------------------------+----------+--------------------+--------------------+--------------+---------------+----------------+----------+------------------+---------------------+--------------------+------------------+---------------------+-------------+----------+-------------------+---------------+---------------+
| game_id | player_id | button_id | original_recipe               | alt_recipe                    | position | did_win_initiative | is_awaiting_action | n_rounds_won | n_rounds_lost | n_rounds_drawn | handicap | is_player_hidden | last_action_time    | was_game_dismissed | is_button_random | has_player_accepted | player_name | autopass | fire_overshooting | button_name   | n_target_wins |
+---------+-----------+-----------+-------------------------------+-------------------------------+----------+--------------------+--------------------+--------------+---------------+----------------+----------+------------------+---------------------+--------------------+------------------+---------------------+-------------+----------+-------------------+---------------+---------------+
|       4 |         2 |     10003 | NULL                          | B(4) zf(4) (10) f(12) zB(20)  |        1 |                  1 |                  0 |            3 |             0 |              0 |        0 |                0 | 2016-02-28 21:38:42 |                  0 |                0 |                   1 | c2          |        1 |                 0 | RandomBMMixed |             3 |
|       5 |         2 |     10003 | tM(4) Md(6) t(6) (10) d(20)   | tM(4) Md(6) t(6) (10) d(20)   |        1 |                  0 |                  0 |            3 |             0 |              0 |        0 |                0 | 2016-02-28 21:45:37 |                  0 |                0 |                   1 | c2          |        1 |                 0 | RandomBMMixed |             3 |
|       6 |         2 |     10003 | (6) b(10) s(10) sg(12) bg(20) | (6) b(10) s(10) sg(12) bg(20) |        1 |                  0 |                  0 |            3 |             0 |              0 |        0 |                0 | 0000-00-00 00:00:00 |                  0 |                0 |                   1 | c2          |        1 |                 0 | RandomBMMixed |             3 |
+---------+-----------+-----------+-------------------------------+-------------------------------+----------+--------------------+--------------------+--------------+---------------+----------------+----------+------------------+---------------------+--------------------+------------------+---------------------+-------------+----------+-------------------+---------------+---------------+
3 rows in set (0.00 sec)
```

and here's some games with RandomBMAnime, whose recipe does change during the game (ditto):
```sql
mysql> select * from game_player_view where button_name='RandomBMAnime';
+---------+-----------+-----------+------------------------------------------+----------------------------------------+----------+--------------------+--------------------+--------------+---------------+----------------+----------+------------------+---------------------+--------------------+------------------+---------------------+-------------+----------+-------------------+---------------+---------------+
| game_id | player_id | button_id | original_recipe                          | alt_recipe                             | position | did_win_initiative | is_awaiting_action | n_rounds_won | n_rounds_lost | n_rounds_drawn | handicap | is_player_hidden | last_action_time    | was_game_dismissed | is_button_random | has_player_accepted | player_name | autopass | fire_overshooting | button_name   | n_target_wins |
+---------+-----------+-----------+------------------------------------------+----------------------------------------+----------+--------------------+--------------------+--------------+---------------+----------------+----------+------------------+---------------------+--------------------+------------------+---------------------+-------------+----------+-------------------+---------------+---------------+
|       4 |         1 |     10002 | NULL                                     | (6) (6) (10) (20) r(8) (10) r(10) (10) |        0 |                  0 |                  0 |            0 |             3 |              0 |        0 |                0 | 2016-02-28 21:38:36 |                  0 |                0 |                   1 | c1          |        1 |                 0 | RandomBMAnime |             3 |
|       5 |         1 |     10002 | (4) (6) (8) (12) r(8) r(8) (10) r(12)    | (4) (6) (8) (12) r(8) (8) (10) r(12)   |        0 |                  1 |                  0 |            0 |             3 |              0 |        0 |                0 | 2016-02-28 21:46:09 |                  0 |                0 |                   1 | c1          |        1 |                 0 | RandomBMAnime |             3 |
|       6 |         1 |     10002 | (10) (10) (12) (12) r(4) r(6) r(6) r(10) | (10) (10) (12) (12) r(4) (6) (6) r(10) |        0 |                  1 |                  0 |            0 |             3 |              0 |        0 |                0 | 2016-02-28 21:53:27 |                  0 |                0 |                   1 | c1          |        1 |                 0 | RandomBMAnime |             3 |
+---------+-----------+-----------+------------------------------------------+----------------------------------------+----------+--------------------+--------------------+--------------+---------------+----------------+----------+------------------+---------------------+--------------------+------------------+---------------------+-------------+----------+-------------------+---------------+---------------+
3 rows in set (0.00 sec)
```

cgolubi1:
So, my question based on those examples --- particularly in the middle game in the tables which were already in progress at update time, why does it make sense for original_recipe to be null for Avis and set for RandomBMMixed?

Implementation-wise, why set original_recipe when you're *updating* the button recipe?  Why not set it at game start time (when you're initially instantiating the recipe) and at no other time?  I don't see what bug this inconsistency causes in the current UI (other than the weirdness i noted in https://github.com/buttonmen-dev/buttonmen/pull/1948#issuecomment-189951131), but i also don't see any reason to introduce inconsistency without a benefit.

blackshadowshade:
For me, the NULL is a useful indication that nothing has changed with the recipe, which is going to be the case most of the time. I was thinking that this might simplify debugging in the future. If you'd prefer us to always hold the original recipe, we can do that, but it means that we will need to run some SQL to update all old games so that they have a valid non-empty original_recipe field.

cgolubi1:
Okay, what you have submitted right now is preferable to your interpretation of my suggestion.  So let's keep it as is.  :>)

blackshadowshade:
The issue with doing things at game create time is that we don't know (without adding more logic) whether a recipe is going to change at game create time. The only time that we know that a recipe is going to change is when it actually changes.

Thus, if we do things at game create time, we would need to *always* assign original_recipe.

There is no technical reason not to do this, apart from padding out the database with more data.

cgolubi1:
> Thus, if we do things at game create time, we would need to always assign original_recipe.

Right, that's exactly my suggestion --- always assign original_recipe at game create time, and then you don't need to use logic to determine whether the recipe is changing later, because you always just have it.  I don't think the amount of database storage space is significant.

blackshadowshade:
In that case, then I think that we should iterate on the SQL needed to backfill original_recipe for current games. Of course, we could just fill with what the recipe is currently, but I don't think that it'll be awfully difficult to get things correct (or at least more correct).

I'd appreciate some help on this to get the syntax right.

blackshadowshade:
Actually, thinking about this some more, it might be as simple as filling recipes directly out of the button table, and then for those that are still empty, copying the current recipe.

cgolubi1:
I'm starting to feel a bit frustrated by this, because i feel like i said explicitly that my opinion is that the "if" statement in `get_originalRecipe()` is harmless, and we should keep it that way as an alternative to backfilling recipes.  And yet you keep saying "okay, well, if we do it your way, we'll need to backfill recipes".  I don't have a problem with it if you want to backfill recipes.  Go for it, have fun.  I think it's the wrong thing to do, but i don't think it's very wrong.  But i will be kind of annoyed if you do a lot of work on backfilling recipes and it turns out to be a mess (which i think it will be), and i get told it was what i asked you to do.

So i'm going to try one more time.  I'm happy with the logic you've submitted for `get_originalRecipe()`.  It's fine with me to have an "if" statement in there *purely for backwards compatibility with older database entries*.  The thing i don't like is *updating the table* conditionally based on whether the recipe is changing, because i think *that's* thorny logic that's hard to get right all the time and will cause problems.  So i think the full logic should be:
1. At game creation/button selection time *and at no other time*, populate original_recipe with the original recipe
2. In `get_originalRecipe()`, return the original_recipe if it exists, and return recipe if it doesn't (where the latter case will be true for games which were created before this change was made).

Again, if you really really want to backfill, go for it.  I don't think it's a good use of development or testing time, but if you feel strongly about it, i'm willing to go along with it.

cgolubi1:
Having gotten that off my chest: i'd be happy to have a conversation explicitly about the subject of how we can minimize remaining time spent on this pull.  I don't think it's a very important pull, and i think we are on the edge of spending many more hours of both of our time on it.  How can we bring this back to what's actually reasonable?  I'm pretty willing to finish testing and merge it as it is right now if that's easiest.

cgolubi1:
We talked about this offline:
* shadowshade doesn't want to see game/button entries with original_recipe=NULL unless that is a valid state going forward (i.e. he doesn't like it as just a backwards-compatibility entry)
* i don't want to see anything in the original_recipe field which isn't actually the original entry

So the only reasonable compromise is to backfill the entries.  Shadowshade is going to follow up on his original suggestion of filling in the entries from the button->recipe where that exists.  Then: (1) for Echo and Zero we can copy the opponent's original_recipe; (2) for any buttons which still don't have recipes, copy from the current game/button->recipe; (3) fixup RandomBMAnime by hand/script to go back to the original four-reserves format of the recipe.

blackshadowshade:
Okay, I think that this should be ready to go.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1099/ (if it passes)

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1128/ (if it passes)

cgolubi1:
Okay, so now that the logic has been simplified, i'm actually a lot more comfortable with this pull.  The responderTest diff shows that the right things appear to be happening.  My updated test plan is really just:
* do the database update vs load test
* try this out with a couple of by-hand games and make sure it actually looks right
* Run this database update on backups of the dev and prod databases, and look into all the buttons where recipe != originalRecipe, and where originalRecipe == NULL, and make sure i understand both cases and what to do about the latter case.

cgolubi1:
Ah, the update test found a hidden merge conflict:
```diff
sandbox,[/buttonmen/deploy/database/updates],23:35(1)$ diff ~/buttonmen_db.rebuilt.sql ~/buttonmen_db.updated.sql 
682d681
<   `is_on_vacation` tinyint NOT NULL,
1090c1089
< /*!50001 VIEW `game_player_view` AS select `m`.`game_id` AS `game_id`,`m`.`player_id` AS `player_id`,`m`.`button_id` AS `button_id`,`m`.`original_recipe` AS `original_recipe`,`m`.`alt_recipe` AS `alt_recipe`,`m`.`position` AS `position`,`m`.`did_win_initiative` AS `did_win_initiative`,`m`.`is_awaiting_action` AS `is_awaiting_action`,`m`.`n_rounds_won` AS `n_rounds_won`,`m`.`n_rounds_lost` AS `n_rounds_lost`,`m`.`n_rounds_drawn` AS `n_rounds_drawn`,`m`.`handicap` AS `handicap`,`m`.`is_player_hidden` AS `is_player_hidden`,`m`.`last_action_time` AS `last_action_time`,`m`.`was_game_dismissed` AS `was_game_dismissed`,`m`.`is_button_random` AS `is_button_random`,`m`.`has_player_accepted` AS `has_player_accepted`,`p`.`name_ingame` AS `player_name`,`p`.`autopass` AS `autopass`,`p`.`fire_overshooting` AS `fire_overshooting`,(length(`p`.`vacation_message`) > 0) AS `is_on_vacation`,`b`.`name` AS `button_name`,`g`.`n_target_wins` AS `n_target_wins` from (((`game_player_map` `m` left join `player` `p` on((`m`.`player_id` = `p`.`id`))) left join `button` `b` on((`m`.`button_id` = `b`.`id`))) left join `game` `g` on((`m`.`game_id` = `g`.`id`))) */;
---
> /*!50001 VIEW `game_player_view` AS select `m`.`game_id` AS `game_id`,`m`.`player_id` AS `player_id`,`m`.`button_id` AS `button_id`,`m`.`original_recipe` AS `original_recipe`,`m`.`alt_recipe` AS `alt_recipe`,`m`.`position` AS `position`,`m`.`did_win_initiative` AS `did_win_initiative`,`m`.`is_awaiting_action` AS `is_awaiting_action`,`m`.`n_rounds_won` AS `n_rounds_won`,`m`.`n_rounds_lost` AS `n_rounds_lost`,`m`.`n_rounds_drawn` AS `n_rounds_drawn`,`m`.`handicap` AS `handicap`,`m`.`is_player_hidden` AS `is_player_hidden`,`m`.`last_action_time` AS `last_action_time`,`m`.`was_game_dismissed` AS `was_game_dismissed`,`m`.`is_button_random` AS `is_button_random`,`m`.`has_player_accepted` AS `has_player_accepted`,`p`.`name_ingame` AS `player_name`,`p`.`autopass` AS `autopass`,`p`.`fire_overshooting` AS `fire_overshooting`,`b`.`name` AS `button_name`,`g`.`n_target_wins` AS `n_target_wins` from (((`game_player_map` `m` left join `player` `p` on((`m`.`player_id` = `p`.`id`))) left join `button` `b` on((`m`.`button_id` = `b`.`id`))) left join `game` `g` on((`m`.`game_id` = `g`.`id`))) */;
1142c1141
< -- Dump completed on 2016-05-21 23:35:04
---
> -- Dump completed on 2016-05-21 23:34:59
sandbox,[/buttonmen/deploy/database/updates],23:35(1)$ 
```

cgolubi1:
Okay, at a glance the update of the prod database looks very good --- there are only two games with original_recipe NULL, both of which are RandomBM* games which were cancelled before starting.  And the map entries where original_recipe != alt_recipe look like cases of reserve and auxiliary dice, and the original_recipe entries look correct per spot checks in those cases.

cgolubi1:
I think this can be merged once the view update merge conflict has been fixed.

blackshadowshade:
Right, the database update file has been corrected.

blackshadowshade:
Remember that there's a last manual change if we want to have absolutely everything correct, namely fixing RandomBMAnime original recipes. I'm personally not too fussed about them, but just making the point that they're not fixed automatically.

blackshadowshade:
So, it's been a while since this branch was merged with master. The new Jenkins run to check is http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1138/ (if it passes).

cgolubi1:
> Remember that there's a last manual change if we want to have absolutely everything correct, namely fixing RandomBMAnime original recipes. I'm personally not too fussed about them, but just making the point that they're not fixed automatically.

Yep.  I agree that we don't need to worry about this for past games --- future games will do the right thing.

cgolubi1:
Your Jenkins run failed because of #1999

blackshadowshade:
Argh. I thought I'd already merged that pull, but obviously not.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1141/ (if it passes)

cgolubi1:
Okay, jenkins passed.  Redoing the DB update check.

cgolubi1:
Looks good.  Merging.

PR #1952, opened 2016-02-16
migrate all API level tests to generate test data in responderTest and consume it in DummyApiResponder

### LINKED ISSUE #1858: all API tests should work the way the loadGameData test works

cgolubi1:
* Fixes #1858
* Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/429/ (if it passes)

Testing done: ran responder tests twice without recreating the database in between, and verified that UI tests all still pass in that condition.


cgolubi1:
In case this helps, here's a gist of the output of `git diff origin test/src/api/responderTest.php`: (https://gist.github.com/cgolubi1/81234d8765bb86d54912)

I'm not sure why github is so obnoxious about this, but there it is.

blackshadowshade:
So, this passes on my local WAMP stack. It produces a huge number of directories in src/api/dummy_data as expected.

Once we have a clean Jenkins run with the new code, I'm happy to merge this.

blackshadowshade:
Jenkins looks good. Merging.

Issue #138, opened 2013-11-19
Acquire button images


blackshadowshade:
We need images for all the buttons, so that we can display them in the game.

blackshadowshade:
So, I've managed to start up some player-driven button image acquisition and processing, and I'm optimistic that this might actually bring us loads of images.

I've also asked alwayslurking whether he's willing to write a short wiki for us to say how he processes the scanned images within Gimp.

Hurtay:
I've met (via chat) a player with quite a few physical buttons he'll scan for me. Sadly none of them is a button we haven't got the image of yet. So I'll try to improve the images we've got already. If anyone somehow finds new images I now can get them into the right format with lightreflex/shadow etc.

blackshadowshade:
@Hurtay pointed me to some interesting button art online at BoardGameGeek:
- EuroGames set (using Soldiers recipes): https://boardgamegeek.com/image/337399/button-men
- Ameritrash set (using Soldiers recipes): https://boardgamegeek.com/image/337400/button-men
- BoardGameGeek button: https://boardgamegeek.com/image/1411068/button-men

The buttons themselves don't bring any interesting gameplay, in my opinion, but the art might be interesting for some of you.

blackshadowshade:
I'm going to close this, since I think we've pretty much got all the images that we're going to get. Any further images can be added as we find them.

PR #1954, opened 2016-02-21
New and updated button images

blackshadowshade:
Addresses part of #138.

Kaori has created new images of Elihu and Frasquito which have a light reflex in the top left.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1091/ (if it passes)


blackshadowshade:
I've just noticed that the background colour isn't set to transparent. I'll get onto her about fixing that.

cgolubi1:
Since there's been no iteration on this yet, let's close it, and open another pull request when the branch is actually ready.  (Reopen if you disagree.)

blackshadowshade:
Okay, reopening this now that it's been updated. I've also changed the title to make it more general.

blackshadowshade:
So, the new images are up for discussion. Are all of these improvements, or would we rather keep some of the old images?

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1103/ (if it passes)

blackshadowshade:
More images added.

blackshadowshade:
I think these are clearly better: Aldric, Bruno, Dirgo, Dr Speculo, Elihu, Flaire, Frasquito, Guillermo, Kiki, Luce, Nerni, Nikolai, Pappy, Synthia, The GM

I think that these are clearly worse: Mothman, Riff, Torg, Vincent, Yeti, Zoe

I think that these are about the same: Aylee, Bun-Bun

Harry Pudding's button has the dice in the recipe in a different order, which is interesting.

irilyth:
I generally agree with blackshadowshade's comments. Couple questions:

* Mothman didn't look any worse to me, and was maybe a little sharper. What didn't you like about it?
* I was amused that neither Torg has the name parallel to the horizontal -- I think *that* would look better.

I don't feel strongly about any of those.

> Harry Pudding's button has the dice in the recipe in a different order, which is interesting.

Wacky. I think it's better to have them in the same order as the recipe in the DB. How'd they even end up different? Is one of them not just a scan of the actual button?

blackshadowshade:
For Mothman, I guess I don't like the look of the numbers in the recipe circles, which look artificial.

As for Harry Pudding, I'm guessing that the one that we currently have is a scan of the original button, while the other one has had the recipe Photoshopped on. It's only a guess though.

blackshadowshade:
Okay, I've reverted the images for Bun-Bun, Riff, Torg, Vincent, Yeti, and Zoe.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1108/ (if it passes)

blackshadowshade:
This only alters existing images, so no new Jenkins run should be necessary.

cgolubi1:
Okay, my understanding from the discussion so far is that y'all would be fine with all these new images, so if i want to merge them, i can.  It sounds like y'all were on the fence about the new Mothman, but decided to keep it anyway?  And i'm not clear from this discussion which Harry Pudding has the same button order as the recipe, so i'll check that.

cgolubi1:
So, for testing, i'm going to bring up a site and:
* Start games with all these buttons and look for problematic artifacts
* Look at the Sluggy Freelance buttonset page - the new Aylee and Kiki are significantly differently shaded (more green, less brown) than the old ones.  Is that good, or does it mean the set looks incongruous as a whole, since you reverted some of the other Sluggy buttons?

cgolubi1:
Games look fine.

I don't see a problem with the Sluggy set overall --- all of the buttons are colored differently from each other, so the pattern i saw in the shading change of Aylee and Kiki doesn't make them look incongruous.

Merging this.

PR #1957, opened 2016-02-24
Changed Zodiac recipes, added images and flavour text

blackshadowshade:
Continues to address #1944.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1094/

PR #1949, opened 2016-02-13
Update prevOptValueArray when adding a reserve die

blackshadowshade:
Fixes #1883.

Tested bobby vs bobby.

We may need a responder test for this to lock in this behaviour.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1087/

PR #1946, opened 2016-02-08
pull changes from master to dev on 2016-02-08

cgolubi1:
This update brings these approved pulls to the dev branch:
* User-visible:
    * #1945: add the Zodiac button set
    * #1923: change UI error message when the server is not reachable
* Ideally not user-visible:
    * #1929: in the API, make swingRequestArray empty except during swing selection
    * #1935: in the backend, change the REQUESTED game state to CANCELLED
    * #1938: add more PHPDoc code comments

Database updates:
```
UPDATES="01926_CANCELLED_01.sql 01944_zodiac_01.sql"
```

Once this pull is merged, i will load the changes on dev.buttonweavers.com as part of #1941.

Special actions: add a patch to hack Random Button selection to pick one of the Zodiac buttons, for playtesting:
```diff
$ cat ~/zodiac.patch 
diff --git a/src/engine/BMInterface.php b/src/engine/BMInterface.php
index 7bef6f5..0e33d53 100644
--- a/src/engine/BMInterface.php
+++ b/src/engine/BMInterface.php
@@ -1067,7 +1067,7 @@ class BMInterface {
             }

             if (empty($allButtonData)) {
-                $allButtonData = $this->get_button_data(NULL, NULL, TRUE);
+                $allButtonData = $this->get_button_data(NULL, 'Zodiac', TRUE);
                 $nButtons = count($allButtonData);
             }
```

PR #1928, opened 2016-01-17
Fixed bug where unsuccessful trip attacks targeting radioactive fail to split

blackshadowshade:
Fixes #1919.

This pull request fixes a regression that was apparently introduced by Rage.

The issue will probably need a responder test to lock in the correct behaviour.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1052

PR #1959, opened 2016-02-28
pull changes from master to dev on 2016-02-28

cgolubi1:
* User-visible:
    * #1957: change some Zodiac recipes, and add Zodiac art
    * #1928: fix bug in which die fails to split after targeting a radioactive die with an unsuccessful trip attack
    * #1949: fix problem with displaying option reserve die values from previous rounds, affecting bobby 5150
* Ideally not user-visible:
    * #1952: migrate all API tests to a new test format

Database updates:
```
UPDATES="01944_zodiac_02.sql"
```

Once this pull is merged, i will load the changes on dev.buttonweavers.com as part of #1941.

Special actions: leave the zodiac patch to BMInterface.php, which was installed in #1946, in place.


cgolubi1:
There's only one database update, and it's simple, so i'm going to load this directly on dev, and then verify that existing games containing the buttons with the changed recipes continue to load.

cgolubi1:
Updated; only database changes are in the button recipes and flavor text, as expected.

PR #1960, opened 2016-02-28
Fixed recipe of Libra, and flavour text of Aries and Capricorn

blackshadowshade:
Continues to address #1957.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1097/

PR #1958, opened 2016-02-28
Update network error message per tester feedback

cgolubi1:
* Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/430/ (if it passes)

PR #1963, opened 2016-03-04
pull changes from master to dev on 2016-03-03

cgolubi1:
* User-visible:
    * #1958: update message printed when the UI can't access the server
    * #1960: fixed recipe of Libra, and flavour text of Aries and Capricorn
* Ideally not user-visible: none

Database updates:
```
UPDATES="01944_zodiac_03.sql"
```

Once this pull is merged, i will load the changes on dev.buttonweavers.com as part of #1941.

Special actions: remove the zodiac patch to BMInterface.php, which was installed in #1946; we don't need it any more.


cgolubi1:
I'll just go ahead and merge this and load it directly on dev --- the database change is small.

Issue #1961, opened 2016-03-01
Art for 50 States set


irilyth:
It'd be cool to have some art for the 50 States set. Three possible sources:

* If we don't mind non-round images, we could use state flags.
* If we prefer round images, we could use state seals, or the backs of state quarters.

The flags are probably slightly more recognizable, and are surely in the public domain. The quarter designs seem to be public domain as well, with some caveats that we could investigate if we decide we like those. I don't know about the state seals, and think they're the least cool option, unless we prefer round and don't like quarters (or they turn out to have a problematic or complicated copyright status).

jimmosk:
I definitely vote for round, to be consistent with everything else and be true to the Button aspect.  But I think circular versions of the state flags (just crop a circle from the center) would work just fine, so we needn't eliminate that possibility.
I lean towards state quarters, because I think that's the most widely recognized of these options.

irilyth:
Another nice feature of the quarters idea is that they'd be pretty visually distinct from everything *else*. It's cool when sets have a consistent art style (when that makes sense).

blackshadowshade:
Here's a suggestion by Kaori:

[20states.pdf](https://github.com/buttonmen-dev/buttonmen/files/160628/20states.pdf)

blackshadowshade:
Personally, I think these look great!

blackshadowshade:
And here are the other 30 states.

[30states.pdf](https://github.com/buttonmen-dev/buttonmen/files/162278/30states.pdf)

blackshadowshade:
So, Kaori would like discussion on what you like and what you don't like. Some things that's she's mentioned:
- whether it's okay that she's slightly changed wording or shapes
- whether the white circles are the right size
- whether you like the images being off-centre
- what you think about the font
- whether the wording should be straight or bent
- whether each button needs a separate name when it's obvious in the image

AdmiralJota:
Could you give some examples of what you're referring to by wording or shape changes?

AdmiralJota:
From in-game chat with Kaori:

> I've erased lots of wording (state names, dates, motto or whatever) where it was just the plainly written letters because I thought it irritating on the image (i.e. Delaware, Illinois).
> Where the wording is part of the image I've kept it. Except Arkansas and California but I'm not happy with these and would prefer to replace at least CA's name with the usual font.
> 
> Shape changes were even smaller:
> I had to enlarge top and bottom in NC, TX.
[...]

PR #1968, opened 2016-03-12
Added images for 50 states buttons

blackshadowshade:
Addresses part of #1961.

@Hurtay has provided me with these images of the 50 States buttons.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1105/


blackshadowshade:
I've checked the recipes against the buttons on my local MAMP server (http://localhost:8178/ui/buttons.html?set=50%20States), and they seem to be correct except for Texas, since we use a backtick for Warrior and not an apostrophe.

The representation of warrior dice that we currently have (Shadow Warriors, fernworthy) is to show them as diamonds on the left side, while reserve dice (Sailor Moon, Heather) are shown as circles on the left side. I would suggest that we stick with this convention, which would mean that the following buttons would need to be reworked: Illinois, New York, Texas.

Hurtay:
I don't know how to change the uploaded image files or how to upload new files. 
But here are the correct files.
image
image
image

blackshadowshade:
Okay, those images are now updated.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1106/

blackshadowshade:
The remaining images have also been added.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1107/

blackshadowshade:
I've fixed the responder tests so that they now pass.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1109/ (if it passes)

cgolubi1:
Okay, so i think all i need to do here is bring up a test site, create some games, make sure the images actually show up, and look for any dealbreaker issues with how the images fit into the boxes.

cgolubi1:
I don't see any issues.  Thanks very much for putting these together, @Hurtay!

PR #1970, opened 2016-03-26
Better images

blackshadowshade:
@Hurtay has newly-scanned button images that appear to be clearer and better.

Since these images replace old images, there is no need for Jenkins.


blackshadowshade:
This only changes existing button images, so no new Jenkins run should be necessary.

cgolubi1:
I don't see problems with any of these.  I'll bring up a site under virtualbox and create games with the new images to make sure there are no surprise background or size problems or anything.

cgolubi1:
They all look fine.  Merging.

PR #1977, opened 2016-04-02
Changed skill tooltips so that they trigger on click, not mouseover

### LINKED ISSUE #1697: Make "Skills in this game" section more user-friendly

blackshadowshade:
Fixes #1697.

This applies Jota's suggestion to the skill hint section, meaning that they trigger on click instead of on mouseover.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1117/


blackshadowshade:
This doesn't touch any unit tests, so no new Jenkins run should be necessary.

blackshadowshade:
I've just merged in master, since we've had a lot of changes to the UI since this was written.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1196/ (if it passes)

cgolubi1:
It failed on BMGameTest.test_morph_on_unsuccessful_trip_bug (let's fix that, #2068), so i restarted jenkins:

New run: http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1197/ (if it passes)

cgolubi1:
Assuming jenkins passes this time, i'll start looking at this shortly --- this is a different aspect of the UI, so i am going to have to come back up to speed on how this code works now.  And one reason i've been hesitating is that i'm not sure the functionality change is what browser users want, so i'll want to start a conversation about that.  So basically i think it would be reasonable to expect this to take awhile and involve some arguing, but it's on my plate now modulo interrupts from Turbo.

AdmiralJota:
If it's not what we want, then it shouldn't be too hard to tweak it to try different variations on this idea. 

The site currently displays the text when you mouse over an element and makes it disappear when you mouse out. This change makes it ignore mousing over/out, and toggles the text on/off when you click.

One alternative would be to display the text on _either_ clicking or mousing over, and to hide it on _either_ clicking again or mousing out. Then desktop users would continue to experience a UI similar to what we have now, while mobile users would be able to show/hide the text by tapping.

If this goes up on the dev site, then I think that might help facilitate the discussion a bit.

irilyth:
I also like hover as well as click (and said so on #1697, a while ago).

blackshadowshade:
FYI @AdmiralJota, I'd like you to take the lead on suggesting changes/fixes based on suggestions here. I'm happy to submit them for you, but I think you have a better grasp on how to get this working the way that it should work.  :)

AdmiralJota:
Sure. Here's a version that accepts both mousing over and clicking as triggers: http://jsfiddle.net/s8w0qehv/

src/ui/js/Game.js
```diff
@@ -1917,20 +1918,39 @@ Game.pageAddSkillListFooter = function() {
...
+          }
+        },
+      'mouseout': function(event) {
+        event.stopImmediatePropagation();
+      }
+    });
 
     if (info.code) {
       dieSkillSpanArray.push(skillSpan);
+      dieSkillSpanArray.push(tooltipDiv);
```
cgolubi1: This seems weird at a glance, like you have an array of skillSpan, tooltipDiv, skillSpan, tooltipDiv, as opposed to each tooltipDiv being in some way associated with its corresponding skillSpan.  I'll look at the HTML in testing.

cgolubi1: Yeah, indeed, here's what the HTML looks like:

span_div_span_div

That seems untidy --- is there a reason we can't encapsulate each skill's label and its help text into one HTML object of some sort?

cgolubi1:
Okay, another issue --- from my POV using my browser, the look-and-feel of the mouseover text is a lot more usable the old way than the new way.  Here's an example (these aren't quite the same size, because i had to do a timed screen capture to grab the mouseover text, but you should be able to compare the sizes of the action log entries to get the idea:

Here's the old way:

small_visible_help_text

And here's the new way per this PR:

big_offscreen_popup_text

The two major things which seem worse about the new way to me:

1. The popup text generated by this PR is in a giant round font which i find distracting, and which is a bad match for our verbose skill descriptions in terms of actually fitting text on the page, whereas the previous mouseover text was in a pretty reasonable font.
2. In both firefox and safari, the popup help text is brought up so that only some of it is visible on the screen, and i have to scroll the screen to see all of it, whereas the mouseover text is smaller to begin with, but also can expand outside of the browser window, so i don't have to scroll to see the text i just brought up.

cgolubi1:
Okay, i can't promise not to have other comments, but i think that gives us enough to talk about in terms of ways in which the new behavior could be a regression for browser users, and what we can do about them.

cgolubi1:
Oh, i guess another thing is: if we're going to make the links clickable, IMO we should make them keyboard-selectable too to match our other keyboard-selectable elements.  I know they weren't keyboard-selectable before, but they weren't clickable before, and we've generally been trying to make all clickable elements keyboard-selectable as well.  Here's a vimperator follow overlay of the page --- everything vimperator highlights is a link which is keyboard-selectable, and as you can see all of our existing links are keyboard-selectable, but the new help text isn't:

skills_not_keyboard_selectable

For my personal use case, if the links were keyboard selectable, i wouldn't care so much whether they were mouse-over-able, but that's definitely a thing where other browser players may feel very differently.

cgolubi1:
Enumerating all those issues in one place:

1. Players may prefer to still be able to mouseover skill text from a browser
2. HTML output of skill descriptions is inelegant, because it's a flat interleaving of skill descriptions and tooltips, rather than having each tooltip be part of the same object as the skill description.
3. The popup skill text defaults to a big space-inefficient font, and used to default to a smaller font
4. The popup skill text is centered, and used to default to being left-justified
5. The popup skill text box runs off the screen, and you have to scroll the browser window to see it, whereas it used to hover outside of the browser window if appropriate so it was all immediately visible
6. The skill text click elements should be keyboard-selectable

AdmiralJota:
> Okay, another issue --- from my POV using my browser, the look-and-feel of the mouseover text is a lot more usable the old way than the new way.

I believe the things you're talking about can be found in the js/extern/jquery-ui.css file. In particular, the rules for .ui-widget include:

    .ui-widget {
    	font-family: Verdana,Arial,sans-serif/*{ffDefault}*/;
    	font-size: 1.1em/*{fsDefault}*/;
    }
    .ui-widget .ui-widget {
    	font-size: 1em;
    }
    .ui-widget input,
    .ui-widget select,
    .ui-widget textarea,
    .ui-widget button {
    	font-family: Verdana,Arial,sans-serif/*{ffDefault}*/;
    	font-size: 1em;
    }

We could remove the font-family lines and change that first font-side to 1em, for instance. We could also add a rule for ".ui-widget div { text-align: left; }" to override the automatic centering that the site does on all divs in gui.css. 

(By the way, why _do_ we have CSS to automatically center every single div on the site that doesn't explicitly override it?)

AdmiralJota:
> HTML output of skill descriptions is inelegant, because it's a flat interleaving of skill descriptions and tooltips, rather than having each tooltip be part of the same object as the skill description.

For this, you can do something like:

    skillSpan.apped(toolTipDiv);

...after creating the `toolTipDiv `(e.g. around line 1930).

> The skill text click elements should be keyboard-selectable

And for that, when you're creating the `skillSpan` (around line 1923), include this:

    'tabindex': '0',

@blackshadowshade: Is there anything else I should be proposing fixes/changes for, or does that cover it?

blackshadowshade:
I'll let Chaos change the appropriate lines in her CSS files and comment on whether that meets her aesthetic wishes.

One question, though. Should we be altering the jquery-ui.css file directly, or should we be overriding the behaviour in our gui.css? My question relates to whether the jquery-ui.css file is a standard file (meaning that we could eventually reference the version hosted by Google if we wanted to), or rather a customised file that is always hosted by each website separately.

AdmiralJota:
You can download different versions of jquery-ui.css from jqueryui.com for different themes, so I don't think creating our own "theme" will be a problem. If we want to be extra safe, we could rename the file to indicate that it's our customized version and add a comment at the top, to make sure we don't confuse future generations of BM devs.

blackshadowshade:
Okay, I've downloaded a whole pile of themes, and each theme does indeed have a different jquery-ui.css. Because of this, I'm happy just to customise our version of the jquery-ui.css.

cgolubi1:
One question i have about overriding the class definitions for `.ui-widget` - what *else* do we have on the site which is a `ui-widget`?  Asking because we don't explicitly reference that class by name anywhere, so it must be being magically applied on the basis of some other class.

cgolubi1:
`skillSpan.apped(toolTipDiv);` doesn't work because the `skillSpan` is what's shadowshade has set to be clickable, but setting up a parent span and appending both `skillSpan` and `toolTipDiv` to it seems to work.

cgolubi1:
Okay, i don't think we need to futz around with jquery-ui.css at all.  This, in gui.css, seemed to suffice to move the text to the left:
```
.ui-tooltip-content {
    text-align: left;
}
```

My preference would be to put as much as possible of our own styling in our own CSS file, so i think if this works, we should go for it.  As for "how can we tell what's affected?" --- it looks like "tooltip" is the magic word here, so if we have no other tooltips inside the code (which we currently don't), we can override `ui-tooltip-content` to our heart's content without changing existing stuff.

cgolubi1:
Okay, finally, this is the advice i'm looking for, and this is what i want to do.  Let me try it out if i can, but "imitate the browser's native tooltip" is *exactly* what i think we should be doing here - http://jquerytools.github.io/demos/tooltip/imitate.html

cgolubi1:
Okay, so i tried using the styling advice (not the fade advice, because that's for the hover behavior, not th click behavior) from that article whole-cloth, and got:
```diff
diff --git a/src/ui/gui.css b/src/ui/gui.css
index 1441dbe..8021f79 100644
--- a/src/ui/gui.css
+++ b/src/ui/gui.css
@@ -359,6 +359,16 @@ div.hidden_tooltip {
     display: none;
 }
 
+.ui-tooltip-content {
+    text-align: left;
+    background-color:#ffa;
+    border:1px solid #cc9;
+    padding:3px;
+    font-size:13px;
+    -moz-box-shadow: 2px 2px 11px #666;
+    -webkit-box-shadow: 2px 2px 11px #666;
+}
+
 span.info_icon {
     color: white;
     background: #000080;
```

That looks like this:
tooltip_styling

which is not bad - we'd have to figure out how to get rid of the border, but it's obviously much closer.

AdmiralJota:
We _want_ it to look/act like the browser default tooltip? 

> One question i have about overriding the CSS definitions for .ui-widget - what else do we have on the site which is a ui-widget? Asking because we don't explicitly reference that class by name anywhere, so it must be being magically applied on the basis of some other property.

The only other pages that reference jquery-ui.css are buttons.html and history.html, so those are the ones we should test to see if anything looks odd. In particular, I'd check the date picker popups.

But that said, if you don't like Verdana or the larger font there, are you likely to like it anywhere?

cgolubi1:
AdmiralJota: high-level, what *i* want is for this pull to not be a regression in behavior for browser users.  The current behavior for browser users is:
* The skill text works the same way as the die description text, and looks similar to it (which is slightly important because the skill text and the die description text are on the same page)
* The skill text has a basically usable visual look-and-feel (it's left-justified, it's a sensible font given that it's a large amount of text, the box is differentiated from the background, it's easy to make the box disappear again when you're done with it, you don't have to scroll the browser window around to see the box)

All i want is to not lose the browser functionality which exists right now, and which i use.

cgolubi1:
I guess one backing-up-a-step question i have is, how would mobile users feel about being able to click on a die skill and be taken to a new page which described that skill?  Because we certainly could and should implement pages that describe die skills, and we could link to them.

I don't know whether that would accomplish what mobile users want better or worse than having a clickable popup.  Does anyone else know?

irilyth:
> I also like hover as well as click (and said so on #1697, a while ago).

To be more clear about this: The current behavior of tooltip-on-hover works, and is useful, and we shouldn't break it. Right?

cgolubi1:
Okay, so we discussed this on IRC a bit, which got me to go back and
look at what's actually *in* #1697 (which i should have done earlier).
And the only things that issue says in the description are (1) we should
separate die skills from button skills, which we've done, (2) we should
include the letter in die skill descriptions, which is easy.

There's a lengthy discussion in the conversation part of that issue
about possible support for mobile browsers and things which can't hover.
But there's no consensus reached there, and no statement of "this is
what we are planning to implement".

IMO this pull has successfully disproved the claim "we can simply do
something very easy which will add support for mobile users without
impairing the experience of desktop users" --- this pull as proposed
impairs the experience of (at least some) desktop users a lot, and making
it not do that is complicated and fiddly.

So i'd like to request that we *stop* looking for a shortcut solution, and
instead get a working consensus on what we actually want here to support
our various users (browser users, mobile users, users who use big fonts),
and make that statement of what we're trying to implement somewhere.
That way we can invest the time in figuring out how to cleanly do what
we actually want to do.

cgolubi1:
Caching this link for later - http://knackforge.com/blog/karalmax/how-deal-hover-touch-screen-devices - it mentions a possibility which i think is intriguing, that we could have our elements listen for a touchscreen-only event called "touchstart", and thereby implement some sort of small-device-targeted UI thingy only for touchscreens, leaving us free to use the existing hover for desktops.

blackshadowshade:
Referencing #2073.

blackshadowshade:
I'm going to close this pull request for the moment, since it looks like there needs to be quite a bit more discussion on this issue before we can move forwards. Feel free to copy/paste relevant bits into the newly created issues.

Issue #822, opened 2014-05-03
Option dice representation on game page is cluttered


irilyth:
On the game page, in the section where you're selecting dice to attack with (or against), representing Option dice as things like "(1/8=1)" seems to me to be cluttered and confusing. I guess it's consistent with the "(X=17)" phrasing for swing dice, but I'm not all that fond of that either. Why not just put the size there? The fact that it's an Option die generally isn't relevant during the game, and if you need to know, the recipe is right there under the button.

(How did the old site do this? Ah, it spelled out the whole thing like "Option 4/10 (with 10 sides)", hmm.)

irilyth:
I got bitten by this again just now, mistakenly reading a (4/20=20) as a four-sider rather than a 20-sider, because the thing immediately after the opening parens is a 4. I don't have a current proposal for this, but wanted to bump that I'm thinking about it.

(A half-assed proposal: What about something Javascripty that only displayed the current size, including for swing dice, but with a color and/or adjecent symbol (like an asterisk) indicating that there was more going on than a regular die of that size, which you could hover over and/or click on to get more details?)

blackshadowshade:
What about using font size or bolding the appropriate part of the option representation?

irilyth:
I made a quick mockup of bolding, and I think it's better than it is now, but the current display still just seems terrible to me, full of irrelevant and confusing information. Like, the thing where S swing dice look like Shadow dice, and Z swing dice like Speed dice, which I guess is beyond the scope of this ticket; #216 seems to have been somewhat about the more general issue, but that's been closed for ages, so probably no one wants to revisit that.

Here's what bold looks like, anyway.

![bm](https://cloud.githubusercontent.com/assets/4923138/14237205/a6e8f9b0-f9ee-11e5-8a5c-f1d660274e77.png)

blackshadowshade:
Just letting people know that irilyth is contemplating having this for his BotM. I've told him that I'd like for there to be some consensus before I'll spend any time programming this, so pipe up now if you have an opinion about how option dice (and perhaps swing dice) are displayed.

dwvanstone:
Could you send in a screenshot with the confusion between S dice and shadow dice or a Z dice and speed dice?

AdmiralJota:
Is the bolding the only change that you've made so far? If so, it looks fine to me.

[...]

PR #1981, opened 2016-04-18
Refactored code that displays dice

blackshadowshade:
This pull request is in preparation for #822.

The idea of this pull request is to refactor the code so that it becomes simple to change the die backgrounds. Previously, this would have required changes in a monolith function, which would make it difficult to work out if there are side effects of the changes. Now, all changes can be confined to Game.createDieDiv().

There should be no difference in the look and feel of the UI resulting from this refactoring. I *did* make one change, which is to always leave the padding non-breaking spaces around the die value, instead of only having it there when it's needed for the overstrike character (for captured dice). These are effectively invisible to the normal player, and it simplifies the code to leave them in.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1122/ (if it passes)


blackshadowshade:
New Jenkins run: http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1140/ (if it passes)

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1143/ (if it passes)

src/ui/js/Game.js
```diff
@@ -2847,6 +2769,123 @@ Game.gamePlayerDice = function(player, player_active) {
   return allDice;
 };
 
+Game.isDieClickable = function(isPlayerActive, player, die) {
+  // Find out whether this die is clickable: it is if the player
+  // is active and this particular die is not dizzy
```
cgolubi1: Nit: comment doesn't mention warrior

src/ui/js/Game.js
```diff
@@ -2847,6 +2769,123 @@ Game.gamePlayerDice = function(player, player_active) {
...
+};
+
+Game.createBorderDiv = function(color) {
+  return $('<div>', {
+    'class': 'die_border',
+    'style': 'border: 2px solid ' + color,
+  });
+};
+
+Game.createContainerDiv = function(
```
cgolubi1: Nit: since createBorderDiv and createContainerDiv are functions to create borders and containers *for dice*, IMO the function names should contain the word "Die" --- Game.js is a big module with lots of stuff going on, so this could actually reduce confusion.

test/src/ui/js/test_Game.js
```diff
@@ -1702,16 +1864,17 @@ test("test_Game.dieBorderTogglePlayerHandler", function(assert) {
...
     var dieobj = $('#playerIdx_0_dieIdx_0');
     var html = $('<div>').append(dieobj.clone()).remove().html();
-    assert.ok(html.match('die_container die_alive unselected_player'),
+    assert.ok(html.match('die_container die_alive hide_focus unselected_player'),
       "die is unselected before click");
 
     $('#playerIdx_0_dieIdx_0').trigger('click');
     var html = $('<div>').append(dieobj.clone()).remove().html();
-    assert.ok(html.match('die_container die_alive selected'), "die is selected after first click");
+    console.log(html);
```
cgolubi1: Take your debugging statement out.

test/src/ui/js/test_Game.js
```diff
@@ -1702,16 +1864,17 @@ test("test_Game.dieBorderTogglePlayerHandler", function(assert) {
     // and unselected on click
     var dieobj = $('#playerIdx_0_dieIdx_0');
     var html = $('<div>').append(dieobj.clone()).remove().html();
-    assert.ok(html.match('die_container die_alive unselected_player'),
+    assert.ok(html.match('die_container die_alive hide_focus unselected_player'),
```
cgolubi1: Are there any side effects to always specifying hide_focus when we didn't before?  I'll keep an eye out for that in testing.

test/src/ui/js/test_Game.js
```diff
@@ -1548,6 +1548,168 @@ test("test_Game.gamePlayerDice_warrior", function(assert) {
   });
 });
 
+test("test_Game.isDieClickable", function(assert) {
+  var player = '';
```
cgolubi1: Nit: IMO these tests would be more readable if you didn't keep redefining player as a variable, but just used an inline string every time you invoked `isDieClickable()` --- i know that's picky, it's just that it would make it easier to scan the tests at a glance and see the settings.

cgolubi1:
Okay, how am i going to test this?
* Do a close read of both the "before" and "after" code, and convince myself:
    * That the functionality shouldn't have changed
    * That the JS language elements being used are similar to the ones we already use, and not adding new language elements which would need more testing
* Understand the circumstances in which `hide_focus` is now active on unselected dice (while toggling die selection on and off?  All the time, and the border toggle checks are just the only tests which actually check for it?), and make sure the visual behavior is okay
* Make sure the QUnit tests pass in vagrant (i.e. that nothing has been added which depends on ordering that different browsers might do differently)
* Pull up a database from a live site on vagrant, and look at the old and new UI across many games simultaneously
    * Also make sure keyboard-based play seems to still work

The main issue is that i don't trust the UI regression tests to catch minor flag changes which will look visually distracting, because a lot of them are grep-based tests, so if something the test wasn't looking for changed, the test wouldn't see it.  I think side-by-side comparison of before and after sites should be pretty good, but we'll see how that goes.  Anyway, there's some kind of plan.

blackshadowshade:
New Jenkins run: http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1158/ (if it passes)

cgolubi1:
This failed Jenkins's audit (because you renamed some functions but didn't rename the corresponding tests, bleah, sorry).

cgolubi1:
Anyway, nothing in your changes *should* have caused Jenkins to fail for a code-specific reason, so i'm not going to let that stop me from testing before you fix it.

cgolubi1:
Nit: it's not important, but since your goal is to clean up `Game.gamePlayerDice()`, why not order elements consistently within that?  In particular, move `isSelected` up with the other data-gathering variable definitions before you start defining divs, and define the four divs in the same order in the `activeDieArray` and `capturedDieArray` cases.

cgolubi1:
It seems to me that the new `Game.createDieContainerDiv()` is a bit of a mess:

First off, in the `activeDieArray` and `capturedDieArray` cases, you invoke the function with different numbers of arguments (you're missing `isSelected`, the fourth argument of six, in the latter case).  Oddly enough, this seems basically harmless, presumably because the latter arguments aren't actually used in the captured die case, so it doesn't matter that the wrong arguments are being assigned to the wrong variable names.

Second, you don't actually pass the information about whether the die is alive or not, but instead infer it using game logic, so in the `WasJustCaptured` case, you're defining and redefining `dieContainerDiv` in compatible ways.  If somehow the invoking logic changes so that a captured die without `WasJustCaptured` is passed to this function, this function will put that die in the wrong case.

My opinion is that there's no real shared logic between the alive case and the captured case, so you should simply have two functions, `Game.createAliveDieContainerDiv()` and `Game.createCapturedDieContainerDiv()`, and that will get you out of the current trouble.  If you really want them to be one function, you should pass whether the die is alive as a top-level argument and branch on it in the top-level if/else before defining `dieContainerDiv` --- but IMO if you're doing that you might as well just make two functions.

cgolubi1:
So i got partway through close-reading the refactor for what has changed in the logic, and i'll continue doing that tomorrow.

blackshadowshade:
Let's see if I've made the changes correctly enough for Jenkins:

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1160/ (if it passes)

cgolubi1:
Okay, so, let's see if i can explain this regression so it makes sense:
* You modified `dieBorderDiv` so it's populated by the function `Game.createDieBorderDiv()` - so far so good
* However, the previous code did a check for `isSelected` (effectively), and only created a border of the same color as the background color when the die was not selected.  `Game.createDieBorderDiv()` does not contain this check --- it just always creates the background-colored border.
* When you're starting with a fresh page and selecting dice, this is okay --- when you click the dice to select them, the red border appears.
* However, if you submit an attack which is rejected as invalid, your dice should stay selected when the page reloads (via Game.activity.whatnot) --- the regression is that with your new code, the red border around the still-selected dice disappears in this rejected-attack case.

cgolubi1:
Unfortunately, i found this regression by going through your refactor line-by-line and looking at, in each of the possible logical cases (toggling isClickable, isSelected, player, and player_active), what exactly does each of the divs generated by `Game.gamePlayerDice()` look like.  This is a slow process, and i don't think finding these regressions is optional, so i'm going to keep doing it.

blackshadowshade:
Good catch. This should be fixed now.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1163/ (if it passes)

cgolubi1:
Ugh, sorry i didn't look at this sooner --- i think it's kind of hacky to add a style and then take it away again.  I don't know if it will ever do the wrong thing, but it seems wrong.  Why not pass isClickable and isSelected to `Game.createDieBorderDiv()` and have the logic do what you actually want it to do?

cgolubi1:
Anyway, it does seem to fix the bug --- i just don't like the fix.  :>)

src/ui/js/Game.js
```diff
@@ -2789,11 +2783,16 @@ Game.isDieClickable = function(isPlayerActive, player, die) {
...
 
-Game.createDieBorderDiv = function(color) {
-  return $('<div>', {
+Game.createDieBorderDiv = function(color, isSelected) {
+  borderDiv = $('<div>', {
     'class': 'die_border',
-    'style': 'border: 2px solid ' + color,
   });
+
+  if (isSelected) {
```
cgolubi1: Hmm, isn't this backwards?  Shouldn't you be installing the mat-colored background if the die is *not* selected?

cgolubi1:
Another change that's probably harmless, but i'll want to look a bit more closely --- dead dice now have the class "die_greyed", i.e. here's the diff:

```diff
            <div title="This die was just captured in the last attack and is no longer in play." class="die_container die_dead">
             <div>
              <span class="die_recipe_player">
               (4)
              </span>
             </div>
             <div style="border: 2px solid #dd99dd" class="die_border">
-             <div class="die_img">
+             <div class="die_img die_greyed">
               <span class="die_overlay die_number_player">
                &nbsp;1&nbsp;
               </span>
              </div>
             </div>
            </div>
```

That happens to be harmless with the current contents of those classes.  Is it actually correct?  I kind of think not --- i think dead dice are not the same as inactive dice.  But maybe it is.

I'd be happy to have you weigh in on this thread on whether you think it's correct for dead dice to have `die_greyed` (which really means "inactive").

blackshadowshade:
Right, I think I have those changes in place.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1171/ (if it passes)

cgolubi1:
So... i realise the naming is not ideal, but nevertheless, `die_dead` is a property that goes on the die container, not on the die.  It *looks* like it's harmless to apply the same property twice, but it's confusing.

cgolubi1:
At any rate, the logic seems backwards to me in terms of intent, though this is nitpicky --- `die_greyed` is a property which is applied to an alive unclickable die.  So you should pass `isAlive`, not `isDead`, because if we were to add another die state in addition to alive/dead, it would presumably also not use `die_greyed`, since `die_greyed` is a property of alive dice.

blackshadowshade:
Sorry to say this, but all changes to this pull request are having ramifications on the following pull request. Could we leave this logic change until the next pull request so that I don't need to keep dealing with merge conflicts?

blackshadowshade:
(just mentioning this because the database changes in the next pull request are making it difficult for me to go forward and backwards on my MAMP stack)

cgolubi1:
> Sorry to say this, but all changes to this pull request are having ramifications on the following pull request (#2022). Could we leave this logic change until the next pull request so that I don't need to keep dealing with merge conflicts?

So, i've started looking at the follow-on pull request a little bit, and this seems like a worst-of-both-worlds approach to me.  The goal of having a refactor followed by a functionality change is that the refactor should be a noop.  Basically your request is that i not ask the refactor to be a noop, but also that i have to look at the "refactor" pull separately from the new functionality pull, so i can't think about the refactor in light of the new functionality.  Basically i think we have a problem here.

blackshadowshade:
Look, I'm more than willing to work on this pull request to make it a pure refactor, but as we've talked about elsewhere, our major problem is that we don't have the test cases required to prove that it is a pure refactor. Because of this, I thought that it would be enough to have a UI that seems to have the same functionality. The reason for this is because the next pull request tears down parts of the refactored functionality anyway, and these parts are the bits where I don't think it's useful to lock down the current functionality.

I suggested that the two comments you just made be delayed because
(i) they don't appear to change current visible behaviour
(ii) they will be changing anyway

If this isn't good enough, then I'm happy for you to take over this branch and refactor until you're happy with it, and put up a new pull request, with the proviso that I'll may well tearing down parts of your refactor in the follow-up pull request.

blackshadowshade:
Okay, from IRC, we've agreed that Chaos will complete this refactor, with the possibility of additional QUnit tests of key functionality that we want to be able to lock in.

blackshadowshade:
Closing this one in favour of #2050.

Issue #1991, opened 2016-05-13
make the action log smaller in the database


cgolubi1:
Right now, 98% of the disk space consumed by our production database consists of action log entries.  This is not scalable - it will be a bottleneck for us eventually, and it's possible that it is implicated in problems we're seeing now (see #1989).  Come up with a more concise database format for the action log.

cgolubi1:
Here's my plan for this refactor.

Right now, action log entries contain a lot of stuff we don't need, a huge JSON blob of state, because we weren't confident in what info we would or wouldn't need to render action log messages.  That big JSON blob is what i want to get rid of.

As it turns out, the implementation of BMGameAction is currently tidy enough (i think) that we *do* know what we need for each message.  Each `friendly_message_$actionType()` function must contain within its code all the variables from the JSON blob we actually use to render that message type.

With that in mind, i think the end state should be:
* Keep the existing `game_action_log` table (for ordering and all the existing shared variables like timestamp and acting_player), but instead of `message`, it contains an ID to an entry in a per-action-type table
* Each per-action-type table (e.g. `game_action_log_attack`, `game_action_end_winner`) just contains an ID and the set of variables appropriate to that action 

Short-term implementation details:
* I also think the database stuff for game actions should move into BMInterfaceGameAction - right now, the partially-completed refactor kind of has that stuff smeared over BMInterface and BMInterfaceGame, but i'm going to argue that a game action log entry is its own type of object (as you can tell from the fact that it has its own class, BMGameAction), so it should have its own database access class.
* My first step will be to put together the redirect logic but make it optional initially --- that will let me start working on the cutover without solving all action types at once.
* Some slightly complicated scripting will be needed for the actual cutovers of live databases (and i might choose to make some shell or python scripts rather than doing that 100% within SQL), but, again, if i'm right about how to derive the needed fields from the JSON blobs, the cutover scripts shouldn't need to be hackish.

blackshadowshade:
What do we do with messages that contain multiple elements? Do these get decomposed into their elements in the resulting log too?

blackshadowshade:
I would request that #1675 be resolved before this refactor takes place.

cgolubi1:
> I would request that #1675 be resolved before this refactor takes place.

It's a reasonable request, but the only way i can make it happen right now is to merge #1675 completely blind with no further testing.  I'll think for a little bit about whether i want to do that.

cgolubi1:
[...]

PR #1992, opened 2016-05-14
Migrate action_log entries of type end_draw to new database structure

cgolubi1:
* This is the proposed framework for #1991
* Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/434/

In this pull:
* Create BMInterfaceGameActionLog.php and move the handful of existing action_log database functions there
* Add generic logic to action_log load and save methods to use dedicated tables where they exist
* Migrate the `end_draw` action log entry's load/store methods to use such a dedicated table, `game_action_log_type_end_draw`
* Provide an idempotent utility to be run as part of the deployment which migrates all existing `end_draw` entries

### DIFF
--- a/deploy/database/schema.game.sql
+++ b/deploy/database/schema.game.sql
@@ -69,9 +69,16 @@ CREATE TABLE game_action_log (
     action_time        TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
     game_state         TINYINT UNSIGNED DEFAULT 10,
     action_type        VARCHAR(20),
+    type_log_id        INTEGER UNSIGNED DEFAULT NULL,
     acting_player      SMALLINT UNSIGNED NOT NULL,
     message            TEXT,
     INDEX (game_id)
+);
+
+CREATE TABLE game_action_log_type_end_draw (
+    id                 INTEGER UNSIGNED AUTO_INCREMENT PRIMARY KEY,
+    round_number       TINYINT UNSIGNED NOT NULL,
+    round_score        VARCHAR(10) NOT NULL
 );
 
 CREATE TABLE game_chat_log (
--- a/deploy/database/updates/01991_action_log.sql
+++ b/deploy/database/updates/01991_action_log.sql
@@ -1,0 +1,7 @@
+ALTER TABLE game_action_log ADD type_log_id INTEGER UNSIGNED DEFAULT NULL AFTER action_type;
+
+CREATE TABLE game_action_log_type_end_draw (
+    id                 INTEGER UNSIGNED AUTO_INCREMENT PRIMARY KEY,
+    round_number       TINYINT UNSIGNED NOT NULL,
+    round_score        VARCHAR(10) NOT NULL
+);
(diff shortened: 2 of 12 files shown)


cgolubi1:
How i tested this pull:
* Update-vs-rebuild test of database sql files
* On a fresh site, play an Envy vs. Dead Guy game and verify that the end_draw action log entry is stored and loaded correctly
* Load a recent backup of the prod site database onto virtualbox, update it using the test script, and spot-check that games with draws can still be loaded and display their action logs.

blackshadowshade:
So, looking at this pull request, I note that we're requiring the action_type string to match the table name suffix. I wonder if we should eventually (i.e., not affecting this pull request) change action_type to an enumeration. What do you think?

cgolubi1:
> So, looking at this pull request, I note that we're requiring the action_type string to match the table name suffix. I wonder if we should eventually (i.e., not affecting this pull request) change action_type to an enumeration. What do you think?

Yeah, i thought about that, and i do think we eventually should.  I just didn't want to juggle too many things at once, and there are a couple of stale action_types still floating around in the prod (and presumably dev) databases, so i think it would make more sense to get the data tidy first, then move over to ENUMs.

blackshadowshade:
Excellent, we'll keep that on the horizon then.

blackshadowshade:
So, I've tested on my MAMP stack and everything works in a Dead Guy vs Echo matchup.

There's one thing that I'm wondering about, namely your selection of a VARCHAR(10) to store the round_score, instead of a float representation. I can imagine a number of reasons that you might have done that, but I'd like to hear what your reasoning is.

blackshadowshade:
Actually, there is one other point.

You're querying the last insertion ID to retrieve the log ID. However, if another insert happens to occur in the meanwhile, you're going to retrieve the wrong ID. Doesn't this actually _require_ a transaction explicitly to stop bad stuff happening?

cgolubi1:
> You're querying the last insertion ID to retrieve the log ID. However, if another insert happens to occur in the meanwhile, you're going to retrieve the wrong ID. Doesn't this actually require a transaction explicitly to stop bad stuff happening?

Hmm.  Well, this isn't the only place we do it --- i would have flagged it for discussion if it had been.  It happens in `insert_new_game()` and it happens in `insert_forum_thread()`.  Maybe it's wrong in all of those places, and we should define a database helper function which does an insert and returns the ID, wrapped in a transaction?

blackshadowshade:
Okay, answering myself.

According to http://stackoverflow.com/questions/10680943/pdo-get-the-last-id-inserted, a transaction shouldn't be needed, since last_insert_id is connection specific.

cgolubi1:
Yep, i just found http://stackoverflow.com/questions/14827783/auto-increment-and-last-insert-id which says the same thing.

cgolubi1:
> There's one thing that I'm wondering about, namely your selection of a VARCHAR(10) to store the round_score, instead of a float representation. I can imagine a number of reasons that you might have done that, but I'd like to hear what your reasoning is.

So, primarily expedience --- i didn't know what state the current data in the database would be in, and i wanted to make sure not to destroy any data we'd need during the migration.  Probably we should eventually use a float, but deferring that to a later migration as needed, means i don't have to worry about what will happen if something which happens to be in the DB right now has a format which the float i pick doesn't like.

blackshadowshade:
Okay, fine, as long as we revisit this decision later and review whether it's still appropriate.

PR #1993, opened 2016-05-15
Migrate end_winner action_log entry parameters to a separate database

cgolubi1:
* Partially addresses #1991 
* Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/435/

Worth noting:
* I realised while testing this that there are old end_winner and end_draw entries in the dev database which are still in string form rather than JSON array form.  There aren't any in the prod database.  So i modified the script for end_winner to be able to handle either.  I'll go back and do the same for the end_draw script as a separate small pull, since that blocks pushing the end_draw change onto dev
* I chose to rename "resultForced" to "surrendered" within the action log sphere, because the code treats it as meaning surrendered (whenever it's set, the message printed is "surrendered")
* I chose to change the 'true' value of resultForced/surrendered from an array to a boolean.  The false value is a boolean, and, while this nominally loses information about which player surrendered: (1) we weren't using that information in any way, and (2) the way the game is implemented right now is that the player who surrendered is the one who didn't win the round.  So in practice we lose no information, and it's easier/more compact to deal with a boolean FALSE/TRUE than a boolean FALSE/array().

### DIFF
--- a/deploy/database/schema.game.sql
+++ b/deploy/database/schema.game.sql
@@ -81,6 +81,14 @@ CREATE TABLE game_action_log_type_end_draw (
     round_score        VARCHAR(10) NOT NULL
 );
 
+CREATE TABLE game_action_log_type_end_winner (
+    id                  INTEGER UNSIGNED AUTO_INCREMENT PRIMARY KEY,
+    round_number        TINYINT UNSIGNED NOT NULL,
+    winning_round_score VARCHAR(10) NOT NULL,
+    losing_round_score  VARCHAR(10) NOT NULL,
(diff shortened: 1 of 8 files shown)


cgolubi1:
Testing done:
* update vs. rebuild DB test
* load recent prod and dev databases, run the update script, and verify that both recent games and very old string-formatted games from dev are viewable

cgolubi1:
Cancelling for now --- i realised that it would make more sense for some of the action log types to have the new table index back to action_log, rather than adding a new index.  So i'm going to do that for end_draw before trying to get any of the others approved.

PR #1994, opened 2016-05-15
Fix end_draw to work with grandfathered string entries from very early on the dev site

cgolubi1:
* Related to #1991 
* Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/436/ (should pass, i haven't changed anything Jenkins tracks)

Tested by loading databases from each of the dev and prod sites, and seeing that they both could still be migrated, and that early dev games containing draws still looked right.

### DIFF
--- a/deploy/database/updates/scripts/01991_migrate_end_draw_action_logs
+++ b/deploy/database/updates/scripts/01991_migrate_end_draw_action_logs
@@ -6,7 +6,7 @@ import json
 import re
 import MySQLdb
 
-END_DRAW_OLD_STRING_RE = re.compile('^Round (\d+) ended in a draw \((-?[0-9\.]+) vs\. (-?[0-9\.]+)\)$')
+END_DRAW_OLD_STRING_RE = re.compile('^Round (\d+) ended in a draw \((-?\d+(?:\.\d+)?) vs\. (-?\d+(?:\.\d+)?)\)$')
 
 def get_last_insert_id(crs):
   crs.execute('SELECT LAST_INSERT_ID()')
@@ -35,7 +35,7 @@ def migrate_to_type_log_end_draw(row, crs):
     raise ValueError, "Got unexpected return %s from %s" % (result, insert_sql)
   type_log_id = get_last_insert_id(crs)
   update_sql = 'UPDATE game_action_log SET type_log_id=%s,message=NULL WHERE id=%d' % (type_log_id, row_id)
-  result == crs.execute(update_sql)
+  result = crs.execute(update_sql)
   if not result == 1:
     raise ValueError, "Got unexpected return %s from %s" % (result, update_sql)
   print "Moved %s to game_action_log_type_end_draw id %s" % (row[1], type_log_id)
--- a/deploy/database/updates/scripts/01991_migrate_end_draw_action_logs
+++ b/deploy/database/updates/scripts/01991_migrate_end_draw_action_logs
@@ -3,7 +3,10 @@
 # Utility to migrate end_draw action log entries to new format
 
 import json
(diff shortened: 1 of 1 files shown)


blackshadowshade:
I know that we have pretty well behaved log strings in the database, but I think that we should be a little more careful with our regexp matching of the scores. You currently have

    (-?[0-9\.]+)

and I think this should actually be

    (-?\d+(?:\.\d+)?)

to enforce only a single optional decimal point.

cgolubi1:
Hmm, `(-?\d+(?:\.\d+)?)` seems a little illegible to me.

Wouldn't: `(-?\d+(\.\d+)?)` do what you want?

blackshadowshade:
It would do it, but I put it that way so that we can be consistent in the future with later pulls, when you'll need to retrieve both scores. This way, it becomes clearer exactly how the groups will be numbered.

cgolubi1:
> It would do it, but I put it that way so that we can be consistent in the future with later pulls, when you'll need to retrieve both scores. This way, it becomes clearer exactly how the groups will be numbered.

Wait, what?  I don't understand this at all.  :>(

cgolubi1:
Since this is important to you, i think it's on you to give some examples of what `(-?\d+(?:\.\d+)?)` matches, rather than being incumbent on me to reverse-engineer the clever thing you've figured out.  :>)

blackshadowshade:
The ?: makes the group a non-capturing group, meaning that when we extract the group contents via mobj.group(#), the inner group cannot be extracted separately. Thus, for your regexp, there are only three groups, just like your original regexp.

If we left out the ?:, there would be five groups, and I wouldn't be sure (personally) which of the five groups are the ones that need to be extracted.

blackshadowshade:
(altered previous post to make it clearer, going offline for the moment, continue later)

cgolubi1:
I guess i'd also like to make the point that i *don't* think the accuracy of these regexp translation scripts is very important.  The reason is this: 100% of the entries in the prod database have JSON-formatted messages.  I'm double-checking this now, so i'm pretty sure it's true.

In other words, any regexp matching i do, is only for the benefit of old games in the *dev* database.  So i would argue that it isn't very important.  (And i guess to that end, if you think it's important to switch to the new format, i'll shut up and switch to the new format and test it.  But i did want to make the argument for why i don't think this is very important, which is that it doesn't affect prod data.)

cgolubi1:
Anyway, thanks for the explanation of '?:'.  I'm trying this out now with your syntax.

cgolubi1:
http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/438/ - no reason this shouldn't pass, since Jenkins doesn't test the script.  I tested it, and it still seems to work.

blackshadowshade:
Just for information, the reason that I'm concerned with the accurate regexp is because the second round score may possibly be followed by a full stop. This mostly doesn't happen now, but I remember that historically, we've not always had the full stops correctly excluded from the log.

As I said, I'm not too worried about the current pull request, but I'd like us to be consistent across the multiple pull requests where we need to migrate the log.

cgolubi1:
So, there are no action logs, of any type, on the prod site, which are still in string format.  So we will never be using these regexps against prod logs.  We will ONLY be using these regexps against very old dev site games.  When i say this isn't important, what i'm saying is that i don't think keeping early dev site games running is very important.

I can't tell whether you (a) disagree about the importance of keeping early dev site games running, or (b) believe we will be using these regexps to parse prod site data later in this project.  If it's (b), i guess i'd like to understand why, in case it's me who's missing something.

PR #1995, opened 2016-05-16
Remove extraneous parameter for decline_auxiliary

cgolubi1:
This is the example pattern for action_types which don't actually need to store any parameters.  The only reason decline_auxiliary had one was because of the check in BMGameAction.  That check isn't needed, i'm pretty sure it was just inserted to protect against trouble during the long-ago string-to-JSON conversion.

Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/437/


cgolubi1:
Cancelling for now --- this is probably the right approach for this case, but i'm going to get rid of the new column per discussion with shadowshade.

PR #1996, opened 2016-05-16
Change the database format for end_draw to put the index in the secondary table

cgolubi1:
* Related to #1991 
* Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/439/

I apologise for all the bother --- i think it makes much more sense to do things this way, and the fact that it simplifies the database update (no modify to game_action_log needed), saves space while we're in transition, and makes it easier to handle the case where no auxiliary table is needed (just have a secondary function with the usual name which returns empty params, rather than hardcoding an exception), makes me confident that this change is the right thing.


blackshadowshade:
Okay, I've tested this on my WAMP stack with Lucky vs Crusher (which according to game 950 on the live site is fully capable of producing lots of draws, as long as Lucky sets X=4 and Crusher sets X=12).

The logging code seems to work fine, and the database entries of the draw seem to be reasonable.

PR #2004, opened 2016-05-22
Script to fix twin dice in live databases

### LINKED ISSUE #2003: game 1806 on the prod site is unloadable
I get an internal error when i try to load recently-once-again-active game 1806.

It's another bug involving stale games with twin dice that don't have their twin flags set correctly.

cgolubi1:
* Fixes #2003 
* Not bothering with jenkins because there's nothing testable here

This script reports on three die cases and tries to fix one of them:
* die->value is set and die is not a swing die, but there are no twin flags - the script fixes this case by setting the twin flags to something reasonable
* die->value is set and die is a swing die, but there are no twin flags - the script does not fix this case, and it seems from experimentation like this case is somewhat self-repairing as the game progresses
* die->value is not set, and twin flags have sides, but no values - this could be a sign of a game that is already unloadable, or it could represent some other case, so the remedy is just to check by hand and see if the game is loadable

This script will not repair game 1806 on prod, but it should report it, and it should repair other old games which are not yet unloadable, but might become unloadable.

### DIFF
--- a/deploy/database/updates/scripts/02003_fix_all_twin_dice
+++ b/deploy/database/updates/scripts/02003_fix_all_twin_dice
@@ -1,0 +1,56 @@
+#!/usr/bin/python
+#####
+# Utility to fix all twin dice in prod and dev databases
+
+import json
+import MySQLdb
+
+def parse_twin_info_from_recipe(recipe):
+  sides_arr = recipe.split('(')[1].split(')')[0].split(',')
+  assert (len(sides_arr) == 2)
+  if sides_arr[0].isdigit() and sides_arr[1].isdigit():
+    l_sides = int(sides_arr[0])
+    r_sides = int(sides_arr[1])
+    is_swing = False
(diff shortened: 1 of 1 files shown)


cgolubi1:
Verified that the script does report a problem with game 1806 give the DB's current state:
```
...
This game may be unloadable right now, and the row may need to be fixed by hand - if sides is non-null but value is null, that's a bad sign: (4326995L, 1806, 1, 'p(6,6)', None, 12, 'Twin__{"sides":[6,6],"values":[null,null]}')
This game may be unloadable right now, and the row may need to be fixed by hand - if sides is non-null but value is null, that's a bad sign: (4326997L, 1806, 1, 's(8,8)', None, 16, 'Twin__{"sides":[8,8],"values":[null,null]}')
...
```

blackshadowshade:
Looks okay on a readthrough.

PR #2005, opened 2016-05-22
Exclude common ignorable errors from apache log monitor so it is usable

cgolubi1:
Goal here is just to make the apache log monitor a bit more usable - i've been finding that i fall pretty far behind on the mail it sends, which means i'm not using it as a straightforward way to detect internal errors in real-time.


cgolubi1:
Testing done: generated an internal error, and verified that it is still reported:
```bash
chaos@sandbox:~$ sudo /usr/local/sbin/monitor_apache_logs
[Sun May 22 17:12:52 2016] [error] [client 10.0.2.2] Caught exception in BMInterface::load_game: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'v.original_recipe' in 'field list', referer: http://localhost:8080/ui/game.html?game=1
chaos@sandbox:~$ 
```

blackshadowshade:
I'm happy to accept this.

PR #2021, opened 2016-05-27
Move turndown_focus action_log secondary information to dedicated tables

cgolubi1:
* Partially addresses #1991 
* Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/449/ (if it passes)

Note that this is the first example of a case in which there might be more than one secondary table entry per primary log entry.  IMO it's pretty clean, but obviously i'll be doing a bunch more of this when i get to stuff like specify_dice and attack.  Worth noting is that i named the secondary table `...turndown_focus_die` rather than `...turndown_focus` as a marker that it's a per-turndown-focus-entry-die table rather than a per-turndown-focus-entry table (and when i get to specify_dice and attack there may be more than one secondary table, each storing a different type of secondary data).

Also worth noting is that the savings here are pretty big per-entry (though of course there are a small number of entries) --- the new secondary data is 6% of the size of the old secondary data.  Turns out that storing a bunch of die information we don't need in JSON with big string prefixes is expensive.

### DIFF
--- a/deploy/database/schema.game.sql
+++ b/deploy/database/schema.game.sql
@@ -101,6 +101,15 @@ CREATE TABLE game_action_log_type_add_auxiliary (
     INDEX (action_log_id)
 );
 
+CREATE TABLE game_action_log_type_turndown_focus_die (
+    id                 INTEGER UNSIGNED AUTO_INCREMENT PRIMARY KEY,
+    action_log_id      INTEGER UNSIGNED NOT NULL,
+    recipe             VARCHAR(20) NOT NULL,
+    orig_value         SMALLINT,
+    turndown_value     SMALLINT,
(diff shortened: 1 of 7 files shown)


cgolubi1:
In addition to the usual testing, i also verified that:
```sql
mysql> select * from game_action_log_type_turndown_focus_die where orig_value=turndown_value;
Empty set (0.00 sec)

mysql>
```

for each of the dev and prod databases - in other words, there are no old entries which contain information about extraneous dice which were not turned down.

blackshadowshade:
Okay, this works fine on my MAMP stack. Tested with single turndown, double turndown, alternating turndowns. Jenkins looks good, and I'm encouraged that there are no changes to the responder tests.

PR #2022, opened 2016-05-28
Added user preference for die backgrounds

### LINKED ISSUE #822: Option dice representation on game page is cluttered

blackshadowshade:
Fixes #822. Depends on #1981.

Tested with a standard Soldiers matchup, and changed the user preference multiple times, then logged in as the opponent to see that their view is unchanged.

Known issue: twin dice are not dealt with specifically, they are displayed as their combined maximum

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1149/ (if it passes)


blackshadowshade:
So, I've just merged to keep this up-to-date with the changes happening on #1981.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1165/ (if it passes)

blackshadowshade:
Next merge to keep things up to date.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1172/ (if it passes)

blackshadowshade:
From #1981:

> So... i realise the naming is not ideal, but nevertheless, die_dead is a property that goes on the die container, not on the die. It looks like it's harmless to apply the same property twice, but it's confusing (i.e. it's confusing to future readers of the code).

In this pull request, I currently rely on being able to deduce whether the die is dead by looking at the die_dead class, which I add to the die itself. Are you saying that I should instead be looking at whether the die container is dead?

blackshadowshade:
So, from Chaos on IRC, we'll change the current container classes die_alive and die_dead to die_container_alive and die_container_dead.

This then allows me to use die_dead to refer to the contained die representation, which I then use to select the die image.

blackshadowshade:
And finally, a passing Jenkins run!

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1174/

blackshadowshade:
So, I'll be redoing this pull request from scratch to work with the new refactor in #2050. The addition of the new die images will be split out from this into its own pull request.

cgolubi1:
One note for your redo here is: i think the various attribute classes (`die_dead` and whatnot) are outputs not inputs from the POV of the code --- where possible, they are used to affect browser behavior, not as inputs to the code about logic to use in assembling the page in the first place.  That's because the code already has the logic it used to needs to figure out the state of a die, from the API data, and doesn't need to rederive it from the page.

In other words, IMO you should be able to use dieClickableInfo and dieStatus to figure out which image to display for each die, you shouldn't need the CSS properties.

blackshadowshade:
Closing this one, replaced by #2055.

PR #2024, opened 2016-05-28
Added images from old site, part 1 (a to i)

blackshadowshade:
This is currently a working pull request, which I've opened so that people can comment on what they think about the replacements. I've broken up the general pull request into multiple smaller ones, so that Github will show the image diffs.

(see also https://help.github.com/articles/what-are-the-limits-for-viewing-content-and-diffs-in-my-repository/)

One question about Fanatics that have a generic and a special image: do we want to keep the generic image or not?

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1153/


blackshadowshade:
To my eye, all of these look like improvements.

AdmiralJota:
These all look good to me.  And if we have the real Fanatic button image that was submitted by the person whose button it is, I don't see any need to keep the generic placeholder we created (except in the git history, obviously).

cgolubi1:
The only one of these i'm not totally convinced is an improvement is Bill --- i kind of like the brighter/darker background on the older image.  I'm happy to go with popular opinion either way, but wanted to raise the question.

Hurtay:
I agree: the colours of the old image were more brilliant but in the scan of the real buttons the colours are as dull as in the image uploaded:
dv
If the colours are more intense than here someone probably changed them digitally.

cgolubi1:
Hmm, fair enough.  I'll see how the whole set looks together when i test.

AdmiralJota:
I noticed that too, but I think the new image is more consistent with the other new images from the set, and the others did seem to be improvements. So even if that one isn't, I think that the old version would look out of place next to the rest.

cgolubi1:
Shadowshade: this still has the "question" label on it.  Do you need more feedback before it can be merged, or can we go ahead?

Anyway, i'll go ahead and test, by starting games with each of the new images and looking for any weirdness.

cgolubi1:
Hmm, so ABCGi's twin swing is `(X,X)!` in our database, but `(X,Y)!` on the art.  Thoughts?

cgolubi1:
Also CestWhat's recipe has `p(2)`, whereas the art shows two skill letters - i believe it's `pk(2)`

cgolubi1:
I don't see any other oddities.

irilyth:
> Hmm, so ABCGi's twin swing is `(X,X)!` in our database, but `(X,Y)!` on the art.  Thoughts?

Wow! It was definitely (X,X)! on the old site. I wonder if the old site couldn't support (X,Y)! so it was implemented as something close, or what.

> Also CestWhat's recipe has p(2), whereas the art shows two skill letters - i believe it's pk(2)

I don't remember CestWhat well enough from the old site to be 100% sure, but I don't recall it having a Konstant die.

AdmiralJota:
Where did we get the recipes from? If we're pretty sure they match what was on the old site, I'd keep them the way they are. It's possible that the art and the intended recipe might just not have matched (e.g., if they revised their recipe after playtesting but didn't update the artwork to match).

blackshadowshade:
The question label was left there to remind me that we still didn't have your input, nothing more.  :)

cgolubi1:
There are several iterations of button data from the old site in `notes/database/`, and they all show CestWhat as having a `p2`.  So i think the art is just wrong, but it's canonical wrong art, so, hey.

I'm going to go ahead and merge this now, having not heard any reason not to.

PR #2025, opened 2016-05-28
Added images from old site, part 2 (j to sc)

blackshadowshade:
Second part of the image pull request, open for comments.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1154/


blackshadowshade:
@Hurtay has already noted that Maya has no recipe.

blackshadowshade:
To my eye, all these look like improvements.

AdmiralJota:
Something about the Ronin image doesn't look quite right to me, but that might just be me.

Hurtay:
> Something about the Ronin image doesn't look quite right to me, but that might just be me.

All L5R buttons are very extremely coloured I was told. When I got the scan of the real buttons I asked back because I thought the colours a bit overdone. But that's what the real buttons look like.

AdmiralJota:
Then I guess it's just authentic!

cgolubi1:
I don't see any issues in the code.  I'll bring this up and see how games look with each button.

cgolubi1:
One note/request about these pulls - it's not a huge amount of work to test them regardless, but i think it would make things a little easier if you split things by buttonset, rather than strictly alphabetically.  That is, if you'd put all the new LotFR buttons in the same pull as each other, it would be easier to scan them for set coherency.  And if a pull containing fanatic images contained only fanatic images, it'd be faster to create games.

Again, no need to recreate these pulls, but if you wind up doing a similar thing again for a future pull, please group by set as the high-order bit.

cgolubi1:
Anyway, i don't see any issues with these.  Merging.

PR #2028, opened 2016-05-31
Strip a test hook out of reroll_chance because we can just use the RNG override

cgolubi1:
* Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/450/

Obviously, there's no reason at all to merge this until after the merge freeze - i just noticed the issue and figured i'd submit the fix while i was thinking about it.


blackshadowshade:
This looks fine, and Jenkins passed, so I'm happy with this. We'll merge this once the merge freeze is over.

Issue #1626, opened 2015-05-25
Size changes not logged when Chance dice are re-rolled for initiative


irilyth:
In the first round of http://dev.buttonweavers.com/ui/game.html?game=1669, jimmosk re-rolled a Weak Chance die, and gained initiative:

> jimmosk rerolled a chance die and gained initiative: ch(6) rerolled 4 => 1

We didn't think about this at the time, but wondered later if the die shrank before re-rolling. The log doesn't say, and probably should. (And if it doesn't shrink, it probably should.)

irilyth:
Aha, it came up again on the third round, and it did shrink, it's just not logged.

cgolubi1:
Snagged this --- i'm planning to fix it as part of the action_log refactor work, per previous discussion with shadowshade.

PR #2033, opened 2016-06-07
Migrate "reroll_chance" action log entries to new database format

cgolubi1:
* Partially addresses #1991 
* Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/451/ (if it passes)

I think this is a pretty standard action log migration pull with no unusual design decisions.


blackshadowshade:
This is apparently conflicted.

cgolubi1:
Ah, whoops --- of course removing the hardcoded test value stuff i noticed while creating this pull, conflicted with this pull.

Simple enough.  Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/452/ (if it passes)

blackshadowshade:
So, I have one reservation with this change. In RandomBM* buttons, it's possible to get something like a ch or cH die, where the die changes recipe on the chance action. We're apparently not logging such changes at the moment, but the information is there, and with the current pull request, we lose the recipe after reroll.

blackshadowshade:
Linking #1626 for reference.

cgolubi1:
Makes sense.  Updated everything to keep the post-reroll recipe (but not use it yet).

Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/453/ (if it passes)

blackshadowshade:
I've tested this with normal chance buttons (John Kovalic vs FuzzFace) and confirmed that both winning and losing initiative with chance rerolls works fine.

I've also tested this with a cH die via RandomBMPentaskill, and the recipe changes correctly in the database.

PR #2037, opened 2016-06-09
Remove extraneous parameter from init_decline action log messages

cgolubi1:
* Partially addresses #1991 
* Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/454/ (if it passes)

This one should be pretty boring - unless i mucked up the merge and it fails tests, i doubt there's anything to see here.


blackshadowshade:
I've tested this with Ulthar vs Crab, and the two actions of declining chance and focus log correctly. I'll accept this once Jenkins passes.

PR #2038, opened 2016-06-10
Remove unnecessary action_log parameter for decline_reserve

cgolubi1:
* Partially addresses #1991 
* Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/455/ (if it passes)

This is another very boring change which just removes a noop message parameter.


blackshadowshade:
This works fine on my MAMP stack. Tested using Jedite vs Malachite, both reserve decline and reserve accept work for both players, and the logging is correct. The database entries also look fine (i.e., empty).

PR #2046, opened 2016-06-26
add unit tests for UI gamePlayerDice() method

cgolubi1:
* Related to #822 
* Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/460/

This doesn't refactor anything yet --- it just significantly fleshes out the unit tests of `Game.gamePlayerDice()`, in particular:
* Using a test method i wrote some time ago but haven't used much yet, fully regression-test the set of DOM properties for some of the dice
* Add testing for dizzy dice (this is just a minor variant on the existing warrior test, of course, but it helps test coverage)
* Add testing for the case in which `Game.activity.dieSelectStatus` is set before loading the dice, to verify that dice will have the correct classes to be outlined in red in that case

### DIFF
--- a/test/src/api/responderTest.php
+++ b/test/src/api/responderTest.php
@@ -8413,6 +8413,9 @@ class responderTest extends PHPUnit_Framework_TestCase {
         $retval = $this->verify_api_loadGameData($expData, $gameId, 10);
     }
 
+    /**
+     * @depends test_request_savePlayerInfo
+     */
     public function test_interface_game_019() {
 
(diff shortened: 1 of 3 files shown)


cgolubi1:
Needless to say, this unit test change is in preparation for a refactor --- the point of adding the tests first is so that when i do the refactor, we can see what it changes in detail.

blackshadowshade:
Okay, the PHPUnit tests pass on my WAMP stack, and the QUnit tests pass under both Chrome and Firefox.

PR #2048, opened 2016-06-28
Added turbo

blackshadowshade:
Addresses much of #91.

This pull request implements the guts of turbo swing/option and an initial suggestion for the turbo UI. Logging is currently not present, since this will be added by Chaos.

I've tested with Zeppo vs Mister Peach, and Paris vs bobby 5150, and it looks like basic turbo works fine. No explicit testing has been done for interactions with other skills.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1181/


src/engine/BMDieTwin.php
```diff
@@ -165,7 +165,7 @@ protected function skillStr() {
         $skillStr = '';
         if (count($this->skillList) > 0) {
             foreach (array_keys($this->skillList) as $skill) {
-                if (('Mood' != $skill) && 'Mad' != $skill) {
+                if (('Mood' != $skill) && 'Mad' != $skill && 'Turbo' != $skill) {
```
cgolubi1: Drive-by nit: balance these parens (i.e. remove them from the Mood comparison if you don't want them on the others).

test/src/ui/js/test_Game.js
```diff
@@ -1442,18 +1441,54 @@ test("test_Game.pageAddDieBattleTable", function(assert) {
...
   BMTestUtils.GameType = 'washu_hooloovoo_cant_win';
   Game.getCurrentGame(function() {
     var statusJQuery = Game.gamePlayerStatus('player', false, true);
     var statusPropArr = BMTestUtils.DOMNodePropArray(statusJQuery[0]);
-    var expectedPropArr = [
-      "DIV", { "class": "status_player" }, [
-        [ "DIV", {}, [ "W/L/T: 1/1/0 (3)", "foobar", [ "B", {}, [ "Score: 13 (-30.7 sides)" ] ] ] ],
-        [ "DIV", {}, [ [ "SPAN", {}, [ "Dice captured: q(Z=4)" ] ] ] ],
-      ]
-    ];
```
cgolubi1: What's up with this change?  Bad merge from upstream?

blackshadowshade: I seem to remember that NetBeans was telling me this variable is unused. Was there supposed to be a test involving expectedPropArr?

src/engine/BMDieTwin.php
blackshadowshade: Ah, sure. They should all be bracketed.

test/src/ui/js/test_Game.js
cgolubi1: Hmm, okay.  In general i'd appreciate this kind of one-off tidying being done in separate pulls, if we can swing that.

cgolubi1:
Right now i'm just doing miscellaneous testing and putting in feedback on things i notice.

Here's one: the error message when you fail to fill in a turbo value is quite unfriendly.  This error is the API's "hey, API client, you are trying to break me by giving me bad data, i refuse to talk to you" value --- nothing a player can do via the UI should lead to them getting errors in this format.  In this case, the UI should notice that no Turbo value was provided, and print its own error without even trying to talk to the backend:

error_message_is_unfriendly

test/src/ui/js/test_Game.js
blackshadowshade: Sure, I'll see what I can do. That begs the question, though, of whether there is something I missed while merging. I'm also happy to leave the description of expectedPropArr just as information (as a comment) if it's likely to be useful for you in the future.

cgolubi1:
Visually, i really dislike the always-present-when-it's-your-turn context-free vertical-space-consuming swing ranges.  My suggestion would be to have the X range appear in the left portion of the turbo swing adjustment box which pops up, paralleling the format of between-rounds swing die settings in which the valid range appears to the right of the box.  Anyway, this is what the current interface looks like when you're *not* manipulating turbo swing dice, and i don't like that look:

x_ranges

blackshadowshade:
I'm pretty sure that the information on swing range can actually be important when deciding on what attack you might want to perform.

blackshadowshade:
This is especially true when the opponent also has a turbo swing die.

AdmiralJota:
So we're going with the two-stage turbo adjustment, where you make your move, then get prompted afterward to resize your turbo die?

In any case, I think that this issue could be solved if we included swings in the list of die skills down below.

blackshadowshade:
No, the turbo box appears as you select the turbo die.

AdmiralJota:
Ah, nice!

AdmiralJota:
(Also, for an example of passing the swing info and displaying it alongside skills, there's the Buttons page, although I think that handles things slightly differently.)

cgolubi1:
> I'm pretty sure that the information on swing range can actually be important when deciding on what attack you might want to perform.

Sure, but if you select one of the turbo dice, the box pops up, and you can see the information then.

blackshadowshade:
The issue is when you're trying to decide if you're going to capture an opponent's turbo die, and when you have no turbo dice yourself. Do we then not show the swing ranges there?

blackshadowshade:
Okay, dealt with the nits. I've also added logic to consistently show swing ranges even if only the opponent has turbo dice---which will still be necessary, even if we change how we display the swing range information from how I've currently coded it.

cgolubi1:
Okay, here's a suggestion about swing die ranges.  At the moment, we have Die Skills and Button Skills which appear in rows at the bottom of the game mat.  What if we add a row for Swing Dice to that table, and just have the swing die ranges appear there, formatted the same way the die skills and button skills are, literally 100% of the time in a game with swing dice?  I think that would satisfy your desire for the ranges to be visible more of the time, and it would satisfy my desire to stop adding ugly one-off UI elements.

What do you think?  If you like the idea, let's make a separate issue for that, and then figure out what to do here as a result.

AdmiralJota:
I generally agree, but I would suggest grouping them with the die skills rather than making an extra row for them, since they are basically just extra things that are special about a given die. E.g.:

bm1

cgolubi1:
Jota: i don't feel strongly about whether it's the same row or a different row --- my guess is that shadowshade will prefer the swing range to be literally always visible on the page, so it should be part of the page text, not a mouseover.

Mostly, i would like to veto the approach taken in https://github.com/buttonmen-dev/buttonmen/pull/2048#issuecomment-229831083 which i feel looks terrible.

cgolubi1:
Shadowshade: another thing --- why did you merge the jquery_sourcemap branch into this one?  Is it needed for this pull, or was that an accident?

cgolubi1:
Hmm, okay, expanding my picture of the API messiness that needs to be cleaned up here:
* If the API client calls `submitTurn` with the argument `turboVals[N]` with an empty value, you get "Argument (turboVals) to function submitTurn is invalid", and the game remains in `START_TURN` (i mentioned this above)
* If the API client does calls `submitTurn` with no `turboVals[N]` value at all (when one is needed), the backend cheerfully advances from `START_TURN` to `CHOOSE_TURBO_SWING`... and then stays there.  However, there's no visible API function you can invoke from `CHOOSE_TURBO_SWING`.

Okay, looking into BMInterfaceGame.php, i see what's happening here.  You're actually (correctly) attempting to handle the case in which something goes wrong during `CHOOSE_TURBO_SWING`, by bailing out without saving the game if `set_turbo_sizes()` fails.  However, `set_turbo_sizes()` is implemented wrong --- it starts with:
```php
            if (empty($turboSizeArray)) {
                return TRUE;
            }
```
That case shouldn't be a success --- it should be a fatal error, in which the backend returns a friendly "You attacked with a turbo swing and did not specify any turbo die values" message and remains in `SUBMIT_TURN`.  (Incidentally, you have exactly the same check in `BMGame::set_turbo_sizes()`, and it's wrong there too --- you're never going to be in those functions if you don't need to be.)

So basically that's two fixes needed:
* The UI needs to catch the user's failure to specify needed turbo values, and error before invoking the backend (see any of `Game.formSpecifyDiceActive()`, `Game.formChooseReserveDiceActive()`, `Game.formReactToInitiativeActive()`, `Game.formAdjustFireDiceActive()`, for examples of the kind of error handling that would be appropriate to add to submitTurn to handle this case).
* The backend needs to handle the case in which the user attacks with a turbo die and does not submit values for `turboVals`, by returning a friendly error message and remaining in `START_TURN`.

src/engine/BMGame.php
```diff
@@ -1792,14 +1796,66 @@ protected function update_game_state_commit_attack() {
...
+                return FALSE;
+            }
+        }
+
+        // set turbo sizes
+        foreach ($turboSizeArray as $dieIdx => $dieSize) {
+            $die = $this->playerArray[$playerIdx]->activeDieArray[$dieIdx];
+            if (isset($die->swingType) &&
+                $die->has_skill('Turbo') &&
+                $die->has_flag('IsAttacker')) {
```
cgolubi1: More input sanity-checking needed: this code needs to have an else case in which it returns an error saying, "Hey, you tried to specify Turbo values for a non-Turbo die", rather than just ignoring erroneous input.

src/engine/BMGame.php
```diff
@@ -1792,14 +1796,66 @@ protected function update_game_state_commit_attack() {
...
+                $setSuccess = $die->set_swingValue(array($die->swingType => $dieSize));
+                if (!$setSuccess) {
+                    $this->message = 'Invalid swing value for turbo die';
+                    return FALSE;
+                }
+            }
+        }
+
+        $this->playerArray[$playerIdx]->waitingOnAction = FALSE;
```
cgolubi1: More error-checking needed: you have to make sure that Turbo values were specified for all dice which meet the conditions in `do_next_step_choose_turbo_swing()` (has skill Turbo and flag IsAttacker) --- you can't just move on to the next state even if the player didn't specify values for some of their Turbo dice.

cgolubi1:
An error in responderTest, in test game 045, you have:
```php
        $this->verify_api_submitTurn(
            array(23, 5),
            'responder003 performed Trip attack using [tz(1/30=30)!:30] against [(6):1]; Attacker tz(1/30=30)! rerolled 30 => 23; Defender (6) rerolled 1 => 5, was captured. ',
            $retval, array(array(0, 1), array(1, 1)),
            $gameId, 3, 'Trip', 0, 1, '', array(0 => 30));
```

But the array index of the turbo swing die isn't 0, it's 1.  Backend validation of either spurious or missing turbo settings (two of my suggestions above about the BMGame function) would have caught that.

cgolubi1:
Another suggestion about responderTest --- you modify `verify_api_submitTurn()` to add the optional `$turboVals` property, but then you always stick in a default value when actually calling the API.  Instead, let's have responderTest do what the actual API does, which is make this property optional.  This change would have helped catch the fact that it's possible for a game to get trapped in `CHOOSE_TURBO_SWING`.  Here's my suggested diff to responderTest:

```php
$ git diff test/src/api/responderTest.php 
index ffd4da3..acc4903 100644
@@ -1105,7 +1105,7 @@ class responderTest extends PHPUnit_Framework_TestCase {
     protected function verify_api_submitTurn(
         $postSubmitDieRolls, $expMessage, $prevData, $participatingDice,
         $gameId, $roundNum, $attackType, $attackerIdx, $defenderIdx, $chat,
-        $turboVals = array()
+        $turboVals = NULL
     ) {
         global $BM_RAND_VALS;
         $BM_RAND_VALS = $postSubmitDieRolls;
@@ -1120,8 +1120,10 @@ class responderTest extends PHPUnit_Framework_TestCase {
             'attackerIdx' => $attackerIdx,
             'defenderIdx' => $defenderIdx,
             'chat' => $chat,
-            'turboVals' => $turboVals,
         );
+        if ($turboVals != NULL) {
+            $args['turboVals'] = $turboVals;
+        }
         $retval = $this->verify_api_success($args);
         $this->assertEquals($expMessage, $retval['message']);
         $this->assertEquals(TRUE, $retval['data']);
```

cgolubi1:
Hmm, now i'm seeing one reason you did what you did with the swing display UI --- the API doesn't actually tell the UI what swing type a given die is, and you didn't want to parse it from the string, so you just put the swing die information some other random place.  Well, that's actually reasonable, but you could have said so!

Anyway, thinking aloud: this isn't an issue for normal swing setting, because when we set swing values normally, we don't actually care which dice are affected by the swing setting --- in principle, there could be a bug in which a request was made for swing values which don't actually appear on any dice, and the UI would never notice a thing.

But your implementation here is that turbo swing is per-turbo die, not per-swing value.  (Yes, i did a test using two-turbo-X Ping, and we really ought to add a responder test using Ping to lock in whatever we do here to make sure it never regresses and breaks Ping in the future.)  Anyway, so we actually need to know that `qv(S)!` is an `S` swing in particular, and nothing tells us that.

Well, my two cents is, let's fix that --- let's have the API tell the UI what it needs to know.  The easiest and most Turbo-specific way to do that would be to add a turboSwingRequestArray, which has per-die indices like optRequestArray, but range values like swingRequestArray.  Arguably a better solution would be to have individual swing die properties always say that they are swing dice and of what type as part of their API information.  But that's a much bigger refactor.

Anyway, i discovered this while trying to trick out RandomAI to be able to react to Turbo properly, because of course it needs to do the same calculation as the UI does to figure out what values are legal, and there's actually no way to get that information short of parsing the recipe.  So, okay, let's fix that.

cgolubi1:
Okay, so, when i start a game using Darthcliff, who has an `f(16)!`, the UI gives me a blank screen, with the javascript error being that the `parseInt()` call in `Api.parseGamePlayerData()` blows up, because the backend is passing:
```json
"swingRequestArray":{"":null}
```

So.  I'm going to expand on my (tl;dr) previous comment by suggesting that we get out of the business of overloading `swingRequestArray` altogether here --- either make a `turboRequestArray` variables which handles both of the cases of a set of option values (including a length-1 set) and swing values, or add a `turboChoices` property to each Turbo die which does the same thing.  In either event, let's have the API pass something which the UI can use directly.  (And one suggestion i would make is that we not ask the UI to differentiate between swing and option --- we can afford the bits to just have the backend pass every number between 4 and 20, and then your UI widget can stay out of the logic business, and just use a dropdown for everything.)

cgolubi1:
Okay, so, this is all stuff i've encountered trying to get to the point where i can run novel tests using RandomAI.  In principle, nothing is stopping me from running regression tests of non-Turbo games using RandomAI right now, but i think i've raised a number of topics, and i'd like to get those discussed or iterated on before i go too much further.  So i'm going to pause until we discuss the raised topics.

Here's what i think i've raised so far which hasn't been addressed:
* UI for giving players information about (turbo) swing die ranges at all times.  I don't like the provided solution, and have an alternate proposal on the table, which i suggest we farm off into a separate issue independent of Turbo.
* Why is jquery_sourcemap merged into this pull?
* The UI needs to locally sanity-check bad turbo swing values which can be produced via the UI
* The backend needs to handle missing or incorrect turbo values correctly.  Specifically it MUST be impossible for a player using the API normally to end up stuck in `CHOOSE_TURBO_SWING`, and the backend SHOULD be friendlier about bad values in other ways as well.  Rundown of all my requests here:
    * Both `set_turbo_sizes()` should not have the code which returns success if `$turboSizeArray` is empty
    * `BMGame::set_turbo_sizes()` should return an error (rather than silently continue) if a turbo value is provided for an index for which a turbo value is not needed
    * `BMGame::set_turbo_sizes()` should return an error (rather than silently continue) if no turbo value is provided for an index for which a turbo value is needed
    * Fix turbo index bug in responderTest game 045
    * `verify_api_submitTurn` should specify `$turboVals` optionally to better match what the API actually does
* Information about needed turbo settings SHOULD be conveyed from the backend to the UI more cleanly:
    * Fix the Darthcliff bug in which a non-swing-or-option Turbo die leads to `swingRequestArray` being corrupted
    * Ideally create a Turbo-specific API structure for conveying information to the UI about available values, so that programmatic API clients can figure out what values are legal without parsing recipes, and have the UI use that structure too

Let's take a break and talk about all that before i continue.

cgolubi1:
This has merge conflicts, presumably with jquery_sourcemap, though who knows it might have been affected by the Game.gamePlayerDice refactor too.

src/engine/BMGame.php
blackshadowshade: I think we're going to run into trouble here with Morphing Turbo dice, which may become non-swing dice that just happen to still have Turbo.

I've also got to think more about whether we really require the player to input Turbo values in this case, where the new swing value will just be ignored.

src/engine/BMGame.php
blackshadowshade: Again, Morphing Turbo.

blackshadowshade:
Right, let me think out loud here about a die like m(X)! capturing a (4).

We've decided in the past that this should end up giving us a m(4)! die, which could then revert to swing if it later captures a swing die.

What should the correct behaviour of the front-end be? I can see a couple of options:
- ignore any swing value specified for this die
- require the player to actively not specify a swing value
- somehow do some smarts in the front-end alone working out that no swing value will be necessary
- somehow do some smarts in a combination of the front-end and the back-end working out that no swing value will be necessary

It's important to consider this question here, since it changes the way that I address many of the points that you have mentioned above, especially regarding validation and error messages.

What do you (testers included) think we should be doing for this particular case?

blackshadowshade:
Chaos said in IRC:

i think we want all three of:
(1) backend owns telling the UI what valid turbo choices are
(2) one-stage submit of turbo choices
(3) UI doesn't make surprising API calls when you interact with page elements


Based on the discussion there, we're going to outlaw morphing turbo for the moment, and thus require all turbo values to be set. The back-end will provide full information to the front-end about which values are valid, which makes things easier with validation and also testing using RandomAI.

AdmiralJota:
I think that my expectation would be that m(X)! would become an m(4)!, and I would still be provided with a turbo selector of some sort (whether a real one or a dummy non-interactive one) which would only be able to select 4. So it might be a drop-down with only one option, or a greyed-out textbox with 4 already filled in or whatever.

irilyth:
I think it's relatively fine if the UI pops up an element to indicate what's going on, as long as the player doesn't have to try to guess what to do with it (like a box where if you type a "4" it works, and anything else gives you an error, I think would not be fun).

I think it'd also be fine for the UI to not even pop up anything at all. Jota, do you think it's important for the UI to pop up something even if you don't have to (and can't) do anything with it?

AdmiralJota:
Not important, no. I think that the consistency would be nice, but I can't think of a case where the absence of the UI element for a turbo non-swings would be a significant problem.

blackshadowshade:
So, I think this is now ready for further iteration.

- I have removed all swing information from the screen as requested
- Possible turbo sizes are now provided by the API in the new turboSizeArray
- Non-swing, non-option turbo dice are now supported
- Validation of submitted turbo sizes has been beefed up both on the front-end and the back-end

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1201/ (if it passes)

cgolubi1:
Great.  I'll look at the code as soon as i can, and take it from there.

blackshadowshade:
So, I've attempted to rebase this to master, but the merge conflicts in that process were so intractable that I'm going to say that this particular pull request will unfortunately not be entirely rebased into one commit.

One other unrelated comment for Chaos: When you're implementing logging for this, you'll probably find it useful to set up a die flag for the change of size caused by turbo.

blackshadowshade:
I guess I have a general question here (possibly to @irilyth) on how we might possibly be able to get this into one single commit, since I failed via rebase. Would it be realistic to try to do this with a brute force method involving file diffs, since we know which files have been changed? And if so, is there a standard way of doing this?

blackshadowshade:
Okay, answering my own question, I think:

http://www.thegeekstuff.com/2014/03/git-patch-create-and-apply

@irilyth, would you mind taking a look at that and seeing if it makes sense to you?

irilyth:
Yeah, this seems basically correct, although I sort of think it's not likely to actually get you out of dealing with merge conflicts, since if you create a diff between your branch's HEAD and its branch point, you're still going to see conflicts when you try to apply those diffs to master.

I've sort of lost track of what you're trying to do here. Why would rebasing this into one commit help? Given a branch with a hundred commits and a bunch of intractable merge conflicts with master, I'm not sure why it'd make your life any easier to squash it down to a single commit -- whatever files have conflicts are still going to have conflicts.

I'm happy to poke at this a bit, if it's still in a relatively intact state. I was able to clone your repo, and can try do some stuff with it, if you can say a little more about what effect you want to accomplish (e.g. "I wish that the changes I had made in the various commits in branch 91_turbo_swing were now in a branch that was a child of the current master" or whatever).

irilyth:
So one thing here is that it looks like you've merged master into your branch a bunch, which means that you have commits from master that are now part of your branch. That's sort of basically fine, but it means that if you try to rebase your branch, all of those commits get new commit IDs. Maybe Github can then deal with this in some sensible manner, but I sort of predict that it'll make a mess; and that if you want to be able to rebase a branch in your fork in order to submit a PR (which I think is a good idea), then you don't generally want to merge master into your branch.

(That doesn't really speak to the current situation, but it's something we could talk more about some time if you like.)

blackshadowshade:
If I understand correctly, the thing that I really want to do is take the current state, which is fully merged with the current master, and diff it to the current master. This gets around the whole issue of merge conflicts, and creates a single commit (patch) that can be applied to master to get to the current state.

irilyth:
Ok, so if you do

    git diff master 91_turbo_swing --stat

in your clone, you'll get a list of files; I see "42 files changed, 378 insertions(+), 911 deletions(-)" at the bottom. If that looks right, then

    git diff master 91_turbo_swing > turbo.patch

will create a patch file.

I'm not so sure about this, though, because e.g.

    git diff master 91_turbo_swing deploy/database/data.button.sql

shows changes about Cheese Weasel and Perf, which don't seem like they should be different between your branch and master.

Hmm, something's weird with this, because when I do

    git checkout master
    grep 576 deploy/database/data.button.sql

in my clone of your fork, I get

    (576, 'Peri',            '(6) (6) (10) (X) (X)',              0, 1, (SELECT id FROM buttonset WHERE name="ZOECon")),

so maybe master in your fork isn't up to date with master in upstream?

blackshadowshade:
I'd do the diff to upstream/master, and make sure that I fetched the latest changes as necessary.

irilyth:
> so maybe master in your fork isn't up to date with master in upstream?

I think that's it. In my clone of your fork, I did

    git remote add upstream git@github.com:buttonmen-dev/buttonmen.git
    git fetch upstream

and then

    git diff origin/master upstream/master --stat

shows a bunch of changes, "17 files changed, 48 insertions(+), 9 deletions(-)" being the bottom line. So you probably want to push that to origin/master, like so:

    git checkout master
    git pull upstream master
    git push origin master

Then the diff with your branch should be smaller:

    git diff master 91_turbo_swing --stat

says "26 files changed, 863 insertions(+), 369 deletions(-)" as the bottom line, and

    git diff master 91_turbo_swing > turbo.patch

gives a patch file which seems at a glance to be more just about Turbo.

So if that is, you should then be able to do

    git checkout master
    git checkout -b 91_turbo_swing_one_commit # or whatever, some new branch name
    git apply turbo.patch

and then

    git diff --stat

should say "26 files changed, 863 insertions(+), 369 deletions(-)", just like the bottom line of 'git diff master 91_turbo_swing --stat' before.

You can verify that by lookingaround, and then do

    git commit -m "Consolidated amalgamated Turbo Swing commit." -a

or whatever, and voila. Put in a PR, and sleep the sleep of the just.

(And, if other changes come in to master before your PR is merged, rebase your branch onto master, and push it again. I think this shouuld work:

    git checkout master
    git pull upstream master
    git push origin master
    git checkout 91_turbo_swing_one_commit
    git rebase master
    git push --force origin HEAD

I think I haven't fully tested that sequence with Github PRs, but I think it'll work. I intend to try it once there are changes in upstream/master to rebase onto. :^)

blackshadowshade:
Thanks, irilyth, nicely comprehensive. I'll try that tonight.  :)

blackshadowshade:
Closing this in favour of #2152.

PR #2051, opened 2016-06-30
Added new die images

blackshadowshade:
This pull request just adds the new images that we'll need, once we add the logic allowing players to select the die images that they desire. This has been split out of #2022.

I presume that Jenkins shouldn't care one iota about this pull request.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1182/

PR #2050, opened 2016-06-30
refactor the Game.gamePlayerDice() function

cgolubi1:
* Related to #822 and #2047, a suggested replacement refactor for #1981
* Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/461/ (if it passes)

This splits the monolith `Game.gamePlayerDice()` function into a number of distinct functions, which ideally has the effect of splitting things cleanly into functions which gather information and functions which lay out particular elements, and making operation consistent across active and captured dice.

The refactor *should* be a total noop --- the only difference is that it reorders class attributes in a couple of cases, but it should not add or remove attributes in any case, but only move code around.

PR #2055, opened 2016-07-02
New user preference for die image backgrounds

### LINKED ISSUE #822: Option dice representation on game page is cluttered

blackshadowshade:
Fixes #822.

Depends on #2050 and #2051.

This adds a new user preference for the type of die background image: circle, realistic, and symmetric, with a default of circle.

I've changed the design of the user preference page to accommodate the new user preference. Naturally, I'm open to suggestions about this redesign.

I've deliberately moved the encoding of die image URL to gui.css, since different browsers deal with url() differently, which makes it almost impossible to write QUnit deepEqual tests that pass across multiple browsers.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1188/ (if it passes)


cgolubi1:
Three high-level things off the top:
* At the moment, github is showing me the diffs including the die images and the Game.js refactor, so i can't actually view this diff reasonably via github.  So a full review is going to have to wait.
* Should the new property really be "die_background"?  I think "background" is confusing because "background" implies a color like black or grey or white.  I'd prefer "die_style" or "die_shape", or maybe "die_image_style".  Thoughts?
* I dislike the move of the die image processing to gui.css.  It makes the code a lot less maintainable solely to make test happy.  There's got to be another way to make the tests happy.  Without understanding what the problem is, my immediate brainstorming ideas are:
    * hack `BMTestUtils.DOMNodePropArray()` to modify `url()` results and make the test output from the page more consistent
    * mock `url()` so it does something different while testing

But tell me more about the problem and i'll see what else i can come up with.

AdmiralJota:
What's wrong with putting the URL of the background image of an element into a rule in a CSS file and having the code assign CSS classes to elements to match those rules? That seems like a pretty appropriate use of CSS to me.

As for the actual implementation of it, I think I'd suggest using separate classes for the die style (`.die_bg_circle`, `.die_bg_symmetric`, `.die_bg_asymmetric` -- which could be on the outer container holding all the dice), the die status (`.active`, `.inactive`, `.taken`) and the die size (`.2_sided`, `.4_sided`, `.6_sided`, `.8_sided`, `.10_sided`, `.12_sided`, `.20_sided`, `.30_sided`), and then basing the CSS rules off of the combinations of those (such as `.die_bg_realistic .active.2_sided {  background-image: url(images/die/realistic/d2active.png) }`).

cgolubi1:
> What's wrong with putting the URL of the background image of an element into a rule in a CSS file and having the code assign CSS classes to elements to match those rules? That seems like a pretty appropriate use of CSS to me.

I don't like having so much copy/paste duplication of logic, like:
```
.die_bg_symmetric_active_2 {
    background-image: url(images/die/symmetric/d2active.png)
}

.die_bg_symmetric_active_4 {
    background-image: url(images/die/symmetric/d4active.png)
}

.die_bg_realistic_active_2 {
    background-image: url(images/die/realistic/d2active.png)
}

.die_bg_realistic_active_4 {
    background-image: url(images/die/realistic/d4active.png)
}

...
```

This uses JS code to determine that "symmetric" is the right style and "2" is the right size, but then, instead of taking advantage of the file naming convention to have the JS code assemble the URL based on those derived parameters, it counts on someone having accurately copy/paste/modified our naming convention in each line of the file.  That is, nothing stops me from typoing the URL in the third line, or forgetting to insert these CSS entries when we add a new die style, and winding up with a hard-to-debug problem.  Since JS can assemble URLs for images, and since the naming convention for files is totally a function of the logic, duplicating it in CSS introduces opportunities for error to no benefit.  If we're doing that just to make the tests happy, we should figure out another way to make the tests happy.

cgolubi1:
Incidentally, i was googling around to try to make sure i understood what we're doing here, and came across http://www.w3schools.com/cssref/pr_background-image.asp which says:
```
By default, a background-image is placed at the top-left corner of an element, and repeated both vertically and horizontally.
```

Do we need to worry about something causing the background image to be repeated under some circumstances, and set a property like:
```
background-repeat: no-repeat;
```

blackshadowshade:
Ok, the problem here is a difference that I noticed between Safari and Firefox (thus the need for the JQuery map file to enable the console under Safari).

When I had the deepEqual use
  url(images/...), 
it passed under Firefox but failed under Safari and Jenkins.

Alternatively, when I had the deepEqual use
  url(document.location.origin + "\ui\images\..."
it passed under Safari but not under Firefox. It's also potentially problematic under Jenkins because of the resolution of document.location.origin.

I must have spent 3+ hours trying to get this QUnit test work, and gave up in disgust, thus my shift of the paths into the CSS file. I actually think that the shift is not a bad thing, and if Jota can work some CSS magic rules, that probably gets around both the issues of the QUnit tests and the multitude of rules in the CSS.

cgolubi1:
Hmm, okay, that does strike me as likely to be fixable by modifying the test rig.  I'll give it a try once i get to that point in testing.  As for "CSS magic rules", sure, the thing i care about is the fact that we're not using variables here; i don't care whether the variables live in CSS or JS.  That said, i'm not aware of variables in CSS being a thing --- i thought that was what the JS was for.

blackshadowshade:
I'm just dropping this here in case it's actually pertinent, I haven't taken a close look (I have an early start tomorrow):

https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_variables

cgolubi1:
Thanks for linking to that --- i skimmed it, and it looks like browser compatibility might be an issue, so i doubt it would make our lives better.  Let me know if a more detailed read makes you more optimistic about it.

blackshadowshade:
BTW, if you want to play around with the old logic where I was adding the background-image css directly, just revert to 462aaa3.

cgolubi1:
Okay, so it seems like your problem is that using the `.css()` function with a `url()` in the argument behaves differently between different browsers.  But just setting `divOpts['style']` to something containing a `url()` works fine.  The code in other places does both, but i'll note that there's no existing code in Game.js which uses `.css()` --- it all uses `style`, so i don't see it as any kind of cop-out to use `style` within Game.js, and afaict it makes this just work.  Here's the diff i applied to your suggested commit https://github.com/buttonmen-dev/buttonmen/commit/462aaa3517395ed72452e2954a8712e16fd97f45 as a result of which the tests Just Worked in both Firefox and Safari, and the game mat looked good too from the minimal testing i did:

```diff
$ git diff
diff --git a/src/ui/js/Game.js b/src/ui/js/Game.js
index c437379..15deed5 100644
--- a/src/ui/js/Game.js
+++ b/src/ui/js/Game.js
@@ -2797,18 +2797,15 @@ Game.createGameMatDieDiv = function(die, player, dieStatus, isClickable) {
     dieNumberSpanOpts.html = '&nbsp;' + die.value + '&nbsp;';
   }
 
-  var dieDiv = $('<div>', divOpts);
-  dieDiv.append($('<span>', dieNumberSpanOpts));
-
-  dieDiv.css(
-    'background-image',
-    Game.backgroundImagePath(
+  divOpts['style'] = 'background-image: ' + Game.backgroundImagePath(
       die.sides,
       dieStatus,
       isClickable
-    )
   );
 
+  var dieDiv = $('<div>', divOpts);
+  dieDiv.append($('<span>', dieNumberSpanOpts));
+
   return dieDiv;
 };
 
diff --git a/test/src/ui/js/test_Game.js b/test/src/ui/js/test_Game.js
index 3544ab9..269669c 100644
--- a/test/src/ui/js/test_Game.js
+++ b/test/src/ui/js/test_Game.js
@@ -1603,7 +1603,7 @@ test("test_Game.createDieContainerDiv", function(assert) {
         [ "SPAN", { "class": "die_recipe_player" }, [ "(6)" ] ] ]
       ],
       [ "DIV", { "class": "die_border", "style": "border: 2px solid #dd99dd" }, [
-        [ "DIV", { "class": "die_img", "style": "background-image: url(\"images/die/symmetric/d6active.png\");" }, [
+        [ "DIV", { "class": "die_img", "style": "background-image: url(images/die/symmetric/d6active.png)" }, [
           [ "SPAN", { "class": "die_overlay die_number_player" }, [ "2" ] ] ]
         ] ]
       ]
@@ -1622,7 +1622,7 @@ test("test_Game.createDieContainerDiv", function(assert) {
       "title": "6-sided die"
     }, [
       [ "DIV", { "class": "die_border", "style": "border: 2px solid #ddffdd" }, [
-        [ "DIV", { "class": "die_img", "style": "background-image: url(\"images/die/symmetric/d6active.png\");" }, [
+        [ "DIV", { "class": "die_img", "style": "background-image: url(images/die/symmetric/d6active.png)" }, [
           [ "SPAN", { "class": "die_overlay die_number_opponent" }, [ "2" ] ] ]
         ] ]
       ],
```

blackshadowshade:
Such a simple solution! I'll apply this when I have time.

cgolubi1:
When you get a chance, do that, and also update this pull to account for the merges of #2050 and #2051 and resolve the merge conflict.  Then i'll be able to review it.

blackshadowshade:
So, I've updated this pull request by doing a force revert to 462aaa3 and then applying the changes as suggested, plus all necessary changes to the QUnit tests.

That means that you'll probably need to do a forced pull, or alternatively, delete your current local version of this branch and pull a fresh copy of it.

Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1189/ (if it passes)

src/engine/BMInterface.php
```diff
@@ -195,7 +196,7 @@ public function load_api_game_data($playerId, $gameId, $logEntryLimit) {
         return NULL;
     }
 
-       // Can the active player edit the most recent chat entry in this game?
+    // Can the active player edit the most recent chat entry in this game?
```
cgolubi1: Can you skip this change?  It'll cause a merge conflict with #2058.  If you can't, no big deal, i'll handle the merge conflict.

src/ui/js/Game.js
```diff
@@ -2791,11 +2797,64 @@ Game.createGameMatDieDiv = function(die, player, dieStatus, isClickable) {
...
+    sidesRoundedUp = 8;
+  } else if (sides <= 10) {
+    sidesRoundedUp = 10;
+  } else if (sides <= 12) {
+    sidesRoundedUp = 12;
+  } else if (sides <= 20) {
+    sidesRoundedUp = 20;
+  } else {
+    sidesRoundedUp = 30;
+  }
```
cgolubi1: Nit: this isn't so inelegant i'll reject it, but it is inelegant --- you could instead iterate over an array of rounded-up sizes and compare them to `sides`.  If you want to do that and only didn't because you didn't want to figure out the syntax, i can take a few minutes this evening and come up with it.

src/ui/js/UserPrefs.js
```diff
@@ -279,16 +285,22 @@ UserPrefs.actionSetPrefs = function() {
...
+  var matPrefs = {
+    'die_background_select': {
+      'text': 'Die background',
+      'type': 'select',
+      'value': Api.user_prefs.die_background,
+      'source': dieBackgroundDefaults,
+    },
     'noImages': {
-      'text': 'Don\'t load button or player images',
+      'text': '(only this browser) Don\'t load button or player images',
```
cgolubi1: Hmm, is it clear enough what this means?  I think i'd say: "Don't load button or player images (only applies to the current browser)", and likewise for compact mode.

src/ui/js/Game.js
blackshadowshade: I'm happy to do that myself.

test/src/ui/js/test_Game.js
```diff
@@ -1671,28 +1673,101 @@ test("test_Game.createGameMatDieDiv", function(assert) {
...
   var dieDivProps = BMTestUtils.DOMNodePropArray(dieDivJQuery[0]);
-  var expectedDivProps = [ "DIV", { "class": "die_img" }, [
+  var expectedDivProps = [ "DIV", { "class": "die_img die_dead", "style": "background-image: url(images/die/symmetric/d6taken.png)" }, [
     [ "SPAN", { "class": "die_overlay die_number_player" }, [ "\u00A0" + "2" + "\u00A0" ] ] ]
   ];
   assert.deepEqual(dieDivProps, expectedDivProps, "dieDiv looks correct for captured die");
 });
 
+test("test_Game.backgroundImagePath", function(assert) {
+  Api.game = { player: { dieBackgroundType: 'symmetric' } };
```
cgolubi1: Hmm, why do you need to set Api.game in this test?  Doesn't `Game.backgroundImagePath()` take the background type as an argument?

test/src/ui/js/test_Game.js
```diff
@@ -1671,28 +1673,101 @@ test("test_Game.createGameMatDieDiv", function(assert) {
...
+  assert.ok(Game.backgroundImagePath(19, 'active', true, 'realistic').indexOf('images/die/realistic/d20active.png') >= 0, 'active d19');
+  assert.ok(Game.backgroundImagePath(20, 'active', true, 'realistic').indexOf('images/die/realistic/d20active.png') >= 0, 'active d20');
+  assert.ok(Game.backgroundImagePath(21, 'active', true, 'realistic').indexOf('images/die/realistic/d30active.png') >= 0, 'active d21');
+  assert.ok(Game.backgroundImagePath(29, 'active', true, 'realistic').indexOf('images/die/realistic/d30active.png') >= 0, 'active d29');
+  assert.ok(Game.backgroundImagePath(30, 'active', true, 'realistic').indexOf('images/die/realistic/d30active.png') >= 0, 'active d30');
+  assert.ok(Game.backgroundImagePath(31, 'active', true, 'realistic').indexOf('images/die/realistic/d30active.png') >= 0, 'active d31');
+
+  Api.game.player.dieBackgroundType = 'symmetric';
+
+  assert.ok(Game.backgroundImagePath(1, 'active', false, 'symmetric').indexOf('images/die/symmetric/d2inactive.png') >= 0, 'inactive d1');
```
cgolubi1: Actually, okay, what is happening in this test?  You call `Game.backgroundImagePath` with `'active'` as an argument, and verify that the index of an inactive die shows up in the result?  Any chance this test is simply a noop?

cgolubi1:
Okay, so, you have some nits, merge conflicts, and unit tests (i think) to fix, and then when this is back in my court, i will:
* Look at the code changes and make sure they look okay
* Do the standard database rebuild-vs-update test
* Check out the visuals of the new preferences page, and make sure they look okay
* Check out the visuals of the die mat with different background preferences, and make sure they look okay
* Grep in the code for instances of `die_alive` and `die_dead` (i don't mind you keeping them around, though i also wouldn't mind you removing them, but i want to make sure we're not counting on them for anything)

test/src/ui/js/test_Game.js
blackshadowshade: No, I think this is correct. As far as I understand, the status is either 'active' or 'taken'. The thing that distinguishes 'active' from 'inactive' is whether it's clickable or not, thus the different third parameter.

test/src/ui/js/test_Game.js
blackshadowshade: No, Game.backgroundImagePath() doesn't take the background type as an argument. It reads it directly from Api.game.player.dieBackgroundType:

    return 'url(images/die/' +
         Api.game.player.dieBackgroundType +
         '/d' +
         sidesRoundedUp +
         imageType +
         '.png)';

If you think this behaviour is incorrect, let me know at which level you would like the player-selected background type injected into the logic. The higher up it's injected, the more we just passing around a parameter that is only used in this one place.

blackshadowshade:
Okay, addressed the merge conflict, the extended bracketed comment, and the rounded up sides code.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1192/ (if it passes)

cgolubi1:
Okay, you have a subtle merge conflict here --- looks like your database update file, `00822_die_background_pref_01.sql`, dropped vacation_message from the player view:
```diff
sandbox,[/buttonmen/deploy/database/updates],03:22(0)$ diff -u ~/buttonmen_db.*
--- /home/chaos/buttonmen_db.rebuilt.sql        2016-07-09 03:20:57.192989569 +0000
+++ /home/chaos/buttonmen_db.updated.sql        2016-07-09 03:20:51.474131693 +0000
@@ -1126,7 +1126,6 @@
   `image_size` tinyint NOT NULL,
   `uses_gravatar` tinyint NOT NULL,
   `comment` tinyint NOT NULL,
-  `vacation_message` tinyint NOT NULL,
   `homepage` tinyint NOT NULL,
   `favorite_button_id` tinyint NOT NULL,
   `favorite_buttonset_id` tinyint NOT NULL,
@@ -1288,7 +1287,7 @@
 /*!50001 SET collation_connection      = latin1_swedish_ci */;
 /*!50001 CREATE ALGORITHM=UNDEFINED */
 /*!50013 DEFINER=`root`@`localhost` SQL SECURITY DEFINER */
-/*!50001 VIEW `player_view` AS select `a`.`auth_key` AS `auth_key`,`i`.`id` AS `id`,`i`.`name_ingame` AS `name_ingame`,`i`.`password_hashed` AS `password_hashed`,`i`.`name_irl` AS `name_irl`,`i`.`email` AS `email`,`i`.`is_email_public` AS `is_email_public`,`ps`.`name` AS `status`,`i`.`dob_month` AS `dob_month`,`i`.`dob_day` AS `dob_day`,`i`.`gender` AS `gender`,`i`.`autoaccept` AS `autoaccept`,`i`.`autopass` AS `autopass`,`i`.`fire_overshooting` AS `fire_overshooting`,`i`.`monitor_redirects_to_game` AS `monitor_redirects_to_game`,`i`.`monitor_redirects_to_forum` AS `monitor_redirects_to_forum`,`i`.`automatically_monitor` AS `automatically_monitor`,`i`.`image_path` AS `image_path`,`i`.`image_size` AS `image_size`,`i`.`uses_gravatar` AS `uses_gravatar`,`i`.`comment` AS `comment`,`i`.`vacation_message` AS `vacation_message`,`i`.`homepage` AS `homepage`,`i`.`favorite_button_id` AS `favorite_button_id`,`i`.`favorite_buttonset_id` AS `favorite_buttonset_id`,`i`.`die_background` AS `die_background`,`i`.`player_color` AS `player_color`,`i`.`opponent_color` AS `opponent_color`,`i`.`neutral_color_a` AS `neutral_color_a`,`i`.`neutral_color_b` AS `neutral_color_b`,`i`.`last_action_time` AS `last_action_time`,`i`.`last_access_time` AS `last_access_time`,`i`.`creation_time` AS `creation_time`,`i`.`fanatic_button_id` AS `fanatic_button_id`,`i`.`n_games_won` AS `n_games_won`,`i`.`n_games_lost` AS `n_games_lost`,`d`.`dhs_player_id` AS `dhs_player_id`,`d`.`dhs_player_name` AS `dhs_player_name` from (((`player` `i` left join `player_auth` `a` on((`i`.`id` = `a`.`id`))) left join `player_status` `ps` on((`ps`.`id` = `i`.`status_id`))) left join `dhs_player` `d` on((`d`.`bw_player_id` = `i`.`id`))) */;
+/*!50001 VIEW `player_view` AS select `a`.`auth_key` AS `auth_key`,`i`.`id` AS `id`,`i`.`name_ingame` AS `name_ingame`,`i`.`password_hashed` AS `password_hashed`,`i`.`name_irl` AS `name_irl`,`i`.`email` AS `email`,`i`.`is_email_public` AS `is_email_public`,`ps`.`name` AS `status`,`i`.`dob_month` AS `dob_month`,`i`.`dob_day` AS `dob_day`,`i`.`gender` AS `gender`,`i`.`autoaccept` AS `autoaccept`,`i`.`autopass` AS `autopass`,`i`.`fire_overshooting` AS `fire_overshooting`,`i`.`monitor_redirects_to_game` AS `monitor_redirects_to_game`,`i`.`monitor_redirects_to_forum` AS `monitor_redirects_to_forum`,`i`.`automatically_monitor` AS `automatically_monitor`,`i`.`image_path` AS `image_path`,`i`.`image_size` AS `image_size`,`i`.`uses_gravatar` AS `uses_gravatar`,`i`.`comment` AS `comment`,`i`.`homepage` AS `homepage`,`i`.`favorite_button_id` AS `favorite_button_id`,`i`.`favorite_buttonset_id` AS `favorite_buttonset_id`,`i`.`die_background` AS `die_background`,`i`.`player_color` AS `player_color`,`i`.`opponent_color` AS `opponent_color`,`i`.`neutral_color_a` AS `neutral_color_a`,`i`.`neutral_color_b` AS `neutral_color_b`,`i`.`last_action_time` AS `last_action_time`,`i`.`last_access_time` AS `last_access_time`,`i`.`creation_time` AS `creation_time`,`i`.`fanatic_button_id` AS `fanatic_button_id`,`i`.`n_games_won` AS `n_games_won`,`i`.`n_games_lost` AS `n_games_lost`,`d`.`dhs_player_id` AS `dhs_player_id`,`d`.`dhs_player_name` AS `dhs_player_name` from (((`player` `i` left join `player_auth` `a` on((`i`.`id` = `a`.`id`))) left join `player_status` `ps` on((`ps`.`id` = `i`.`status_id`))) left join `dhs_player` `d` on((`d`.`bw_player_id` = `i`.`id`))) */;
 /*!50001 SET character_set_client      = @saved_cs_client */;
 /*!50001 SET character_set_results     = @saved_cs_results */;
 /*!50001 SET collation_connection      = @saved_col_connection */;
@@ -1302,4 +1301,4 @@
 /*!40101 SET COLLATION_CONNECTION=@OLD_COLLATION_CONNECTION */;
 /*!40111 SET SQL_NOTES=@OLD_SQL_NOTES */;
 
--- Dump completed on 2016-07-09  3:20:57
+-- Dump completed on 2016-07-09  3:20:51
sandbox,[/buttonmen/deploy/database/updates],03:22(1)$ 
```

cgolubi1:
So i think the new "only applies" text is okay, but i still think this is pretty weird-looking in the preferences page:
only_applies

It just looks wrong to me to start with a parenthetical, to have a capital letter in the middle of a sentence, and to not put the important part of the preference at the beginning where it's eye-catching.  Reversing the sentence order strikes me as much less weird-looking:

only_applies_after

src/ui/js/Game.js
```diff
@@ -2791,11 +2797,55 @@ Game.createGameMatDieDiv = function(die, player, dieStatus, isClickable) {
...
+ * @param  int     sides          Number of sides of the die
+ * @param  string  dieStatus      Status of the die ('active' or 'captured')
+ * @param  boolean isClickable    Is the die clickable?
+ * @return string
+ */
+Game.backgroundImagePath = function(sides, dieStatus, isClickable) {
+  var imageType;
+
+  var sidesRoundedArray = [2, 4, 6, 8, 10, 12, 20, 30];
+  var sidesRoundedUp = Math.max.apply(null, sidesRoundedArray);
```
cgolubi1: IMO, there'd be nothing wrong with `var sidesRoundedUp = sidesRoundedArray[sidesRoundedArray.length - 1];` if you didn't want to use Math functions, and it'd be slightly more clear (just because the null business is a little confusing).  Your code below assumes the array is sorted, so there's no point stressing about the case in which it isn't.

cgolubi1:
Okay, i don't see any other issues here.  The base functionality looks good.

AdmiralJota:
Two thoughts:

* Another benefit to putting the parenthetical at the end instead of the beginning is that it makes the two relevant options line up more nicely, which both is aesthetically pleasing and also associates the two related things more closely.

* But another alternative to consider would be to leave the "Browser Preferences" the way it was (so no parentheticals are needed) and instead to change "Color Preferences" to something like "Display Preferences" and add the new die style preference there.

blackshadowshade:
What about putting the parentheses after the option?

i.e.,

Don't load button or player images  []  (only applies to the current browser)

blackshadowshade:
So, I've put in a fix for the merge conflict and the nit. I've also added a change to the layout of the browser preferences that may improve the visuals. What do you think?

prefs

AdmiralJota:
That seems fine to me.

cgolubi1:
Yeah, i like that visual.  Obviously, i'll need to look at the code and i may have comments there, but the result looks good from your screencap.

deploy/amp/database_initialise.php
```diff
@@ -8,9 +8,9 @@
 // james: uncomment the following lines to also initialise the
 //        primary 'buttonmen' database
 //
-//        recreate_database('buttonmen');
-//        $conn = db_connect('buttonmen');
-//        run_all_sql_files($conn);
+        recreate_database('buttonmen');
+        $conn = db_connect('buttonmen');
+        run_all_sql_files($conn);
```
cgolubi1: Did you intend to commit this change?  I don't really care about changes to this file, but it's a little distracting --- maybe we should talk about how you invoke this file, and add a CLI flag you can use to also recreate the primary database, so that you don't need to do that by modifying the file.

(FWIW, what /usr/local/bin/create_buttonmen_databases does is always drop-and-recreate buttonmen_test, and recreate buttonmen only if it does not exist.  This is a safety feature (and, ahem, it did save my bacon on one of the real sites once), but it means that my test process can be to drop the buttonmen database and then just run the script as usual.)

Anyway, for purposes of this pull, i just need to know whether you want this change in here, or want to remove it.  But the previous paragraph might have some thoughts for the future.

cgolubi1:
Okay.  Everything looks fine here --- i just want to hear back from shadowshade about whether the modification to deploy/amp/initialize_databases.php was intended to be committed, and then i can merge this.  (I'm being cautious because it looks like an accident, and if there are other testers who use AMP right now, it looks like they could smite their buttonmen databases by accident using it.)

blackshadowshade:
Ah, didn't realise that checked in also. I've reverted it.

cgolubi1:
Thanks.  Merging.

Issue #464, opened 2014-01-24
allow privacy settings for game chat


cgolubi1:
Players in a game should be able to specify whether the game chat is visible only to the participants, or is visible to anyone who looks at the game.

dwvanstone:
I'm adding this as my bugfix of the month -- I think as new players come in they would expect their chats to be private (by default).

AdmiralJota:
I'll place my vote, then: I'd like it to be public by default. 

There's a lot of potentially entertaining chat out there, such as story games, or interesting tidbits like http://buttonweavers.com/ui/game.html?game=1, or cases where one player wants to share something cool that happened but the other player isn't around to change the game settings to public. But on the old site virtually no one ever bothered to turn on public chat, so all of that was lost. 

But of course, it should be very clear to players when the game chat is public, and it should be very easy for players to unilaterally make it private (both from the game creation screen and from the game page -- I like @cgolubi1's suggestion of a button that's independent of submitting your turn, to make it even easier).

Perhaps it's voyeuristic of me, but I like being able to go back and see what people were talking about during a game, their thoughts on strategies and so on. And it's very rare that I'd be using game chat to discuss anything private -- and that'll be *especially* true once we implement player mail.

irilyth:
I've been in chats where people said things that they didn't realize were public, not catastrophically so, but things they wouldn't have said if they'd realized.

I still think they should be public by default, for basically the reasons that Jota mentioned. Just put "(PUBLIC)" in bold under the word "Chat" next to the box (and perhaps "(private)", perhaps not in bold, if its private), and I think it'll be quite obvious. (Or something else if that turns out not to be obvious enough.)

dwvanstone:
I think game creation would be a good place to add a dropdown option of "Chat is public" [by default, which is fine by me] and "Chat is private". Marking somewhere on the page itself that Chat is PUBLIC would be helpful. 

Maybe in a alter fix we could allow users to change the chat from public to private or private to public.

blackshadowshade:
I don't like the idea of an interim fix here. If we're going to do this, then let's do it right from the beginning.

I'm all for having chat default to public, but with an easy way on the game screen to change the privacy setting. We would store and display both player's settings, and if either player has chosen private, then that game's chat would only show for the two of them.

blackshadowshade:
Chaos, you're currently assigned to this issue. Will you be tackling this issue this month, or will I?

cgolubi1:
[...]

PR #2058, opened 2016-07-03
Refactor all chat database methods into BMInterfaceGameChat module

cgolubi1:
* Related to #464 and to #974
* Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/466/ (if it passes)

The current state of the BMInterface refactor in progress is somewhat unfriendly to chat-related work, because chat methods are located in both BMInterface.php and BMInterfaceGame.php.  This refactor simply moves all of those methods into a new module, BMInterfaceGameChat.php, which parallels BMInterfaceGameAction.php.  This will make the chat privacy work, which will pretty much all happen at the database and API layers, more tractable.

The only change i made other than moving functions and adding comments is that i got rid of the `log_game_chat()` method, folding its one line of logic into `save_chat_log()`.  My guess based on a read of the historical record is that having two very similar small functions was introduced when the `save_chat_log()` function was initially split out of `save_game()`, and was never actually necessary.


blackshadowshade:
This looks fine, works fine on my MAMP stack. Tested inserting, deleting, altering chat by both players.

PR #2059, opened 2016-07-04
add tests for Login.js

cgolubi1:
* Intended to help with testing #1859 
* Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/469/

This doesn't bring us to 100% code coverage of Login.js, but i think it hits the majority of the code which defines HTML contents of header and footer elements, so i believe it should be sufficient for getting a handle on what #1859 changes.


blackshadowshade:
Right, this works fine on my WAMP stack under Chrome and Firefox.

PR #2061, opened 2016-07-05
Script to install DHS site stats into a buttonweavers database

cgolubi1:
* Part of #614 
* Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/470/ (if it passes)

On the theory of getting projects unblocked by taking small steps on them, this is a small step towards stats, namely, actually getting the DHS stat data installed in the dev and prod site tables which we setup a long time ago to hold it.

We don't want this data in the repo, because it identifies players and is also big.  So this script assumes you have a copy of the JSON blob i've assembled from my DHS records and have shared with a couple of people, and transcribes that blob into the appropriate database tables.

I tested this on virtualbox using both a blank database, and a database backup taken from the prod site.  With the prod site database, it identifies 103 users as being carried over from DHS to BW on the basis of their usernames (and any which were missed or extraneous, we can fix after the fact by hand, since that's a site-by-site thing which will become part of the relevant database state).

This adds 12.7MB of data to the prod database, which is about 2% of the current size of that database.  That's sane at a glance, given that the JSON source file (which has extraneous key names) is 17MB.


blackshadowshade:
So, I'm happy for us to try this. It looks reasonable, and if it's wrong for some reason, the INSERTs in the code only affect dhs_* tables, which are currently all empty.

I think that it'll be easier to judge whether it's done the right thing by browsing the database after the script has run and comparing it to the JSON.

PR #2064, opened 2016-07-09
Migrate choose_die_values to new action_log format

### LINKED ISSUE #2029: action log loses option die settings when a button has multiple identical option dice
It's obvious that this must be the case if you look at the code, but see any Lady Luck game, e.g. http://www.buttonweavers.com/ui/game.html?game=13278 for a live example.

cgolubi1:
* Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/471/
* Partially addresses #1991 
* Fixes #2029 

As mentioned in the code comments, it would be possible for the swing values to continue to be represented as a dict (with keys like 'X' and values like 13), but fixing #2029 meant that the option values couldn't be represented as a dict (that representation was the bug), so i decided to go with array-of-dicts for both swing and option.  There are lots of other places in the code where swing and option settings are represented differently as data structures (e.g. in the API), but from an action log point of view, setting 'X' to 13 is the same kind of thing as setting 'o(7/13)' to 13, so i thought they should be parallel.

Also of note: there are a small number of old entries of type choose_swing in the live site action logs, and there's still code to handle those.  Once this is merged, i'll submit another pull to migrate those guys over to choose_die_values in the DB and strip the choose_swing code.


blackshadowshade:
Okay, I've tested this on my WAMP stack with:
- The Fictitious Alan Clark vs ifni
- Social Class vs Gordo
- Lady Luck vs Lady Luck

It all looks fine, and the database entries look reasonable.

PR #2065, opened 2016-07-10
Implement the concept of private chat in the backend and chat-loading layers

cgolubi1:
* Partially addresses #464 
* Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/472/

This pull adds a private chat flag to the game_player_map database table, and implements the concept that, if that flag is set for either player in a game, non-players cannot load the chat for that game, but instead load, as part of the normal chat stream, a message saying that game chat is private.

This pull does not provide support for setting the private chat flag, so it does not make any chat privacy functionality available to players.

This pull changes the behavior of continuation chat in all games per https://github.com/buttonmen-dev/buttonmen/issues/464#issuecomment-231558844 option 4 --- non-players no longer see chat from previous games when they load a game which is a continuation, but do still see the "Continued from game N" message (both of those things hold regardless of the chat privacy settings on either the original or continuation games).

### DIFF
--- a/src/engine/BMInterface.php
+++ b/src/engine/BMInterface.php
@@ -165,8 +165,8 @@ class BMInterface {
                 $data['gameActionLogCount'] = $actionLogArray['nEntries'];
             }
 
-            $chatLogArray = $this->game_chat()->load_game_chat_log($game, $logEntryLimit);
-            if (empty($actionLogArray)) {
+            $chatLogArray = $this->game_chat()->load_game_chat_log($playerId, $game, $logEntryLimit);
+            if (empty($chatLogArray)) {
                 $data['gameChatLog'] = NULL;
                 $data['gameChatLogCount'] = 0;
             } else {
@@ -284,7 +284,8 @@ class BMInterface {
                  'UNIX_TIMESTAMP(v.last_action_time) AS player_last_action_timestamp, '.
                  'v.was_game_dismissed, '.
                  'v.has_player_accepted, '.
-                 'v.is_on_vacation '.
+                 'v.is_on_vacation, '.
+                 'v.is_chat_private '.
                  'FROM game AS g '.
                  'LEFT JOIN game_status AS s '.
                  'ON s.id = g.status_id '.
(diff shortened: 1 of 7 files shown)


cgolubi1:
I did some testing of this by manually setting the chat privacy flag to 1 for player c2 of game 4.  Game 5 is a continuation of game 4, and has no chat privacy set (so it is public).  Here are some screencaps from those games:

Private game 4, from the player's POV:
game_4_as_player

Private game 4, from the non-player's POV:
game_4_as_nonplayer

Public game 5, from the player's POV:
game_5_as_player

Public game 5, from the non-player's POV:
game_5_as_nonplayer

src/engine/BMInterface.php
```diff
@@ -165,8 +165,8 @@ public function load_api_game_data($playerId, $gameId, $logEntryLimit) {
                 $data['gameActionLogCount'] = $actionLogArray['nEntries'];
             }
 
-            $chatLogArray = $this->game_chat()->load_game_chat_log($game, $logEntryLimit);
-            if (empty($actionLogArray)) {
+            $chatLogArray = $this->game_chat()->load_game_chat_log($playerId, $game, $logEntryLimit);
+            if (empty($chatLogArray)) {
```
cgolubi1: This seemed like an obvious copy/paste-o bug, so i just fixed it.  It couldn't have ever come up yet because we don't yet support chatting any time before there are action log entries (and also `load_game_chat_log()` always returns something nontrivial anyway), but i figured i'd fix it just for cleanliness.

test/src/api/responderTest.php
```diff
@@ -3194,6 +3194,7 @@ public function test_api_game_002() {
 
     /**
      * @depends test_request_savePlayerInfo
+     * @group fulltest_deps
```
blackshadowshade: You might want to remove this debug flag

cgolubi1:
Done.  Need a new jenkins run, or enh?

blackshadowshade:
I can't seem to turn on chat privacy for the continuation game, no matter what I do with the chat privacy booleans in the databes.

blackshadowshade:
Hold that thought --- I seem to be having problems with my database connection on my WAMP stack.

blackshadowshade:
Okay, now I fixed the database connection, the chat privacy is back to working on the continuation game.

blackshadowshade:
Okay, this looks fine, tested with Bauer vs Hannah on a standard game with multiple chats and a reverse game. Viewed with both players and an extra player.

Issue #2070, opened 2016-07-11
Comment on new die backgrounds


blackshadowshade:
Die backgrounds were introduced in #2055. Here's a place where we can comment on them.

To start the ball rolling, I have the impression that the numbers are slightly below the midline in the case of the symmetric images, as viewed on Firefox.

blackshadowshade:
This also appears to be the case with the circle and the realistic images.

blackshadowshade:
On Chrome, this appears to be less of a problem, if at all.

blackshadowshade:
The tops and sides of the circles appear to be partly transparent in Chrome and Firefox. I presume this has something to do with antialiasing or scaling?

Hurtay:
Both in Firefox and Safari (ipad) the numbers in the new dice-shapes are as off-center as the numbers in the circles always have been before. 
So far I can't see the transparency in the circles.
At home I already got different opinions about which shapes look the best and are easiest to recognize. So it's a good thing we decided for user preferences.

AdmiralJota:
I think the images _themselves_ are off-center with each other. For instance, this is what I'm seeing in Firefox for http://dev.buttonweavers.com/ui/game.html?game=2481:

![bm1](https://cloud.githubusercontent.com/assets/5848173/16733222/ae82a13c-474e-11e6-94ba-8f06e2a2113b.png)

The tops of the images line up, but the bottoms don't. E.g., the bottom of the 4-sided die is notably higher up than the bottom of the 8-sider next to it.

The images should be aligned so that they're all vertically centered within their 72px canvases; shorter images should have empty space at both the top _and_ the bottom, not just the bottom.

Hurtay:
I don't see an easy way to change the images as you suggested. But feel free to try it yourself. I'm curious to what you'll come up with.
Right now they use up the space of their canvases as much as possible without having to change the shape. 

In your example the bottom line of the d4 is the maximum width. I could move the whole shape down but then the number would not even be readable any more since it's too high up in the shape. The die shape might be centered then but the number IN the die would be even less centered. 
The same with the d8: both diameters are the maximum possible size. If you'd like to move it up or down you'd have to decrease the size of the whole die shape. And I don't expect this to solve the original problem of easier size recognition.
And if you'd like all dice to have the same top line AND bottom line some of them would be really out of shape, i.e. the d4 which would look more like an A instead of an isosceles triangle.
What do you think?

blackshadowshade:
[...]

PR #2077, opened 2016-07-16
Made die value number smaller, added specific vertical spacing for Firefox

blackshadowshade:
Partially addresses #2070.

This changes the font size from 24 px to 20 px, and adds a Firefox-specific CSS rule that has different vertical spacing, to account for the unusual amount of padding at the bottom of the die value span that Firefox adds for some odd reason.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1199/ (if it passes)

### DIFF
--- a/src/ui/gui.css
+++ b/src/ui/gui.css
@@ -287,16 +287,8 @@ select.center {
 .die_overlay {
     position:relative;
     font-size: 21px;
-    top: 14px;
-}
-
-/* Added slightly different vertical spacing for Firefox only */
-@-moz-document url-prefix() {
-    .die_overlay{
-        position:relative;
-        font-size: 21px;
-        top: 17px;
-    }
+    line-height: 25px;
+    top: 13.5px;
 }
 
 .die_number_player {
--- a/src/ui/gui.css
+++ b/src/ui/gui.css
@@ -286,8 +286,17 @@ select.center {
 
 .die_overlay {
     position:relative;
-    font-size: 24px;
-    top: 13px;
+    font-size: 21px;
+    top: 14px;
+}
+
+/* Added slightly different vertical spacing for Firefox only */
+@-moz-document url-prefix() {
+    .die_overlay{
+        position:relative;
+        font-size: 21px;
+        top: 17px;
+    }
 }
 
 .die_number_player {


cgolubi1:
Hmm, do we think having a Firefox-specific override is the right approach?  I'm a little worried about going down the path of having to maintain a bunch of different CSS settings for a bunch of different browsers.  So i'm going to be looking for how bad the default Firefox experience really is, and whether it seems worth overriding, by looking at both types.

Also: your description says you're changing the font size to 20 px, but the CSS file says 21 px.  Which did you mean?

blackshadowshade:
Apparently 21 px.  :)

blackshadowshade:
If anyone has a better solution for the Firefox problem, I'm open to it, but no one has suggested anything to date. I also spent two hours investigating today using Firebug and Google, and couldn't find any obvious reason why Firefox adds the extra space to the bottom of the die value span.

src/ui/gui.css
```diff
@@ -286,8 +286,17 @@ select.center {
...
+    top: 14px;
+}
+
+/* Added slightly different vertical spacing for Firefox only */
+@-moz-document url-prefix() {
+    .die_overlay{
+        position:relative;
+        font-size: 21px;
+        top: 17px;
+    }
```
cgolubi1: Syntax question: do you actually have to redefine the things which are unchanged (position, font-size) in this block, or could you just define the thing which is an override?  If the latter is supported, it would be clearer IMO.

cgolubi1:
Anyway, it seems like all there is to do here is check this out in some browsers and see how it looks.

cgolubi1:
> couldn't find any obvious reason why Firefox adds the extra space to the bottom of the die value span.

FTR, comparing firefox to safari, i think it's not that Firefox adds extra space at the bottom, it's that Firefox *doesn't* have space at the top which most browsers have.

cgolubi1:
This post - http://stackoverflow.com/questions/14061228/remove-white-space-above-and-below-large-text-in-an-inline-block-element - pointed me to the idea of setting line-height.

Let me know what you think of:
```
.die_overlay {
    position:relative;
    font-size: 21px;
    line-height: 25px;
    top: 14px;
}
```

as a browser-independent solution.  It LGTM in firefox and safari, but my eyes may be crossing by this point.  :>)

src/ui/gui.css
blackshadowshade: I'm pretty sure that I could have left off the repeated stuff, but I didn't test with it. Anyway, hopefully we won't need this override, based on your suggestion below. Testing now.

blackshadowshade:
Yeah, that solves the problem on Firefox, and looks good on Safari and Chrome.

blackshadowshade:
FYI, I've shifted the numbers up another half pixel, which seems to centre them better across all three alternatives.

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1204/ (if it passes)

cgolubi1:
Okay.  This looks reasonable enough in firefox and safari - let's see what the dev testers think.

PR #2078, opened 2016-07-16
Fixed trip mighty bug when targeting a konstant die

### LINKED ISSUE #2076: mighty trip not allowed against konstant defender
So... this trip attack should be allowed, shouldn't it?

trip_failure

If everyone agrees, i'll make a responderTest for it.

blackshadowshade:
Fixes #2076.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1200/ (if it passes)

### DIFF
--- a/src/engine/BMAttackTrip.php
+++ b/src/engine/BMAttackTrip.php
@@ -153,7 +153,13 @@ class BMAttackTrip extends BMAttack {
             return TRUE;
         }
 
-        if ($def->has_skill('Konstant') && ($att->max < $def->value)) {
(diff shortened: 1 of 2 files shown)


cgolubi1:
Looks straightforward.  I'll want to do some replay testing as usual (so i'll have to partly get my test rig organized), but this looks to me like it should work assuming we understood what the bug is.

cgolubi1:
Okay, i replayed 1500 games and played 100 novel games.  Given that the test passes and that this should be the right fix, i think that's plenty.  Merging.

PR #2080, opened 2016-07-17
Fixed JSDoc comments

### LINKED ISSUE #2079: Fix JSDoc comments
Currently, the JSDoc comments have an incorrect syntax, which is causing NetBeans to show me loads of warnings.

As an example, we have something like

    @param  object die      The die whose recipe is to be displayed

This should actually be

    @param  {object} die      The die whose recipe is to be displayed

See http://usejsdoc.org/tags-param.html

Also

    @return  object

should actually be

    @returns  {object}

See http://usejsdoc.org/tags-returns.html

blackshadowshade:
Fixes #2079.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1202/ (if it passes)

### DIFF
--- a/src/ui/js/Game.js
+++ b/src/ui/js/Game.js
@@ -2699,11 +2699,11 @@ Game.gamePlayerStatus = function(player, reversed, game_active) {
  * The return object contains a boolean saying whether the die is clickable,
  * and a string describing the reason why or why not.
  *
- * @param  object  die            The die to be displayed
- * @param  boolean die_status     Status of the die ('active' or 'captured')
- * @param  string  player         Whose is the die? ('player' or 'opponent')
- * @param  boolean player_active  Is the player displaying the die active?
- * @return object
+ * @param  {object}  die            The die to be displayed
+ * @param  {boolean} die_status     Status of the die ('active' or 'captured')
+ * @param  {string}  player         Whose is the die? ('player' or 'opponent')
+ * @param  {boolean} player_active  Is the player displaying the die active?
+ * @returns {object}
  */
 Game.dieClickableInfo = function(die, player, die_status, player_active) {
   if (die_status == 'captured') {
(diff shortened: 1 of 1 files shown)


src/ui/js/Game.js
```diff
@@ -2699,11 +2699,11 @@ Game.gamePlayerStatus = function(player, reversed, game_active) {
...
- * @param  object  die            The die to be displayed
- * @param  boolean die_status     Status of the die ('active' or 'captured')
- * @param  string  player         Whose is the die? ('player' or 'opponent')
- * @param  boolean player_active  Is the player displaying the die active?
- * @return object
+ * @param  {object}  die            The die to be displayed
+ * @param  {boolean} die_status     Status of the die ('active' or 'captured')
+ * @param  {string}  player         Whose is the die? ('player' or 'opponent')
+ * @param  {boolean} player_active  Is the player displaying the die active?
+ * @returns {object}
```
cgolubi1: This is nit-level, but there were previously two spaces after each param so that the param types would line up with the return types.  Can you add a third space after each param so they line up with the 'returns' types?

src/ui/js/Game.js
```diff
@@ -3019,10 +3020,10 @@ Game.createDieContainerDiv = function(
  * Return a game-mat-style display of all dice belonging to the
  * requested player
  *
- * @param  string  player        Whose dice to display ('player' or 'opponent')
- * @param  boolean player_active Is the game awaiting action from the player
- *                               who is loading the page?
- * @return object                jQuery containing the div of dice
+ * @param  {string}  player        Whose dice to display ('player' / 'opponent')
```
cgolubi1: Hmm.  What's wrong with "or"?

blackshadowshade: I'm running up against the 80 character limit here. Would you prefer 'or' with less aligned spacing or the '|' symbol?

src/ui/js/Game.js
blackshadowshade: Here, I can add a third space, but in later comments, I start running up against the 80 character limit.

blackshadowshade:
Okay, I've prioritised alignment and shortened some lines by using the '|' symbol to stay within the 80 character limit.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1206/ (if it passes)

cgolubi1:
Yeah, sorry, i forgot the spacing was going to bite you.  Probably we should deprioritize alignment or normalize line-wrapping, but this is fine for now.

Issue #2081, opened 2016-07-17
players can't load overview if they have an active game with emoji in the description


cgolubi1:
Game 14996 on the prod site has a game description containing emoji.  This game could be created, but `loadActiveGames` returns a `null` in `gameDescriptionArray` when the game is loaded.  The UI function `Overview.addDescCol()` can't handle a null description, and blows up, leading the overview not be rendered.

Apply a fix to the prod site immediately so players can load the overview page, and apply a long-term fix so that we can't get into this state in which a player can create a game with a description that can't be loaded.

cgolubi1:
Okay, here's my immediate-term patch for Overview.js:
```diff
$ cat ~/2081_overview_js.patch 
diff --git a/src/ui/js/Overview.js b/src/ui/js/Overview.js
index fbc57ac..5cb5e68 100644
--- a/src/ui/js/Overview.js
+++ b/src/ui/js/Overview.js
@@ -422,10 +422,14 @@ Overview.addScoreCol = function(gameRow, gameInfo) {
 };
 
 Overview.addDescCol = function(gameRow, description) {
+  var descText = '';
+  if (typeof(description) == "string") {
+    descText = description.substring(0, 30) +
+               ((description.length > 30) ? '...' : '');
+  }
   gameRow.append($('<td>', {
     'class': 'gameDescDisplay',
-    'text': description.substring(0, 30) +
-            ((description.length > 30) ? '...' : ''),
+    'text': descText,
   }));
 };
 
```

cgolubi1:
```bash
cd ~/src/git/buttonmen-dev/buttonmen-dev
git status
patch -p1 < ~/2081_overview_js.patch
cd deploy/vagrant
vagrant status
vagrant provision
```

My dev site UI still loads.  So that's good.

cgolubi1:
```bash
cd ~/src/git/buttonmen-dev/buttonmen-prod
git status
patch -p1 < ~/2081_overview_js.patch
cd deploy/vagrant
vagrant status
vagrant provision
```

My prod site UI still loads, and all of my active games still have descriptions

cgolubi1:
[...]

PR #2082, opened 2016-07-17
Overview.addDescCol() should not blow up if description is null

cgolubi1:
* Mitigates the effect of #2081 
* Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/475/ (if it passes)

This is just the fix i already manually applied to the dev and prod sites.  Obviously the backend mismatch is what we need to fix, but regardless it's not bad for the UI to be resilient if the backend gives it bad data.

### DIFF
--- a/src/ui/js/Overview.js
+++ b/src/ui/js/Overview.js
@@ -423,7 +423,7 @@ Overview.addScoreCol = function(gameRow, gameInfo) {
 
 Overview.addDescCol = function(gameRow, description) {
   var descText = '';
-  if (typeof(description) == "string") {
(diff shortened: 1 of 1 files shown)


blackshadowshade:
So, JSHint failed here because of the use of double quotes.

cgolubi1:
Thanks.  New jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/477/ (if it passes)

blackshadowshade:
Okay, Jenkins is happy, and this just formalises what we've implemented as an interim solution, so I'll merge this.

PR #2084, opened 2016-07-22
final: Removed word 'Game' from the first column on the overview page

### LINKED ISSUE #1955: 5-digit game numbers causing two-line-high rows on Overview screen
Recently the Overview page got a lot less dense, because 5-digit game numbers aren't fitting on a single line, yuk:
screen shot 2016-02-21 at 11 12 58 pm

blackshadowshade:
Fixes #1955.

Testing pull request: #2107.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1210/

### DIFF
--- a/src/ui/js/Overview.js
+++ b/src/ui/js/Overview.js
@@ -378,7 +378,7 @@ Overview.addGameCol = function(gameRow, gameInfo, gameType) {
   gameLinkTd.append($('<a>', {
     'href': 'game.html?game=' + gameInfo.gameId,
     'text': Overview.linkTextStub(gameInfo, gameType) +
-            ' Game ' + gameInfo.gameId,
+            ' ' + gameInfo.gameId,
   }));
 
   gameRow.append(gameLinkTd);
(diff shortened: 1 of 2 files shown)


irilyth:
This looks ok to me. One suggestion: I think it'd look better without the "#" in the column header, i.e. just "Game" rather than "Game #". Anyone else have a preference?

If no one else cares enough to change it, or if anyone prefers to keep the "#", I think it's fine to shipit as is.

dwvanstone:
My overview page is almost always too wide and thus everything word wraps. This looks great everywhere except for "Game #" where the "#" wraps to the second line and does seem funny. Also, it's not as if the contents of the column is the game number - it is the action to be taken for the game. So I agree that the "#" can disappear without issue.

blackshadowshade:
So, what do others think? Does the title "Game" (without the #) work for the first column? Or do we change it to something like "Action"?

AdmiralJota:
It's fine with me.

cgolubi1:
Okay, the hash symbol removal has been merged onto the dev site, so folks should weigh in on whether this is good to go.

irilyth:
This looks good to me. Shipit!

cgolubi1:
Merging to master now.

PR #2085, opened 2016-07-25
Stop Dead Guy from turning up in random matches

### LINKED ISSUE #2032: Exclude completely un-fun buttons from random
As discussed on the forums, it really sucks to get Dead Guy as your random button, since he doesn’t get to do anything most of the time.

It would be fairly easy to add a flag to Dead Guy to have him excluded from random selection.

Are there any other buttons that need excluding? (I can’t think of any.)

As a side note, I’d suggest implementing this as a many-to-many relation to a new table of button flags instead of a new column in the button table, because it would allow easy adding of other flags in the future. I have thoughts, but they’re better raised separately.

blackshadowshade:
Fixes #2032.

Testing pull requests: #2110, #2137.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1255/

### DIFF
--- a/deploy/database/data.button.sql
+++ b/deploy/database/data.button.sql
@@ -1364,3 +1364,11 @@ UPDATE button SET flavor_text="The Queen of Hearts is a megalomaniac who enjoys 
 UPDATE button SET flavor_text="(The Jabberwock has no flavor text. The Jabberwock needs no flavor text. You don't taste The Jabberwock, The Jabberwock tastes you.)" WHERE name="The Jabberwock";
 UPDATE button SET flavor_text="Tweedledum takes out his jealousy over his brother's way with the ladies by picking fights over the little things. He enjoys listening to poetry and, of course, beating people up. Tweedledee takes out his jealousy over his brother's superior intellect by breaking Tweedledum's toys. He enjoys reciting poetry and, of course, beating people up." WHERE name="Tweedledum+dee";
 UPDATE button SET flavor_text="The White Rabbit dresses like a dandy and harbors a thinly-veiled animosity towards the opposite sex. He likes cucumbers, kid gloves, and beating people up." WHERE name="White Rabbit";
+
+INSERT INTO tag (name) VALUES ('exclude_from_random');
+
+INSERT INTO button_tag_map (button_id, tag_id)
+VALUES (
+  (SELECT id FROM button WHERE name = 'Dead Guy'),
+  (SELECT id FROM tag WHERE name = 'exclude_from_random')
+);
--- a/deploy/database/updates/02032_no_random_dead_guy_01.sql
+++ b/deploy/database/updates/02032_no_random_dead_guy_01.sql
@@ -1,0 +1,9 @@
+INSERT INTO tag (name) VALUES ('exclude_from_random');
+
+INSERT INTO button_tag_map (button_id, tag_id)
+VALUES (
+  (SELECT id FROM button WHERE name = 'Dead Guy'),
+  (SELECT id FROM tag WHERE name = 'exclude_from_random')
+);
+
+
--- a/src/api/ApiResponder.php
+++ b/src/api/ApiResponder.php
@@ -298,17 +298,20 @@ class ApiResponder {
      * @return NULL|array
      */
     protected function get_interface_response_loadButtonData($interface, $args) {
+        $buttonName = NULL;
+        $buttonSet = NULL;
+        $tagArray = NULL;
+
         if (isset($args['buttonName'])) {
             $buttonName = $args['buttonName'];
-        } else {
-            $buttonName = NULL;
         }
         if (isset($args['buttonSet'])) {
             $buttonSet = $args['buttonSet'];
-        } else {
-            $buttonSet = NULL;
         }
-        return $interface->get_button_data($buttonName, $buttonSet);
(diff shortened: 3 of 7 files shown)


cgolubi1:
This pull appears to add a `forceImplemented` argument to the public `loadButtonData` API function.  I don't think that's a good idea at all --- we don't want to allow customers on the prod site to load unimplemented buttons, because loading unimplemented buttons could cause internal errors or other site problems.  So if this change actually allows players to decide whether they want to see unimplemented buttons, that's not going to be okay.

blackshadowshade:
So, the code here includes all buttons by default. Would it address your issue to change the default to only loading fully implemented buttons by default, and explicitly having to override this to load all buttons? Or are you also concerned about other potential user-written programs explicitly hitting the API to query unimplemented buttons?

cgolubi1:
RIght now this second, can people query the prod site via the API and get information about unimplemented buttons?  Or not?

blackshadowshade:
Okay, I think I've tracked down the bit of the code that I think is relevant.

BMInterface->assemble_button_data() has the following override for $forceImplemented:

        if ('production' == $site_type) {
            $forceImplemented = TRUE;
        }

So, it's not possible to query unimplemented buttons, even if the function receives an input parameter of $forceImplemented that is FALSE.

cgolubi1:
Okay, fair enough.  I guess let me ask my question another way: you've added a new parameter, `$forceImplemented`, to the public API.  Why have you added this parameter to the public API?  Under what circumstances do you think players will use it?  Why is adding a new parameter to the public API needed in order to stop Dead Guy from turning up in random matches?

(I think those are three separate questions, though it's possible that more than one of them has the same answer.)

blackshadowshade:
I think I remember now why I added the extra parameter. It was to simplify the old responder test, so that we didn't need to duplicate the logic of filtering out unimplemented buttons (the old lines 638 - 644 in responderTest.php), because it seemed to me at the time that filtering out Dead Guy was a similar sort of thing.

I'm happy to reinstate the duplicated logic and remove the extra API parameter for $forceImplemented.

blackshadowshade:
(I guess I was motivated by the comment on the old lines 626 - 632 to see if I could remove the hackiness of the test code.)

cgolubi1:
Yeah, please undo that --- i agree that the old responderTest logic is a little silly, but it worked, and maybe some later BMInterface refactors can give us a tidier interface to use.  Except in extreme cases, we shouldn't be changing the public API to make tests happy, though, and i don't think this is an extreme case.

blackshadowshade:
So, this pull request has become bloated because of the fact that master and dev are now different, and that pull request #2137 is now targeted at a more advanced view of the world.

I'd love some help in getting this pull request back to what it should be.

blackshadowshade:
Ignore that. I managed to fix things using git cherry-pick.  :)

irilyth:
Do you need more like code help or more like Git help?

blackshadowshade:
Is all good. It was git help that I needed, but I'm worked out what I needed to do.

blackshadowshade:
Okay, as agreed, I'll be creating a new pull request that attempts to maintain a common history with the testing pull requests, and I'll document what happens. If that works out, I'll close this pull request and we'll go with the new one.

blackshadowshade:
So, I've recreated the content of this pull request based on master by cherry-picking the two commits in #2110 and #2137.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1255/ (if it passes)

cgolubi1:
As noted on #2133 punting this for December until the new workflow is sorted.

cgolubi1:
Shadowshade and i discussed whether this is now trivial enough that it could be brought in for December, but in fact there's two issues --- one is verifying the git stuff, and the other is that this is still an API change, albeit a smaller one.  So: continuing to punt it.

cgolubi1:
Shadowshade: this is currently two commits and should be rebased down to one.

After that i'll test it using RandomAI, which presumably is patient enough to try enough games to be confident none of them are Dead Guy.  :>)

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1286/ (if it passes)

blackshadowshade:
Rebased to apply after the turbo commit.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1309/ (if it passes)

blackshadowshade:
This has been rebased to remove the conflict with the new Wonderland button descriptions.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1355/ (if it passes)

cgolubi1:
Sorry for dropping this; i'm standing up a replay site now, and i'll have it archive a few thousand random/random games and make sure they work and none of them is Dead Guy.  It may take me a day to get to debugging my replay tester to work on actual API `__random` --- usually i randomize button choice client-side.

cgolubi1:
Looks like this is closing in on 8200 replayed games.  Unless my math is wrong (this is effectively the birthday problem, yeah?), if Dead Guy can appear in replay games, it's vanishingly unlikely that it won't have showed up in one of these.  :>)  So i'm downloading the games and double-checking, but assuming the results from that are what i expect, i'll be able to merge this.

PR #2086, opened 2016-07-25
Added more PHPDoc comments

blackshadowshade:
Continues to address part of #1130.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1215/

### DIFF
--- a/src/api/ApiResponder.php
+++ b/src/api/ApiResponder.php
@@ -12,12 +12,21 @@
 class ApiResponder {
 
     // properties
-    protected $isTest;               // whether this invocation is for testing
+    /**
+     * whether this invocation is for testing
+     *
+     * @var bool
+     */
+    protected $isTest;
 
-    // functions which allow access by unauthenticated users
-    // For now, all game functionality should require login: only
-    // add things to this list if they are necessary for user
-    // creation and/or login.
+    /**
+     * functions which allow access by unauthenticated users
+     * For now, all game functionality should require login: only
+     * add things to this list if they are necessary for user
+     * creation and/or login.
+     *
+     * @var array
+     */
     protected $unauthFunctions = array(
         'createUser',
         'verifyUser',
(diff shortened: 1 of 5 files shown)


blackshadowshade:
So, I've added a rule to ignore the PMD warning that says that all functions returning a boolean must be named "is..." or "has..." because we're just using responder to redirect calls.

src/api/ApiResponder.php
```diff
@@ -132,6 +178,13 @@ protected function get_interface_response_joinOpenGame($interface, $args) {
         return $success;
     }
 
+    /**
+     * Interface redirect for selectButton
+     *
+     * @param BMInterface $interface
+     * @param array $args
+     * @return bool
```
cgolubi1: Huh, is it actually the case that some of these interface functions return NULL and some return FALSE?  We might want to standardize that at some point if so.

src/api/ApiResponder.php
```diff
@@ -232,11 +369,25 @@ protected function get_interface_response_loadPlayerName() {
...
         $result = $interface->player()->get_player_info($_SESSION['user_id']);
         return $result;
     }
 
+    /**
+     * Interface redirect for savePlayerInfo
+     *
+     * @param BMInterface $interface
+     * @param array $args
+     * @return NULL|FALSE|array
```
cgolubi1: Yeah, this can't be right.  I mean, i believe you that it's true, but it can't be desirable for there to be a difference between NULL and FALSE, can it?

cgolubi1:
Anyway, looks fine --- let me know what you think about whether we need to open an issue to standardize failure return codes for API functions.

blackshadowshade:
I think that there's certainly a good argument for standardising return codes. It's nice that the addition of documentation has highlighted this issue for us.  :)

cgolubi1:
> I think that there's certainly a good argument for standardising return codes. It's nice that the addition of documentation has highlighted this issue for us. :)

Oh, yeah, i'm not saying don't accept the pull --- just wondering if we should open a new issue before merging this, and since you did the research to find out what return values everything has, i figured you could speak to the current state of things.

blackshadowshade:
Okay, opening a new issue now.

cgolubi1:
Thanks!  Let's merge this.

PR #2091, opened 2016-07-27
Handle cookies which don't contain stay_logged_in yet

### LINKED ISSUE #2090: python client is broken against dev code in some cases (old sessions?)
There's some sort of transition problem involving the change to logins, the cutover, existing sessions, and the python client.

I was able to get this error message using a script that works against the prod site:
```python
$ ./w game_info dev 1
could not parse return:
Notice: Undefined index: stay_logged_in in /var/www/api/api_core.php on line 99
{"data":{"userName":"glassonion"},"message":null,"status":"ok"}
  File "/Users/chaos/src/git/buttonmen-dev/buttonmen-dev/tools/api-client/python/game_info", line 44, in <module>
    if not b.verify_login():
  File "/Users/chaos/src/git/buttonmen-dev/buttonmen-dev/tools/api-client/python/lib/bmapi.py", line 88, in verify_login
    if retval.status == 'ok':
AttributeError: 'bool' object has no attribute 'status'
```

But then i ran `b.login()`, and since i did that, everything works fine.

cgolubi1:
* Fixes #2090 
* Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/478/ (if it passes)

### DIFF
--- a/src/api/api_core.php
+++ b/src/api/api_core.php
@@ -95,8 +95,14 @@ function auth_session_exists() {
                               ':auth_key' => $auth_key));
         $resultArray = $query->fetchAll();
         if (count($resultArray) == 1) {
+            // update old cookies which don't have the stay_logged_in key
+            if (array_key_exists('stay_logged_in', $_COOKIE)) {
+                $stay_logged_in = (bool)$_COOKIE['stay_logged_in'];
+            } else {
+                $stay_logged_in = FALSE;
+            }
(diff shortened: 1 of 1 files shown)


blackshadowshade:
I can replicate this error on my MAMP server by deleting the stay_logged_in and the PHPSESSID cookies after logging in, then hitting the Overview button. Your fix resolves the issue and seems reasonable.

PR #2097, opened 2016-08-03
Die background is now passed as part of the game info, not the player info

### LINKED ISSUE #2094: Die background images should all display based on user preference
Currently, the type of die background image appears to be specified by the preferences of the active player in the game, instead of by the preferences of the player who is viewing the game.

blackshadowshade:
Fixes #2094 better than #2095.

I've intentionally created this one on the basis of #2095 so that you have the option of accepting #2095 first and then this one, or just accepting this one (and closing #2095).

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1222/ (if it passes)

### DIFF
--- a/test/tools/api-client/python/lib/test_bmapi.py
+++ b/test/tools/api-client/python/lib/test_bmapi.py
@@ -127,7 +127,7 @@ class TestBMClient(unittest.TestCase):
   def test_load_game_data(self):
     known_keys = [
       'activePlayerIdx', 'currentPlayerIdx', 'description',
-      'gameActionLog', 'gameActionLogCount',
+      'dieBackgroundType', 'gameActionLog', 'gameActionLogCount',
       'gameChatEditable', 'gameChatLog', 'gameChatLogCount',
       'gameId', 'gameSkillsInfo',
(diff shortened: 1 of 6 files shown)


test/tools/api-client/python/lib/test_bmapi.py
```diff
@@ -127,7 +127,7 @@ def test_create_game(self):
   def test_load_game_data(self):
     known_keys = [
       'activePlayerIdx', 'currentPlayerIdx', 'description',
-      'gameActionLog', 'gameActionLogCount',
+      'dieBackgroundType', 'gameActionLog', 'gameActionLogCount',
```
cgolubi1: Nit: this list is supposed to be alphabetized.  I'm not going to block the pull over it, though, it's a test config, whatever.

cgolubi1:
Okay, this looks simple enough to me, and it's the correct fix.  So let me go ahead and test it, and if it seems to check out, i'll merge it and skip #2095.

cgolubi1:
This looks like it solves the problem to me, and there are tests at every layer which reflect the changes.  I'm going to merge it.

PR #2104, opened 2016-08-09
Display UI messages when joining an open game

### LINKED ISSUE #1755: UI gives no feedback about whether joining an open game succeeded
It looks like joining an open game doesn't set `Env.message`, so you don't get any sort of green message-from-server when you join one.

The corollary is that if you _fail_ to join an open game, you also get no message from the server.  I tested this as follows:
- c1 created an open game
- c2 loaded the open games page
- c3 loaded the open games page
- c2 joined the open game
- c3 joined the open game

In this case, c3 should get a message from the server about failing to join the game (because it's no longer open), but instead http://localhost:8080/ui/open_games.html simply reloads with the message "There are no open games." (and if there were other open games, this would be even more confusing, because the page would just reload with no notification).

blackshadowshade:
Fixes #1755.

The message that is displayed causes the open games table to shift downwards, which I think is undesirable. It's probably better to always reserve the space for the message, but I don't know how to do this. Suggestions are welcome.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1226/ (if it passes)

### DIFF
--- a/src/ui/js/OpenGames.js
+++ b/src/ui/js/OpenGames.js
@@ -136,6 +136,12 @@ OpenGames.displayJoinResult = function(
         'type': 'error',
         'text': 'An internal error occurred while trying to join the game.',
       };
+    } else {
+      Env.message = {
+        'type': 'error',
+        'text': 'The list of open games was out of date. ' +
+                'It has been refreshed.',
+      };
     }
     OpenGames.getOpenGames(OpenGames.showPage);
     return;


AdmiralJota:
One option would be to test if the call was successful and only show the message if it failed. 

Shifting the table down seems like reasonable behavior in the case that there's an error the player needs to be informed of, and there's already visual feedback when the joining is successful.

blackshadowshade:
@cgolubi1, if this suggestion is acceptable to you, I'll remove the success message again, so that the table doesn't move when a game has been successfully joined.

cgolubi1:
This simple change has been moping around for a long time; let's stand up a dev site.

cgolubi1:
irilyth: can you setup transient == 54.224.62.17?  thanks.

cgolubi1:
Okay, so, here's what i see on success:

join_success

and here's what i see on failure:

join_failure

Can someone recap on the issue why having the table move around is a problem?  When we have a message to display, it makes other UI elements move down.  That's the way every other page we have that uses `Env.message` does this.  Why is it worse on the Open Games page?

AFAIC, it would be fine to just merge this.

irilyth:
> irilyth: can you setup transient == 54.224.62.17? thanks.

Done!

blackshadowshade:
In my opinion, it's worse here on the Open Games page because you may want to accept multiple games, one after another. With the move of the table, any quick click to accept a second game will fail, because the link will have moved down.

However, if people are fine with this move of the table, I'm okay with this being accepted as-is.

dwvanstone:
I find this new UI very jarring. When there is a long list of open games (enough to need to scroll) and I accept one of the games below the fold, the interface scrolls me to the top of the page to report "Successfully joined Game ###." This is disorienting. Furthermore, in order to click into the newly created game I have to scroll back down and click on "Go to Game ###". 

I could understand scrolling up to the top when there's an error, but not when there is a success. We already get the success feedback.

dwvanstone:
I don't think we need to support rapidly joining multiple games on the Open Games page. My main use case is to join a game and then click on the game link to go look at the game I've joined. (Especially if random buttons are involved.)

An alternative would be to have a feedback page similar to that found when creating a game. On the Open Games screen, clicking on a button "Join Game ###" could send the user to a new page that only gives the feedback (positive or negative):

"Game ### joined Successfully. LINK: Go to Game ###  LINK: Go/Return to Open Games page"

AdmiralJota:
My usual workflow on the Open Games page is to scroll down until I find a game I want to join, join it, and then continue scrolling down looking for more games I want to join. 

Given that, I'd find something that scrolled me back to the top or took me off the page to be fairly disruptive.

blackshadowshade:
Based on that feedback, I think I'll go and back out the success message again, and only show a message on failure. I'll look at doing that this weekend.

blackshadowshade:
I've backed out the success message. This should be ready for a dev site update once CircleCI passes.

cgolubi1:
Updated dev site.

dwvanstone:
It looks good to me now too.

cgolubi1:
Okay, great, declaring victory.  Shadowshade, can you rebase this for merge?

blackshadowshade:
Rebased.

PR #2120, opened 2016-10-20
testing: Added logic to check what sort of click is used on a pseudolink

blackshadowshade:
Proposed fix for #2119.

See discussion in http://www.buttonweavers.com/ui/forum.html#!threadId=406

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1235/

PR #2121, opened 2016-10-20
final: Added logic to check what sort of click is used on a pseudolink

### LINKED ISSUE #2119: <a class="pseudolink"> doesn't open correctly on middle-click
Taken from http://www.buttonweavers.com/ui/forum.html#!threadId=406

@ericeslinger said:

So, I'm not wild about the pseudolink code in the forums. It means I can't middle-click to open a link in a new tab, because the normal a href handler is overridden by the click handler.

Removing that clickhandler works (if I fiddle with it in the DOM inspector). If you're going to be doing stuff with the URL like that, it might be worthwhile to look at how some of the Big frameworks handle routing and state changes without window reloads. I know I can create functioning a href elements using angular.js that respect all native behaviors but are still just internal (no new network requests) state changes. It even remembers history correctly with the back button and all that.

blackshadowshade:
Fixes #2119.

Testing pull request: #2120.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1235/

### DIFF
--- a/src/ui/js/Forum.js
+++ b/src/ui/js/Forum.js
@@ -422,6 +422,13 @@ Forum.arrangePage = function() {
 // These are events that are triggered by user actions
 
 Forum.formLinkToSubPage = function(e) {
+  // Don't let confused browsers execute click events for things that 
+  // aren't proper clicks!
+  var button = (e.which || e.button);
+  if (button > 1 || e.ctrlKey || e.metaKey || e.shiftKey) {
+    return;
+  }
+
   e.preventDefault();
   var state = Forum.readStateFromElement(this);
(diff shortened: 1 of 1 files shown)


cgolubi1:
No one has commented on this issue.

irilyth:
This works for me in Chrome now: If I look at http://www.buttonweavers.com/ui/forum.html#!boardId=3, and middle click on the link that's the name of the thread, it opens in the same tab, but on http://dev.buttonweavers.com/ui/forum.html#!boardId=3 it opens in a new tab. It still works as before in Firefox. Shipit!

cgolubi1:
Merging to master now.

PR #2124, opened 2016-10-29
final: Changed Cheese Weasel to having a button special

### LINKED ISSUE #2115: class Cheese Weasel as a button with unimplemented skills
**Cheese Weasel has a special rule which is not implemented.** I think we can all agree that this button should be left on the regular site (and _not_ moved to dev site only), but **it might make sense to class him as a button with unimplemented skills** (similar to how Bruno has been handled).

This is not a request to implement the rule (which is covered in issue #1659) but rather a simple request to have Cheese Weasel's button page read:

> The special button skills for this button have not yet been implemented.

However, for reference, I will include Cheese Weasel's special rule here:

> Once per round, before taking his turn, the Cheese Weasel may re-roll one of his own dice that shows a value matching one of his opponent's dice, OR he may re-roll one of his opponent's dice that shows a value matching one of his own.

this was discussed in this thread:
http://www.buttonweavers.com/ui/forum.html#!threadId=395

blackshadowshade:
Fixes #2115.

Testing pull request: #2123.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1236/


irilyth:
Looks fine to me. (I guess we don't generally say what the unimplemented special abilities are? It looks like we don't, e.g. for Bruno or Pappy.) Shipit!

cgolubi1:
Merging to master now.

Issue #2132, opened 2016-11-06
push code to dev.buttonweavers.com for December 2016 release


cgolubi1:
This is the issue for loading code onto dev for the release.  The issue for deciding which pulls to bring to staging is #2133.

As features are merged into the dev codebase branch, redeploy dev.buttonweavers.com with those changes.

Generic procedures for use in deployments:

Use a checked-out copy of buttonmen-dev to find a sort order for database updates:

``` bash
cd deploy/database/updates/
UPDATES="<space-separated list of DB updates>"
git log --reverse --first-parent --oneline $UPDATES
UPDATES="<space-separated list of DB updates, now in the order reported by that output>"
```

Database update plan:

``` bash
# in the correct git working directory
git pull origin dev
cd deploy/vagrant
vagrant provision

# on the sandbox or dev VM
UPDATES="<space-separated ordered list of DB updates, as determined above>"

sudo -u mysql mkdir /tmp/pre /tmp/post
ls /tmp/pre /tmp/post
sudo -u mysql mysqldump -u root --skip-comments -T /tmp/pre buttonmen
cd /buttonmen/deploy/database/updates
for update in $UPDATES; do
  echo "APPLYING: $update"
  cat $update | mysql -u root buttonmen
done

sudo -u mysql mysqldump -u root --skip-comments -T /tmp/post buttonmen
```

How to diff database dumps:

``` bash
diff -r --brief /tmp/pre /tmp/post
```
#217 is the master ticket linking all dev code pushes.

cgolubi1:
Okay, i'm going to deploy dev now with all the above pulls.

```
UPDATES="02032_no_random_dead_guy_01.sql 02115_cheese_weasel.sql 01978_steven_universe_01.sql 02122_perf.sql"

sudo -u mysql mkdir /tmp/pre /tmp/post
ls /tmp/pre /tmp/post
sudo -u mysql mysqldump -u root --skip-comments -T /tmp/pre buttonmen
cd /buttonmen/deploy/database/updates
for update in $UPDATES; do
  echo "APPLYING: $update"
  cat $update | mysql -u root buttonmen
done

sudo -u mysql mysqldump -u root --skip-comments -T /tmp/post buttonmen
```

Testing on virtualbox looks fine - i don't see any issues with database diffs.  Going to load on the actual dev site now.

cgolubi1:
I'm going to deploy dev now with the above pull, which should be a noop.

[...]

PR #2134, opened 2016-11-07
testing: Define puppet_hostname for new IP for staging.buttonweavers.com so its mail will be accepted by more servers

cgolubi1:
* Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/480/ (if it passes)

This code should have no effect on dev.

### DIFF
--- a/deploy/vagrant/manifests/init.pp
+++ b/deploy/vagrant/manifests/init.pp
@@ -17,6 +17,9 @@ node default {
     "54.83.36.215": {
       $puppet_hostname = "dev.buttonweavers.com"
     }
+    "54.235.150.227": {
+      $puppet_hostname = "staging.buttonweavers.com"
+    }
     "54.83.36.209": {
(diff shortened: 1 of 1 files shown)


blackshadowshade:
Okay, IP address looks right. Let's try this once Jenkins passes.

cgolubi1:
This will be loaded to dev as part of #2132

irilyth:
The syntax looks correct here, and I verified the address:

+$ host staging.buttonweavers.com
staging.buttonweavers.com has address 54.235.150.227

Note that the address  just has the default AWS PTR record:

+$ host 54.235.150.227
227.150.235.54.in-addr.arpa domain name pointer ec2-54-235-150-227.compute-1.amazonaws.com.

It should probably have a PTR record pointing to staging.buttonweavers.com, like the addresses for dev and www have.

That isn't a blocker for this PR, though, so I say: Shipit!

PR #2135, opened 2016-11-07
final: Define puppet_hostname for new IP for staging.buttonweavers.com so its mail will be accepted by more servers

cgolubi1:
Testing pull request: #2134.

* Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/480/ (if it passes)

I intended to load this onto staging as soon as we've verified that it's harmless on dev, so that people can more easily create accounts on staging and receive their verification e-mails.


cgolubi1:
No one has commented on this.  I think we should release this to staging so people can verify their accounts on staging.  :>)

cgolubi1:
Merging to master now.

PR #2110, opened 2016-08-23
testing: Stop Dead Guy from turning up in random matches

blackshadowshade:
Proposed fix for #2032.

In this pull request, I implement generic filtering (both positive and negative) by tag, and then use a negative filter to exclude Dead Guy from random matches.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1231/ (if it passes)

PR #2137, opened 2016-11-07
testing: Removed $forceImplement flag from API

blackshadowshade:
This addresses comments made in #2085, and depends on #2110.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1251/

### DIFF
--- a/src/api/ApiResponder.php
+++ b/src/api/ApiResponder.php
@@ -300,7 +300,6 @@ class ApiResponder {
     protected function get_interface_response_loadButtonData($interface, $args) {
         $buttonName = NULL;
         $buttonSet = NULL;
-        $forceImplemented = FALSE;
         $tagArray = NULL;
 
         if (isset($args['buttonName'])) {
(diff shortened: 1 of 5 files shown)


cgolubi1:
There's no "final" version of this pull.  What's up with that?

blackshadowshade:
The final here has been rebased into #2085.

This raises the issue, which we'll probably need to discuss in IRC, of how we deal with follow-up pull requests that alter previous pull requests, while still maintaining the one-pull-request-per-feature requirement.

cgolubi1:
> The final here has been rebased into #2085.

That doesn't make sense to me --- how could that be given that #2110 and #2137 are different branches?

If they were the same branch as each other, this would make total sense.  If they were the same branch, the workflow would be:
* merge testing pull 1 with commit A into dev
* look at testing pull 2 with commit B; verify it is on the same branch and the final PR shows both commits A and B
* merge testing pull 2 into dev
* rebase the final pull so it now has a single commit C containing both A and B

That workflow seems totally reasonable to me.  But that's not what's happening here.  Shadowshade asserts (i think) that commit 76de17a on #2085 is in fact the rebase of commit a2d11e9 from #2110 and commit f9ad8a3 from #2137.  But this seems pretty tedious to verify.

And it looks like shadowshade munged up the 2032_no_random_dead_guy branch in the process of sorting this out, so we no longer have that branch (i.e. we can't simply build on what was merged in #2110 because that state has been lost).

So: i am kind of asserting that you've come up with a process i don't understand, and in the process made a mess, such that i don't think we can have a lot of confidence in the code in this branch.  And we may want to find a way to start over on this one, but, at any rate, if i understand the current situation correctly, i don't think it is a good situation.

Issue #2072, opened 2016-07-13
Make skill descriptions available on mobile devices


blackshadowshade:
This issue has been spun off #1697.

The major issue here is that the hover event isn't available on mobile devices, so it's difficult to access skill descriptions---it is possible, if you click on the link to the button description, but it would be far better to have the skill description available on the page showing the game mat.

blackshadowshade:
One major consideration that has arisen out of the discussion on #1977 is that we shouldn't be introducing an unintentional regression into the current behaviour of skill hints for desktop users.

cgolubi1:
I'm happy to play with some ideas for this --- see #1977 for a couple of things i came across that seem promising.

slugmuffin:
As a new user trying to play on mobile, this feature would be great to have. 

With the Kickstarter now running, there could be more users on the way.

blackshadowshade:
I'll note that there have been further discussions on this at http://www.buttonweavers.com/ui/forum.html#!threadId=1233 and on the Discord server.

I think the easiest solution is to make the skill names clickable links that link appropriately to the correct part of the skills page.

Other solutions that were investigated in #1977 are still an option, some of those looked promising.

Issue #943, opened 2014-06-10
More consistent die selection page


irilyth:
On the die selection page, the text and order of things is somewhat inconsistent for complicated buttons. See this screenshot including CactusJack, for example:

![button-men](https://cloud.githubusercontent.com/assets/4923138/3225078/6b998e86-f047-11e3-9502-b28cdb6c165d.png)

In the recipe, the option dice are listed before the swing dice, but on the selection screen, the swing dice are first. Also, the skills are listed by the option dice, but not the swing dice. I think it'd be better if they were in the same order, and if the skills were always listed.

PR #2152, opened 2017-01-16
Added Turbo, take 2

### LINKED ISSUE #91: Turbo swing skill needed

### LINKED ISSUE #2156: Turbo interactions with other skills
We've put the initial iteration of turbo up for testing, see #2152.

Bugs and issues about the interaction of turbo with other skills should be listed here for discussion.

blackshadowshade:
Available for testing at: http://54.160.177.210/

Fixes #91. Fixes #2156.

Requires database update 00091_turbo_01.sql

For other comments, see #2048.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1296/

### DIFF
--- a/src/engine/BMGame.php
+++ b/src/engine/BMGame.php
@@ -270,33 +270,43 @@ class BMGame {
      */
     protected $fireCache;
 
+    /**
+     * Internal cache of turbo size info, used to store turbo size selections
+     * when the game pauses in BMGameState::ADJUST_FIRE_DICE
+     */
+    protected $turboCache;
+
     // methods
     /**
      * This is a generic caller function that calls each
      * do_next_step_*() function, based on the current game state.
+     *
+     * @param array $args
      */
-    public function do_next_step() {
+    public function do_next_step(array $args = array()) {
         if (!isset($this->gameState)) {
             throw new LogicException('Game state must be set.');
         }
 
         $funcName = 'do_next_step_'.
                     strtolower(BMGameState::as_string($this->gameState));
-        $this->$funcName();
+        $this->$funcName($args);
     }
 
     /**
      * This is a generic caller function that calls each
      * update_game_state_*() function, based on the current game state.
+     *
+     * @param array $args
      */
-    public function update_game_state() {
+    public function update_game_state(array $args = array()) {
         if (!isset($this->gameState)) {
             throw new LogicException('Game state must be set.');
         }
 
         $funcName = 'update_game_state_'.
                     strtolower(BMGameState::as_string($this->gameState));
-        $this->$funcName();
+        $this->$funcName($args);
     }
 
     /**
@@ -364,7 +374,7 @@ class BMGame {
             throw new LogicException(
                 'maxWins must be set before applying handicaps.'
             );
-        };
+        }
 
         $nWins = 0;
 
@@ -1160,8 +1170,10 @@ class BMGame {
 
     /**
      * Update game state from BMGameState::start_turn if necessary
+     *
+     * @param array $args
      */
-    protected function update_game_state_start_turn() {
+    protected function update_game_state_start_turn($args) {
         if ($this->isWaitingOnAnyAction()) {
             return;
         }
@@ -1171,7 +1183,7 @@ class BMGame {
             return;
         }
 
-        $instance = $this->create_attack_instance();
+        $instance = $this->create_attack_instance($args);
         if (FALSE === $instance) {
             $this->attack = NULL;
             return;
@@ -1180,7 +1192,7 @@ class BMGame {
         $this->add_flag_to_attackers();
         $this->add_flag_to_attack_targets();
 
-        $this->gameState = BMGameState::ADJUST_FIRE_DICE;
+        $this->gameState = BMGameState::CHOOSE_TURBO_SWING_FOR_TRIP;
     }
 
     /**
@@ -1216,9 +1228,11 @@ class BMGame {
      * Create a BMAttack instance that corresponds to the attack
      * specified in $this->attack
      *
+     * @param array $args
+     *
      * @return array|FALSE
      */
-    protected function create_attack_instance() {
+    protected function create_attack_instance(array $args = array()) {
         if (!isset($this->attack)) {
             $this->regenerate_attack();
         }
@@ -1249,11 +1263,17 @@ class BMGame {
             $attack->add_die($attackDie);
         }
 
+        $turboVals = array();
+        if (array_key_exists('turboVals', $args)) {
+            $turboVals = $args['turboVals'];
+        }
+
         $valid = $attack->validate_attack(
             $this,
             $attAttackDieArray,
             $defAttackDieArray,
-            $this->firingAmount
+            array('helpValue' => $this->firingAmount,
+                  'turboVals' => $turboVals)
         );
 
         if (!$valid) {
@@ -1335,9 +1355,45 @@ class BMGame {
     }
 
     /**
+     * Perform the logic required at BMGameState::choose_turbo_swing_for_trip
+     */
+    protected function do_next_step_choose_turbo_swing_for_trip() {
+        if ($this->attack['attackType'] != 'Trip') {
+            return;
+        }
+
+        $attackDice = $this->get__attackerAttackDieArray();
+
+        if (count($attackDice) != 1) {
+            throw new LogicException('There should only be one attacker for a trip attack');
+        }
+
+        $attackDie = $attackDice[0];
+
+        if ($attackDie->has_skill('Turbo') &&
+            $attackDie->doesReroll) {
+            $attacker = $this->playerArray[$this->attack['attackerPlayerIdx']];
+            $attacker->waitingOnAction = TRUE;
+        }
+    }
+
+    /**
+     * Update game state from BMGameState::choose_turbo_swing_for_trip
+     */
+    protected function update_game_state_choose_turbo_swing_for_trip() {
+        if (!$this->isWaitingOnAnyAction()) {
+            $this->gameState = BMGameState::ADJUST_FIRE_DICE;
+        }
+    }
+
+    /**
      * Perform the logic required at BMGameState::adjust_fire_dice
      */
     protected function do_next_step_adjust_fire_dice() {
+        if ($this->playerArray[$this->attack['attackerPlayerIdx']]->waitingOnAction) {
+            return;
+        }
+
         if ($this->needs_firing() || $this->allows_fire_overshooting()) {
             $this->playerArray[$this->attack['attackerPlayerIdx']]->waitingOnAction = TRUE;
         }
@@ -1383,7 +1439,7 @@ class BMGame {
             $this,
             $attackerDieArray,
             $defenderDieArray,
-            0
+            array('helpValue' => 0)
         ));
 
         if ($needsFiring) {
@@ -1459,7 +1515,7 @@ class BMGame {
             $this,
             $attackerDieArray,
             $defenderDieArray,
-            max(1, $defenderDie->value - $attackerDie->value + 1)
+            array('helpValue' => max(1, $defenderDie->value - $attackerDie->value + 1))
         )) {
             $this->log_action(
                 'allows_firing',
@@ -1596,7 +1652,7 @@ class BMGame {
             $this,
             $instance['attAttackDieArray'],
             $instance['defAttackDieArray'],
-            $firingAmount
+            array('helpValue' => $firingAmount)
         )) {
             $this->message = $instance['attack']->validationMessage;
             return FALSE;
@@ -1649,7 +1705,7 @@ class BMGame {
             $this,
             $instance['attAttackDieArray'],
             $instance['defAttackDieArray'],
-            0
+            array('helpValue' => 0)
         )) {
             $this->message = $instance['attack']->validationMessage;
             return FALSE;
@@ -1792,14 +1848,133 @@ class BMGame {
      * Perform the logic required at BMGameState::choose_turbo_swing
      */
     protected function do_next_step_choose_turbo_swing() {
+        if ($this->attack['attackType'] == 'Trip') {
+            return;
+        }
 
+        $attacker = $this->playerArray[$this->attack['attackerPlayerIdx']];
+        foreach ($attacker->activeDieArray as $activeDie) {
+            if ($activeDie->has_skill('Turbo') &&
+                $activeDie->has_flag('IsAttacker') &&
+                $activeDie->doesReroll) {
+                $attacker->waitingOnAction = TRUE;
+            }
+        }
+    }
+
+    /**
+     * Attempt to set the sizes of the attacking turbo dice
+     *
+     * @param array $turboSizeArray
+     * @return boolean
+     */
+    public function set_turbo_sizes(array $turboSizeArray) {
+        if (!$this->validate_turbo_sizes($turboSizeArray)) {
+            return FALSE;
+        }
+
+        $playerIdx = $this->attack['attackerPlayerIdx'];
+        $preTurboDice = array();
+        $postTurboDice = array();
+
+        // set turbo sizes
+        foreach ($turboSizeArray as $dieIdx => $dieSize) {
+            $die = $this->playerArray[$playerIdx]->activeDieArray[$dieIdx];
+
+            if (!$die->has_skill('Turbo')) {
+                $this->message = 'Cannot set turbo size for a non-turbo die';
+                return FALSE;
+            }
+
+            if (!$die->has_flag('IsAttacker')) {
+                $this->message = 'Only attacking turbo dice can set turbo size';
+                return FALSE;
+            }
+
+            $preTurboDice[] = $die->get_action_log_data();
+            $oldRecipe = $die->get_recipe(TRUE);
+            unset($die->value);
+            if (isset($die->dice)) {
+                foreach ($die->dice as $subdie) {
+                    unset($subdie->value);
+                }
+            }
+
+            try {
+                $die->setTurboSize($dieSize);
+            } catch (LogicException $e) {
+                $this->message = $e->getMessage();
+                return FALSE;
+            }
+
+            $die->roll();
+            $die->add_flag('HasJustTurboed', $oldRecipe);
+            $postTurboDice[] = $die->get_action_log_data();
+        }
+
+        $this->log_action(
+            'set_turbo',
+            $this->playerArray[$playerIdx]->playerId,
+            array(
+                'attackType' => $this->attack['attackType'],
+                'preTurboDice' => $preTurboDice,
+                'postTurboDice' => $postTurboDice
+            )
+        );
+
+        $this->playerArray[$playerIdx]->waitingOnAction = FALSE;
+
+        return TRUE;
+    }
+
+    /**
+     * Do some basic validation of the submitted turbo sizes
+     *
+     * @param array $turboSizeArray
+     * @return boolean
+     */
+    protected function validate_turbo_sizes(array $turboSizeArray) {
+        if ((BMGameState::CHOOSE_TURBO_SWING != $this->gameState) &&
+            (BMGameState::CHOOSE_TURBO_SWING_FOR_TRIP != $this->gameState)) {
+            $this->message = 'Wrong game state to choose turbo swing sizes.';
+            return FALSE;
+        }
+
+        $playerIdx = $this->attack['attackerPlayerIdx'];
+        $nDice = count($this->playerArray[$playerIdx]->activeDieArray);
+        foreach (array_keys($turboSizeArray) as $key) {
+            if (!is_int($key) || ($key < 0)) {
+                $this->message = 'The turbo size array must be keyed with non-negative integer die indices';
+                return FALSE;
+            }
+
+            if ($key > $nDice - 1) {
+                $this->message = 'The turbo size array seems to refer to a die index that does not exist';
+                return FALSE;
+            }
+        }
+
+        foreach ($this->playerArray[$playerIdx]->activeDieArray as $dieIdx => $die) {
+            if ($die->has_skill('Turbo') &&
+                $die->has_flag('IsAttacker') &&
+                !array_key_exists($dieIdx, $turboSizeArray) &&
+                !$die->has_flag('HasJustMorphed')) {
+                $this->message = 'Not all turbo values were specified';
+                return FALSE;
+            }
+        }
+
+        return TRUE;
     }
 
     /**
      * Update game state from BMGameState::choose_turbo_swing if necessary
      */
     protected function update_game_state_choose_turbo_swing() {
-        $this->gameState = BMGameState::END_TURN;
+        if (!$this->isWaitingOnAnyAction()) {
+            unset($this->turboCache);
+            $this->gameState = BMGameState::END_TURN;
+        }
     }
 
     /**
@@ -2007,16 +2182,21 @@ class BMGame {
      * (i) the game requires a player action, or
      * (ii) the game reaches a game state that signals that it has ended
      *
-     * The variable $gameStateBreakpoint is used for debugging purposes only.
-     * If used, the game will stop as soon as the game state becomes
-     * the value of $gameStateBreakpoint.
+     * The parameter 'gameStateBreakpoint' (as part of $args) is used for
+     * debugging purposes only. If used, the game will stop as soon as the
+     * game state becomes the value of $gameStateBreakpoint.
      *
-     * @param int $gameStateBreakpoint
+     * @param array $args
      */
-    public function proceed_to_next_user_action($gameStateBreakpoint = NULL) {
+    public function proceed_to_next_user_action(array $args = array()) {
         $repeatCount = 0;
         $initialGameState = $this->gameState;
-        $this->update_game_state();
+        $this->update_game_state($args);
+
+        $gameStateBreakpoint = NULL;
+        if (array_key_exists('gameStateBreakpoint', $args)) {
+            $gameStateBreakpoint = $args['gameStateBreakpoint'];
+        }
 
         if (isset($gameStateBreakpoint) &&
             ($gameStateBreakpoint == $this->gameState) &&
@@ -2024,18 +2204,18 @@ class BMGame {
             return;
         }
 
-        $this->do_next_step();
+        $this->do_next_step($args);
 
         while (!$this->isWaitingOnAnyAction()) {
             $tempGameState = $this->gameState;
-            $this->update_game_state();
+            $this->update_game_state($args);
 
             if (isset($gameStateBreakpoint) &&
                 ($gameStateBreakpoint == $this->gameState)) {
                 return;
             }
 
-            $this->do_next_step();
+            $this->do_next_step($args);
 
             if ($this->gameState >= BMGameState::END_GAME) {
                 return;
@@ -3442,6 +3622,7 @@ class BMGame {
                 'optRequestArray'     => $player->optRequestArray,
                 'prevSwingValueArray' => $player->prevSwingValueArray,
                 'prevOptValueArray'   => $player->prevOptValueArray,
+                'turboSizeArray'      => $this->get_turboSizeArray($playerIdx),
 // BMGame may lie about who's actually waiting
                 'waitingOnAction'     => $this->isWaitingOnAction($playerIdx, $requestingPlayerIdx),
                 'roundScore'          => $player->roundScore,
@@ -3902,7 +4083,7 @@ class BMGame {
      * @return array
      */
     protected function get_swingRequestArray($playerIdx, $requestingPlayerIdx) {
-        // do not show swing requests unless this information is actually necessary
+        // hide non-turbo swing requests when this information is not necessary
         if ($this->gameState > BMGameState::SPECIFY_DICE) {
             return array();
         }
@@ -3925,28 +4106,8 @@ class BMGame {
         }
 
         if (isset($swingRequestArrayArray[$playerIdx])) {
-            foreach ($swingRequestArrayArray[$playerIdx] as $swingtype => $swingdice) {
-                if ($swingdice[0] instanceof BMDieTwin) {
-                    if ($swingdice[0]->dice[0] instanceof BMDieSwing) {
-                        $swingdie = $swingdice[0]->dice[0];
-                    } elseif ($swingdice[0]->dice[1] instanceof BMDieSwing) {
-                        $swingdie = $swingdice[0]->dice[1];
-                    } else {
-                        throw new LogicException(
-                            'At least one of the subdice of a twin swing die should be a swing die'
-                        );
-                    }
-                } else {
-                    $swingdie = $swingdice[0];
-                }
-                if ($swingdie instanceof BMDieSwing) {
-                    $validRange = $swingdie->swing_range($swingtype);
-                } else {
-                    throw new LogicException(
-                        'Tried to put die in swingRequestArray which is not a swing die: ' . $swingdie
-                    );
-                }
-                $swingRequestArray[$swingtype] = array($validRange[0], $validRange[1]);
+            foreach (array_keys($swingRequestArrayArray[$playerIdx]) as $swingtype) {
+                $swingRequestArray[$swingtype] = BMDieSwing::swing_range($swingtype);
             }
         }
 
@@ -3994,6 +4155,33 @@ class BMGame {
         }
 
         return $swingRequestArrayArray;
+    }
+
+    /**
+     * Get the array of turbo sizes possible for all active dice for a particular player
+     *
+     * @param int $playerIdx
+     * @return array
+     */
+    protected function get_turboSizeArray($playerIdx) {
+        $turboSizeArray = array();
+
+        if (BMGameState::START_TURN == $this->gameState) {
+            foreach ($this->playerArray[$playerIdx]->activeDieArray as $dieIdx => $die) {
+                if ($die->has_skill('Turbo')) {
+                    if ($die instanceof BMDieOption) {
+                        $turboSizeArray[$dieIdx] = $die->optionValueArray;
+                    } elseif (isset($die->swingType)) {
+                        $swingRange = BMDieSwing::swing_range($die->swingType);
+                        $turboSizeArray[$dieIdx] = range($swingRange[0], $swingRange[1]);
+                    } else {
+                        $turboSizeArray[$dieIdx] = array($die->max);
+                    }
+                }
+            }
+        }
+
+        return $turboSizeArray;
     }
 
     /**
--- a/src/engine/BMGameAction.php
+++ b/src/engine/BMGameAction.php
@@ -204,6 +204,15 @@ class BMGameAction {
         // in some way we want to report prior to being captured
         if ($attackType == 'Trip') {
             $defenderRerollsEarly = TRUE;
+
+            // deliberately hide the value of a turbo die performing a trip attack
+            if (FALSE !== strpos(
+                $preAttackDice['attacker'][0]['recipe'],
+                BMSkill::abbreviate_skill_name('Turbo')
+            )) {
+                $preAttackDice['attacker'][0]['recipeStatus'] =
+                    $preAttackDice['attacker'][0]['recipe'];
+            }
         } else {
             $defenderRerollsEarly = FALSE;
         }
@@ -242,7 +251,7 @@ class BMGameAction {
             $message .= $this->trip_message($preAttackDice, $postAttackDice, $messageDefender);
         } else {
             $messageAttacker = $this->messageAttacker($preAttackDice, $postAttackDice);
-            $message .= $messageDefender . '; ' . $messageAttacker;
+            $message .= implode('; ', array_filter(array($messageDefender, $messageAttacker)));
         }
 
         return $message;
@@ -400,7 +409,7 @@ class BMGameAction {
      * @return string
      */
     protected function trip_message($preAttackDice, $postAttackDice, $messageDefender) {
-        $message = '';
+        $messageArray = array();
 
         // this only triggers for trip attacks, so there can only be one attacker involved
         $midAttackDice = $preAttackDice;
@@ -415,8 +424,15 @@ class BMGameAction {
                 $postAttackDice['attacker'][0]['recipeAfterTripAttack'];
         }
 
-        $message .= $this->messageAttacker($preAttackDice, $midAttackDice);
-        $message .= '; ' . $messageDefender;
+        if (array_key_exists('forceHideDieReroll', $postAttackDice['attacker'][0])) {
+            $midAttackDice['attacker'][0]['forceHideDieReroll'] =
+                $postAttackDice['attacker'][0]['forceHideDieReroll'];
+            $preAttackDice['attacker'][0]['recipeStatus'] =
+                $preAttackDice['attacker'][0]['recipe'];
+        }
+
+        $messageArray[] = $this->messageAttacker($preAttackDice, $midAttackDice, 'Trip');
+        $messageArray[] = $messageDefender;
 
         $splittingAfterTrip = (count($midAttackDice['attacker']) !=
                                count($postAttackDice['attacker']));
@@ -425,8 +441,10 @@ class BMGameAction {
 
         // deal with splitting after trip
         if ($splittingAfterTrip || $morphingAfterTrip) {
-            $message .= '; ' . $this->messageAttacker($midAttackDice, $postAttackDice);
+            $messageArray[] = $this->messageAttacker($midAttackDice, $postAttackDice);
         }
+
+        $message = implode('; ', array_filter($messageArray));
 
         return $message;
     }
@@ -436,9 +454,10 @@ class BMGameAction {
      *
      * @param array $preAttackDice
      * @param array $postAttackDice
+     * @param string $attackType
      * @return string
      */
-    protected function messageAttacker($preAttackDice, $postAttackDice) {
+    protected function messageAttacker($preAttackDice, $postAttackDice, $attackType = '') {
         $messageAttackerArray = array();
 
         // deal with the possibility of an attacker that splits
@@ -453,7 +472,6 @@ class BMGameAction {
         // Report what happened to each attacking die
         foreach ($preAttackDice['attacker'] as $idx => $attackerInfo) {
             $initialAttackerInfo = $attackerInfo;
-
             $postInfo = $postAttackDice['attacker'][$idx];
             $postEventsAttacker = array();
 
@@ -483,10 +501,18 @@ class BMGameAction {
                 $postEventsAttacker,
                 $this->message_recipe_change($attackerInfo, $postInfo)
             );
-            $this->message_append(
-                $postEventsAttacker,
-                $this->message_value_change($attackerInfo, $postInfo)
-            );
+            if (array_key_exists('forceHideDieReroll', $postInfo) &&
+                !($attackType == 'Trip')) {
+                $this->message_append(
+                    $postEventsAttacker,
+                    'rerolled from ' . $attackerInfo['value'] . ', changed size due to turbo'
+                );
+            } else {
+                $this->message_append(
+                    $postEventsAttacker,
+                    $this->message_value_change($attackerInfo, $postInfo)
+                );
+            }
             $this->message_append(
                 $postEventsAttacker,
                 $this->message_out_of_play($attackerInfo, $postInfo)
@@ -517,13 +543,15 @@ class BMGameAction {
      * @return int
      */
     protected static function max_from_recipe($recipe) {
+        $matches = array();
+
         if (preg_match('/\(.+=.+\)/', $recipe)) {
             // if there is an equals present, it must be a swing or option die, so play safe and
             // only look for numbers directly after an equals sign
             preg_match_all('/=(\d+)/', $recipe, $matches);
         } elseif (preg_match('/\(.+\/.+\)/', $recipe)) {
             // if there is an unspecified option die (which shouldn't occur), then return no value
-            $matches = array();
+
         } else {
             preg_match_all('/(\d+)/', $recipe, $matches);
         }
@@ -632,7 +660,7 @@ class BMGameAction {
      * @param array $postInfo
      * @return string
      */
-    protected function message_size_change($preInfo, $postInfo) {
+    protected function message_size_change(array $preInfo, array $postInfo) {
         $message = '';
 
         if ($preInfo['max'] != $postInfo['max']) {
@@ -655,7 +683,7 @@ class BMGameAction {
      * @param bool $showFrom
      * @return string
      */
-    protected function message_recipe_change($preInfo, $postInfo, $showFrom = TRUE) {
+    protected function message_recipe_change(array $preInfo, array $postInfo, $showFrom = TRUE) {
         $message = '';
 
         if ($preInfo['recipe'] != $postInfo['recipe']) {
@@ -678,8 +706,10 @@ class BMGameAction {
      * @param array $postInfo
      * @return string
      */
-    protected function message_value_change($preInfo, $postInfo) {
-        if ($postInfo['doesReroll']) {
+    protected function message_value_change(array $preInfo, array $postInfo) {
+        if (array_key_exists('forceHideDieReroll', $postInfo)) {
+            $message = 'rolled ' . $postInfo['value'];
+        } elseif ($postInfo['doesReroll']) {
             $message = 'rerolled ' . $preInfo['value'] . ' => ' . $postInfo['value'];
         } else {
             $message = 'does not reroll';
@@ -700,6 +730,46 @@ class BMGameAction {
         } else {
             $message = 'was not captured';
         }
+
+        return $message;
+    }
+
+    /**
+     * Describes the change of die due to turbo
+     *
+     * @return string
+     */
+    protected function friendly_message_set_turbo() {
+        $messageArray = array();
+
+        $preTurboDice = $this->params['preTurboDice'];
+        $postTurboDice = $this->params['postTurboDice'];
+        $attackType = '';
+        if (array_key_exists('attackType', $this->params)) {
+            $attackType = $this->params['attackType'];
+        }
+
+        if (count($preTurboDice) != count($postTurboDice)) {
+            throw new LogicException('The number of dice does not change due to Turbo');
+        }
+
+        foreach ($preTurboDice as $dieIdx => $preTurboDie) {
+            $postTurboDie = $postTurboDice[$dieIdx];
+
+            $messageSubArray = array();
+            $messageSubArray[] = 'Turbo die ' . $preTurboDie['recipe'] . ' ' .
+                                 $this->message_size_change($preTurboDie, $postTurboDie);
+            $messageSubArray[] = $this->message_recipe_change($preTurboDie, $postTurboDie);
+            if ($attackType != 'Trip') {
+                // intentionally hide the die value change to make it look like it happens
+                // in the trip attack
+                $messageSubArray[] = $this->message_value_change($preTurboDie, $postTurboDie);
+            }
+
+            $messageArray[] = implode(', ', array_filter($messageSubArray));
+        }
+
+        $message = implode('; ', $messageArray);
 
         return $message;
     }
--- a/deploy/database/schema.game.sql
+++ b/deploy/database/schema.game.sql
@@ -189,6 +189,13 @@ CREATE TABLE die_status (
     name               VARCHAR(20) NOT NULL
 );
 
+CREATE TABLE game_turbo_cache (
+    game_id      MEDIUMINT UNSIGNED NOT NULL,
+    die_idx      TINYINT UNSIGNED NOT NULL,
+    turbo_size   TINYINT UNSIGNED NOT NULL,
+    INDEX (game_id, die_idx)
+);
+
 CREATE TABLE open_game_possible_buttons (
     game_id            MEDIUMINT UNSIGNED NOT NULL,
     button_id          SMALLINT UNSIGNED NOT NULL
--- a/deploy/database/updates/00091_turbo_01.sql
+++ b/deploy/database/updates/00091_turbo_01.sql
@@ -1,0 +1,6 @@
+CREATE TABLE game_turbo_cache (
+    game_id      MEDIUMINT UNSIGNED NOT NULL,
+    die_idx      TINYINT UNSIGNED NOT NULL,
+    turbo_size   TINYINT UNSIGNED NOT NULL,
+    INDEX (game_id, die_idx)
+);
--- a/src/api/ApiResponder.php
+++ b/src/api/ApiResponder.php
@@ -655,17 +655,15 @@ class ApiResponder {
         if (!(array_key_exists('chat', $args))) {
             $args['chat'] = '';
         }
-        $retval = $interface->game()->submit_turn(
-            $_SESSION['user_id'],
-            $args['game'],
-            $args['roundNumber'],
-            $args['timestamp'],
-            $args['dieSelectStatus'],
-            $args['attackType'],
-            (int)$args['attackerIdx'],
-            (int)$args['defenderIdx'],
-            $args['chat']
-        );
+        if (!(array_key_exists('turboVals', $args))) {
+            $args['turboVals'] = array();
+        }
+
+        $args['playerId'] = $_SESSION['user_id'];
+        $args['attackerIdx'] = (int)$args['attackerIdx'];
+        $args['defenderIdx'] = (int)$args['defenderIdx'];
+
+        $retval = $interface->game()->submit_turn($args);
 
         if (isset($retval)) {
             $interface->player()->update_last_action_time($_SESSION['user_id'], $args['game']);
--- a/src/api/ApiSpec.php
+++ b/src/api/ApiSpec.php
@@ -617,6 +617,12 @@ class ApiSpec {
                 'defenderIdx' => 'number',
             ),
             'permitted' => array(
+                'turboVals' => array(
+                    'arg_type' => 'array',
+                    'has_keys' => TRUE,
+                    'key_type' => 'number',
+                    'elem_type' => 'number',
+                ),
                 'chat' => array(
                     'arg_type' => 'string',
                     'maxlength' => self::GAME_CHAT_MAX_LENGTH,
--- a/src/engine/BMAttack.php
+++ b/src/engine/BMAttack.php
@@ -289,7 +289,13 @@ abstract class BMAttack {
      */
     public function commit_attack(&$game, array &$attackers, array &$defenders) {
         // Paranoia
-        if (!$this->validate_attack($game, $attackers, $defenders, $game->firingAmount)) {
+        if (!$this->validate_attack(
+            $game,
+            $attackers,
+            $defenders,
+            array('helpValue' => $game->firingAmount)
+        )
+        ) {
             return FALSE;
         }
 
@@ -414,7 +420,7 @@ abstract class BMAttack {
      * @param array $defenders
      * @return bool
      */
-    protected function search_onevone($game, $attackers, $defenders) {
+    protected function search_onevone($game, $attackers, $defenders, array $args = array()) {
         // Sanity check
 
         if (count($attackers) < 1 || count($defenders) < 1) {
@@ -427,7 +433,7 @@ abstract class BMAttack {
 
         foreach ($aIt as $att) {
             foreach ($dIt as $def) {
-                if ($this->validate_attack($game, $att, $def)) {
+                if ($this->validate_attack($game, $att, $def, $args)) {
                     return TRUE;
                 }
             }
--- a/src/engine/BMAttackDefault.php
+++ b/src/engine/BMAttackDefault.php
@@ -51,7 +51,7 @@ class BMAttackDefault extends BMAttack {
      * @param array $defenders
      * @return bool
      */
-    public function validate_attack($game, array $attackers, array $defenders) {
+    public function validate_attack($game, array $attackers, array $defenders, array $args = array()) {
         $this->validationMessage = '';
         $this->resolvedType = '';
 
@@ -65,7 +65,7 @@ class BMAttackDefault extends BMAttack {
                     $attack->add_die($die);
                 }
             }
-            if ($attack->validate_attack($game, $attackers, $defenders)) {
+            if ($attack->validate_attack($game, $attackers, $defenders, $args)) {
                 $validAttackTypeArray[] = $attackType;
             }
         }
--- a/src/engine/BMAttackPower.php
+++ b/src/engine/BMAttackPower.php
@@ -39,10 +39,10 @@ class BMAttackPower extends BMAttack {
      * @param BMGame $game
      * @param array $attackers
      * @param array $defenders
-     * @param int|NULL $helpValue
+     * @param array $args
      * @return bool
      */
-    public function validate_attack($game, array $attackers, array $defenders, $helpValue = NULL) {
+    public function validate_attack($game, array $attackers, array $defenders, array $args = array()) {
         $this->validationMessage = '';
 
         if (1 != count($attackers)) {
@@ -63,6 +63,12 @@ class BMAttackPower extends BMAttack {
         if (!$this->are_skills_compatible($attackers, $defenders)) {
             // validation message set within $this->are_skills_compatible()
             return FALSE;
+        }
+
+        $helpValue = NULL;
+
+        if (array_key_exists('helpValue', $args)) {
+            $helpValue = $args['helpValue'];
         }
 
         if (is_null($helpValue)) {
--- a/src/engine/BMAttackSkill.php
+++ b/src/engine/BMAttackSkill.php
@@ -165,10 +165,10 @@ class BMAttackSkill extends BMAttack {
      * @param BMGame $game
      * @param array $attackers
      * @param array $defenders
-     * @param double $helpValue
+     * @param array $args
      * @return bool
      */
-    public function validate_attack($game, array $attackers, array $defenders, $helpValue = NULL) {
+    public function validate_attack($game, array $attackers, array $defenders, array $args = array()) {
         $this->validationMessage = '';
 
         if (count($attackers) < 1) {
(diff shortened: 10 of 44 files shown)


cgolubi1:
This seems to pass the dev site acceptance criteria.

blackshadowshade:
I've opened issue #2156 to allow discussion of issues relating to the interaction between turbo and other skills. That should leave this pull request less cluttered and more focussed on other issues with turbo, e.g., look and feel issues.

irilyth:
I think it'd be better if the selection interface was the same as the selection interface for swing dice in general, i.e. a text box rather than a menu. (Or, the other swing die selection screens could use menus too, but I think a text box is more fun.) I gather that this has some other implications, but that's my two cents about look and feel.

irilyth:
I'm not sure how I feel about the Turbo selection element popping in and out as you select dice. I probably suggested this at one point, but seeing it in action, I don't like that other screen elements jump around as a result.

If you don't have any Turbo Swing dice, I think it shouldn't be present. I wonder if it'd look better if it was always there if you have Turbo Swing dice, and switched from grayed-out to live when you clicked on the relevant die?

blackshadowshade:
One obvious thing: the turbo selector or text box should be prefilled with the old die size.

irilyth:
I'm not sure how I feel about the Turbo selection element popping in and out as you select dice. I probably suggested this at one point, but seeing it in action, I don't like that other screen elements jump around as a result.

If you don't have any Turbo Swing dice, I think it shouldn't be present. I wonder if it'd look better if it was always there if you have Turbo Swing dice, and switched from grayed-out to live when you clicked on the relevant die?

AdmiralJota:
I agree that the Turbo selector should always be there if you still have Turbo dice left. I like the idea of it being a text box like the regular swing selector (and being pre-filled with the old size). I also think that greying out any Turbo selectors for Turbo dice that aren't selected makes sense.

There was also some discussion on showing available ranges, so here are some possible mockups for that:

turbo
turbo2
turbo3

irilyth:
I like the third one: Range to the right, no colons.

(I also think the main swing die selection page would look better this way, if we think harmony is important.)

blackshadowshade:
So, in the original pull request, I raised the issue that I think the swing range information needs to be present whenever there are turbo dice present, so that you know that your opponent's (X,X)! could potentially be as large as (20,20) or as small as (4,4). The current implementation doesn't supply this information.

irilyth:
> I think the swing range information needs to be present whenever there are turbo dice present

Do you think this is also true when there are Mood Swing dice present? Your opponent can't control what size the dice will be, but it still seems like it'd be important to be aware of the potential, even if it's uncontrolled.

A counter-argument: It is important, but new players won't be using Turbo Swing (or Mood Swing) dice while they're still learning the game anyway, and by the time you get to the point that you even understand how to use Turbo Swing dice effectively, you'll probably have the ranges memorized.

Another suggestion: Have swing die ranges on screen whenever there are Swing dice present; put them down below the "Button specials" and "Die skills" help lines. If you don't remember what a die or button does, that's where you'd already be going to look.

blackshadowshade:
Personally, I couldn't tell you what range a W swing die has, because W swing tends to be rather uncommon.

I'm happy to have the swing ranges visible on the screen, but we'd need to discuss exactly where this would be, because we weren't able to get consensus the last time we discussed this issue. Anyway, it's not really important for the current pull request, I was just noting it.

AdmiralJota:
As for keeping the swing ranges visible, my preferred solution would be to include swing dice in the list of skills at the bottom, so that "Die skills: Doppelganger (D) Turbo (!)"  becomes "Die skills: Doppelganger (D) Turbo (!) X Swing (X)", with the range mentioned in the skill description that appears when you hover over it.

AdmiralJota:
> A counter-argument: It is important, but new players won't be using Turbo Swing (or Mood Swing) dice while they're still learning the game anyway, and by the time you get to the point that you even understand how to use Turbo Swing dice effectively, you'll probably have the ranges memorized.

Hmm. How much longer do I need to play for before I reach that point?

blackshadowshade:
Ah, hover. We'll leave *that* discussion for #2072.  XD

blackshadowshade:
Okay, we have a problem in Game 21. The database says:

 select * from die where game_id = 21;

+------+----------+-------------------+---------+-----------+----------+----------+-------+------------+-----------------+
| id   | owner_id | original_owner_id | game_id | status_id | recipe   | position | value | actual_max | flags           |
+------+----------+-------------------+---------+-----------+----------+----------+-------+------------+-----------------+
| 2298 |        3 |                 3 |      21 |         1 | (10)     |        0 |     7 |       NULL | NULL            |
| 2299 |        3 |                 3 |      21 |         1 | (10)     |        1 |     9 |       NULL | NULL            |
| 2300 |        3 |                 3 |      21 |         1 | (12/20)  |        2 |     9 |         12 | NULL            |
| 2301 |        2 |                 2 |      21 |         1 | (4)      |        0 |     1 |       NULL | NULL            |
| 2302 |        2 |                 2 |      21 |         1 | (X)!     |        1 |     1 |          9 | NULL            |
| 2303 |        2 |                 2 |      21 |         1 | (X)!     |        2 |    15 |         10 | NULL            |
| 2304 |        3 |                 2 |      21 |         2 | (8)      |        0 |     5 |       NULL | NULL            |
| 2305 |        2 |                 3 |      21 |         2 | (20/30)! |        0 |    12 |         20 | NULL            |
| 2306 |        2 |                 3 |      21 |         2 | (10)     |        1 |     2 |       NULL | WasJustCaptured |
+------+----------+-------------------+---------+-----------+----------+----------+-------+------------+-----------------+

Somehow, the 10-sider has ended up with a value of 15.

blackshadowshade:
And the relevant bit from the game action log:

select * from game_action_log where game_id = 21;
+-----+---------+---------------------+------------+----------------------+---------------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| id  | game_id | action_time         | game_state | action_type          | acting_player | message                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                          |
+-----+---------+---------------------+------------+----------------------+---------------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| 301 |      21 | 2017-01-17 03:37:17 |         24 | choose_die_values    |             2 | NULL                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             |
| 303 |      21 | 2017-01-17 03:40:34 |         24 | choose_die_values    |             3 | NULL                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             |
| 304 |      21 | 2017-01-17 03:40:34 |         26 | determine_initiative |             0 | {"roundNumber":1,"playerData":{"3":{"initiativeDice":[{"recipe":"(10)","min":1,"max":10,"value":7,"doesReroll":true,"captured":false,"outOfPlay":false,"recipeStatus":"(10):7","included":true},{"recipe":"(10)","min":1,"max":10,"value":6,"doesReroll":true,"captured":false,"outOfPlay":false,"recipeStatus":"(10):6","included":true},{"recipe":"(10)","min":1,"max":10,"value":2,"doesReroll":true,"captured":false,"outOfPlay":false,"recipeStatus":"(10):2","included":true},{"recipe":"(12\/20=12)","min":1,"max":"12","value":9,"doesReroll":true,"captured":false,"outOfPlay":false,"recipeStatus":"(12\/20=12):9","included":true},{"recipe":"(20\/30=20)!","min":1,"max":"20","value":12,"doesReroll":true,"captured":false,"outOfPlay":false,"recipeStatus":"(20\/30=20)!:12","included":true}],"slowButton":false},"2":{"initiativeDice":[{"recipe":"(4)","min":1,"max":4,"value":1,"doesReroll":true,"captured":false,"outOfPlay":false,"recipeStatus":"(4):1","included":true},{"recipe":"(8)","min":1,"max":8,"value":6,"doesReroll":true,"captured":false,"outOfPlay":false,"recipeStatus":"(8):6","included":true},{"recipe":"(X=9)!","min":1,"max":9,"value":1,"doesReroll":true,"captured":false,"outOfPlay":false,"recipeStatus":"(X=9)!:1","included":true},{"recipe":"(X=9)!","min":1,"max":9,"value":6,"doesReroll":true,"captured":false,"outOfPlay":false,"recipeStatus":"(X=9)!:6","included":true}],"slowButton":false}},"initiativeWinnerId":2} |
| 308 |      21 | 2017-01-17 04:07:40 |         44 | attack               |             2 | {"attackType":"Skill","preAttackDice":{"attacker":[{"recipe":"(8)","min":1,"max":8,"value":6,"doesReroll":true,"captured":false,"outOfPlay":false,"recipeStatus":"(8):6"},{"recipe":"(X=9)!","min":1,"max":9,"value":6,"doesReroll":true,"captured":false,"outOfPlay":false,"recipeStatus":"(X=9)!:6"}],"defender":[{"recipe":"(20\/30=20)!","min":1,"max":20,"value":12,"doesReroll":true,"captured":false,"outOfPlay":false,"recipeStatus":"(20\/30=20)!:12"}]},"postAttackDice":{"attacker":[{"recipe":"(8)","min":1,"max":8,"value":5,"doesReroll":true,"captured":false,"outOfPlay":false,"recipeStatus":"(8):5"},{"recipe":"(X=9)!","min":1,"max":9,"value":2,"doesReroll":true,"captured":false,"outOfPlay":false,"recipeStatus":"(X=9)!:2"}],"defender":[{"recipe":"(20\/30=20)!","min":1,"max":20,"value":12,"doesReroll":true,"captured":true,"outOfPlay":false,"recipeStatus":"(20\/30=20)!:12"}]},"fireCache":null}                                                                                                                                  |
| 321 |      21 | 2017-01-17 04:10:16 |         44 | attack               |             3 | {"attackType":"Power","preAttackDice":{"attacker":[{"recipe":"(10)","min":1,"max":10,"value":6,"doesReroll":true,"captured":false,"outOfPlay":false,"recipeStatus":"(10):6"}],"defender":[{"recipe":"(8)","min":1,"max":8,"value":5,"doesReroll":true,"captured":false,"outOfPlay":false,"recipeStatus":"(8):5"}]},"postAttackDice":{"attacker":[{"recipe":"(10)","min":1,"max":10,"value":9,"doesReroll":true,"captured":false,"outOfPlay":false,"recipeStatus":"(10):9"}],"defender":[{"recipe":"(8)","min":1,"max":8,"value":5,"doesReroll":true,"captured":true,"outOfPlay":false,"recipeStatus":"(8):5"}]},"fireCache":null}                                                                                                                                                                                                                                                                                                                                                                                                                                |
| 325 |      21 | 2017-01-17 04:33:00 |         44 | attack               |             2 | {"attackType":"Power","preAttackDice":{"attacker":[{"recipe":"(X=17)!","min":1,"max":17,"value":2,"doesReroll":true,"captured":false,"outOfPlay":false,"recipeStatus":"(X=17)!:2"}],"defender":[{"recipe":"(10)","min":1,"max":10,"value":2,"doesReroll":true,"captured":false,"outOfPlay":false,"recipeStatus":"(10):2"}]},"postAttackDice":{"attacker":[{"recipe":"(X=17)!","min":1,"max":17,"value":15,"doesReroll":true,"captured":false,"outOfPlay":false,"recipeStatus":"(X=17)!:15"}],"defender":[{"recipe":"(10)","min":1,"max":10,"value":2,"doesReroll":true,"captured":true,"outOfPlay":false,"recipeStatus":"(10):2"}]},"fireCache":null}                                                                                                                                                                                                                                                                                                                                                                                                            |
+-----+---------+---------------------+------------+----------------------+---------------+-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+

blackshadowshade:
So, deciphering this:

Player 1 starts the round with
(10):7, (10):6, (10):2, (12/20=12):9, (20/30=20)!:12

Player 2 starts with
(4):1, (8):6, (X=9)!:1, (X=9)!:6

and Player 2 starts the round.

Player 2 performs a skill attack using the (8):6 and the (X=9)!:6 to capture the (20/30=20)!:12, and the reroll gives the new values (8):5 and (X=9)!:2.

*** somewhere here, the turbo die changes from 9 sides to 17 sides ***

Player 1 performs a power attack using the (10):6 to capture the (8):5, and the reroll gives (10):9.

Player 2 performs a power attack using the (X=17)!:2 to capture the (10):2, and the reroll gives (X=17)!:15.

*** somewhere here, the turbo die changes from 17 sides to 10 sides ***

blackshadowshade:
It looks like the reroll is occurring at the wrong time, *before* the turbo size change occurs. I should be able to track that down pretty easily.

irilyth:
A look-and-feel thing:

> One obvious thing: the turbo selector or text box should be prefilled with the old die size.

Agreed, except: If you change the selection or text, and then click on the die, and then click it again, it should show up with the value you changed it to. (Right now, it resets back to the blank at the top of the list.)

blackshadowshade:
Okay, so I'm taking a look at the suggestions here and as far as I can see, there are a few:
- use a type-in field instead of a dropdown
- make turbo selectors always visible, just greyed when the turbo die is not selected
- prefill selector with the current size
- keep selected size in the turbo selector even when deselecting the turbo die

Did I miss any here? I'll be giving my thoughts on these in a follow-up.

blackshadowshade:
1. Using a type-in field instead of a dropdown

I don't like this suggestion because it means that we have a different way of selecting turbo sizes for turbo swing and turbo option dice. I can see the argument for having selection consistent with the swing setting page, although I think this would be better resolved by having the swing setting page using a dropdown selector.

2. Making the turbo selectors always visible, but greyed when the turbo die is not selected

I'm actually surprised that people are annoyed by the sudden appearance and disappearance of the turbo selector. I thought that people would be more annoyed by the additional vertical real estate that is being taken up when the turbo selector is not being used. However, I personally don't have any strong opinion on this either way.

3. Prefilling the selector with the current size

Sure, this makes a lot of sense.

4. Keeping the selected size in the turbo selector even when deselecting the turbo die

I *think* that this makes sense in the case that the turbo selector is always visible (but possibly grey), but I'm not certain that it makes sense when the turbo selector appears and disappears. I'm trying to work out why you would select a die, choose a turbo value, then deselect the die and then select it again. Would this be because you've accidentally clicked on the die again? In this case, then yes, it makes sense to keep the value you've selected. However, is it because you've changed your mind about the attack twice? Then, I'm not so certain that it makes sense.

blackshadowshade:
@cgolubi1, I'd like to suggest that that the testing site for Turbo be updated with the current version of the pull request. I've not addressed any of the look-and-feel issues yet, but the backend play issues should be fixed, and I'd like the testers to find more underlying bugs while we debate the look-and-feel issues.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1261/ (if it passes)

irilyth:
> 1. Using a type-in field instead of a dropdown
>
> I don't like this suggestion because it means that we have a different way of selecting turbo sizes for turbo swing and turbo option dice. I can see the argument for having selection consistent with the swing setting page, although I think this would be better resolved by having the swing setting page using a dropdown selector.

I had said earlier "Or, the other swing die selection screens could use menus too, but I think a text box is more fun.)" and I think we talked more about this on IRC, but so it's on this ticket: I don't like a menu for swing die size selection because they have a wide range of values (most commonly 17, but as high as 27), and no information is conveyed by having them enumerated -- everyone knows what "4 - 20" means. (Compare to the menu to select a button on the game creation page: It's hugely long, but conveys vital information, such as what buttons match the criteria you've selected, and how their names are spelled. Imagine if that was a text box, and you had to type in the button's name, and spell it right, with case and punctuation and whitespace and everything correct -- that'd be much worse.)

I don't feel strongly about option dice; if we want consistency, I think it'd be fine to make the option selector a text box too, but it's not as problematic to make it a menu. Having them be different also makes it easy to visually tell which of your dice are Option and which are Swing, in case you see 6/12 and think 6-12.

I don't think consistency between Swing and Option is important at all, though. I think consistency between start-of-game selection and mid-game Turbo selection is more important.

> 2. Making the turbo selectors always visible, but greyed when the turbo die is not selected
>
> I'm actually surprised that people are annoyed by the sudden appearance and disappearance of the turbo selector. I thought that people would be more annoyed by the additional vertical real estate that is being taken up when the turbo selector is not being used. However, I personally don't have any strong opinion on this either way.

For me, the difference has to do with whether the vertical space is being used for anything. Have a blank bar of vertical space where a Turbo selector would be, in a game with no Turbo dice, would use vertical space for no reason at all. Here, it's doing two things: (1) Another visual indicator that you have Turbo dice; (2) Stopping the other elements on the screen from jumping around when you click on Turbo dice. The latter is more what I don't like about having it pop in and out -- if it appeared and disappeared, but didn't make the rest of the screen jump around, that'd be ok. (If we like "completely absent" better than "gray" for some reason.)

Another way to do that would be to have it off to the side, rather than vertical; I don't think that's what we want here at all, but that's a way to make appearing and disappearing things not make other elements jump around, have them appear and disappear in places where nothing is below or to the right of them.

> 4. Keeping the selected size in the turbo selector even when deselecting the turbo die
>
> I'm trying to work out why you would select a die, choose a turbo value, then deselect the die and then select it again. Would this be because you've accidentally clicked on the die again?

Sure; or because you changed your mind, and then changed it back again. I often find myself going "ok, skill attack, click click, oh wait, I should actually use this other die instead, unclick click, oh shoot no that's not what I want, unclick click". If I set a Turbo size after the first click, I probably want to set it again after the third click.

Conversely, are there any situations where a player *wouldn't* want this behavior? Where you'd unselect and then re-select a Turbo die and be annoyed that the selector hadn't reset?

blackshadowshade:
So, just addressing number 1, I see drop downs with large number of numeric values on a regular basis. Take, for example, birth date, which has often drop-downs for day, month, and year. Or a time selector, with drop-downs for the hour and minute (perhaps in 5 minute steps), and perhaps an AM/PM selector.

I feel that these have typically been used to provide implicit validation of input values. However, there is also a view that these may be less disruptive to use on a mobile device, where you have the choice between (i) having the menu enlarge, and selecting a value, possibly needing to swipe up/down to find the value, and (ii) tapping in the text entry field, having a keyboard/keypad appear, typing in the value, and then pressing enter or tapping out of the keyboard view.

Anyway, I'm open to both ways, I'd just like to have this issue discussed before I go about locking in one way or the other.

AdmiralJota:
For me, the text box for swing selection just feels more natural. I'm not 100% sure why, but I think a part of it might be that (for me, anyway), it's a calculated number. I'm doing math and coming up with a numeric value, then entering that value. I'm not just selecting an enumerated item like I would be with a birth date.

The same isn't entirely true for option dice. With a swing, I might say "set it to 4 + (the result of all this math, capped at 20)". But with an option die, I usually just have to pick the closest value and make do.

> Another way to do that would be to have it off to the side, rather than vertical; I don't think that's what we want here at all, but that's a way to make appearing and disappearing things not make other elements jump around, have them appear and disappear in places where nothing is below or to the right of them.

I think that would actually be worse: you would still have things jumping around. They'd just be jumping _left and right_ rather than up and down. And it would be the whole battle mat jumping back and forth as it tried to re-center itself.

(But personally, I'm not bothered by having the vertical space taken up by the turbo selector. If things are likely to get pushed down the screen _anyway_ when you make your attack, then we might as well just reserve the space from the start.)

I could go either way on keeping the size in the turbo selector when you deselect and reselect it. In fact, I could see a pretty reasonable case for having the selector be _empty_ whenever the turbo dice are unselected (i.e. whenever it's greyed out), and then auto-populating it with a value when you click on your die. 

On the first click, that would be the current value, of course. But if you deselect and reselect it, I think it could be reasonable to repopulate it with either the original value or the last value you typed there. And if it's blank in the meantime, I don't think either behavior would be too terribly confusing.

cgolubi1:
Yep, i'll try to get on top of that this weekend.

AdmiralJota:
> @cgolubi1, I'd like to suggest that that the testing site for Turbo be updated with the current version of the pull request

Would it be possible to fix the bugged games (like #28) at the same time?

blackshadowshade:
Unfortunately, we're either going to have to fix those manually or get a fresh copy of the database.

I think that there are not too many bugged games, so I'm prepared to fix these myself if I get ssh access.

cgolubi1:
Okay, because of a miscommunication about rebasing, the easiest thing to do is to stand up a new site with the new code, so i did, at http://54.160.177.210/.  I haven't taken the older site down yet.  This has the new code, and a new copy of the staging database (so if you have an account on staging, you can use it).

AdmiralJota:
Just created a bunch of open games there. Let the testing commence.

irilyth:
Another UI thing related to #943: When you're setting swing sizes at the start of the round, it'd be nice if there were some indication that you're setting the size for a Turbo die. In a game with Coyote, who has a bunch of Swing and Option dice, I had to double-check a couple of times to make sure I knew which dice were and weren't Turbo.

Fixing #943 would solve this. There might be other ways to solve it, but I think a general fix for #943 would be better. :^)

AdmiralJota:
When you attack with a Turbo Warrior die, I think that the Turbo logging is somehow preventing it from logging the loss of the Warrior skill.

E.g., from http://buttonweavers.com/ui/game.html?game=18110 in production:

> irilyth performed Skill attack using [(1):1,\`(4):4] against [(6):5]; Defender (6) was captured; Attacker (1) rerolled 1 => 1; Attacker \`(4) recipe changed from \`(4) to (4), rerolled 4 => 3

In that last sentence, it mentions the recipe changing to remove the Warrior skill. But in http://54.160.177.210/ui/game.html?game=25 on the test site we have:

> Jota performed Skill attack using [fn(X=4):3,fn(X=4):3,\`d(Y=2)!:2] against [o^(11):8]; Defender o^(11) recipe changed to o^n(11), was captured; Attacker fn(X=4) rerolled 3 => 3; Attacker fn(X=4) rerolled 3 => 2

Immediately followed by:

>Turbo die d(Y=2)! changed size from 2 to 20 sides, rerolled 2 => 7

No mention of the \`d(Y=2) changing recipe to lose Warrior. 

From the subsequent turn in that same game, we do see:

> Jota performed Skill attack using [fn(X=4):3,fn(X=4):2,\`H(Y=2,Y=2):4,d(Y=20)!:7] against [oH(16):16]; Defender oH(16) recipe changed to oHn(16), was captured; Attacker fn(X=4) rerolled 3 => 1; Attacker fn(X=4) rerolled 2 => 4; Attacker \`H(Y=2,Y=2) changed size from 4 to 8 sides, recipe changed from \`H(Y=2,Y=2) to H(Y=4,Y=4), rerolled 4 => 8

In which the last sentence shows that we _do_ keep the notification when the size is changing due to Mighty.

Should the Turbo skill size change perhaps be logged the same way the Might/Weak ones are? And how does it presently work if a Warrior dice is a Mood swing?

AdmiralJota:
It looks like we don't have any Warrior Mood dice yet, so I guess we can't test that. But from http://54.160.177.210/ui/game.html?game=23, I see:

> irilyth performed Power attack using [^(U=30)?:11] against [sf(12):1]; Defender sf(12) was captured; Attacker ^(U=30)? changed size from 30 to 10 sides, recipe changed from ^(U=30)? to ^(U=10)?, rerolled 11 => 8

So it looks like the Mood swing size change is incorporated into the normal log rather than being a separate line like Turbo is. I wonder if they should perhaps work the same way, for consistency?

blackshadowshade:
Jota, the problem here is that most skills, including mood and mighty, trigger during the COMMIT_ATTACK game state. However, turbo triggers either before or after this, depending on the type of attack. In the database, we have a separate set of information for logging purposes per game state.

I'll take a look at why the loss of warrior is hidden when a turbo die is included in the attack. My intuition tells me that the logging *should* still be working here, but obviously, it's not.

AdmiralJota:
I notice that the trip logging looks a bit weird. From http://54.160.177.210/ui/game.html?game=23:

> Turbo die t(T=2,T=2)! changed size from 4 to 24 sides, rerolled 3 => 16

Followed by:

> Jota performed Trip attack using [t(T=12,T=12)!:16] against [^(U=8)?:7]; Defender ^(U=8)? recipe changed to ^(U=30)?, rerolled 7 => 20, was not captured

This suggests that the die was rerolled (and resized) _before_ the attack started. But shouldn't the die be rerolled _during_ the trip attack, not before? So shouldn't Turbo trigger mid-trip?

How do we log it when you Trip with a Morphing die? Does Morphing trigger at a different stage from Turbo? (And if so, is there a reason for that?)

blackshadowshade:
As far as I understand turbo and trip, the following should be true:
- there should only be one reroll
- the size of the turbo trip die changes before the attack occurs

Given the limitations of having the turbo trigger before we enter commit attack, I *could* have the turbo die change size initially with no value, and then have the trip attack reroll it. However, we still have the problem that the die size changes before the trip attack actually occurs, so the log would change to something like:

Turbo die t(T=2,T=2)! changed size from 4 to 24 sides

Jota performed Trip attack using [t(T=12,T=12)!] against [^(U=8)?:7]; Attacker rolled 16, Defender ^(U=8)? recipe changed to ^(U=30)?, rerolled 7 => 20, was not captured

Would that be better?

AdmiralJota:
Yeah, I think that would be better.

Although it seems like an optimal log might be closer to: 

> Jota performed Trip attack using [t(T=12,T=12)!] against [^(U=8)?:7]; Attacker recipe t(T=2,T=2)! changed to t(T=12,T=12)!, rerolled 3 => 16, Defender ^(U=8)? recipe changed to ^(U=30)?, rerolled 7 => 20, was not captured

If the defender's Mood size change can be handled seamlessly inside of the Trip, why can't the attacker's Turbo size change?

blackshadowshade:
I guess, because the logging code currently only reads the logging data for a particular game state. I'll need to talk to Chaos about possibly extending that if it's necessary in this case.

AdmiralJota:
I guess a related question might be: why is Mood able to be triggered during that game state, but Turbo isn't able to be?

blackshadowshade:
From memory, the problem here is that sometimes turbo needs to go before all other skills, and sometimes it needs to go after all other skills. That's why I decided to implement turbo skill changes with an extra game state.

I guess I could put a whole pile of conditional logic in to force it to trigger at the right time, but I think that would be far more difficult to get right and test appropriately.

cgolubi1:
> I guess, because the logging code currently only reads the logging data for a particular game state. I'll need to talk to Chaos about possibly extending that if it's necessary in this case.

The logging code outputs zero or one messages per entry in a game_action_log* table.  And i'm pretty sure we don't want to futz with that.  If you have two a separate game_action_log* entry for the turbo size change, why not just output a separate log message for it?

blackshadowshade:
Currently, I *am* outputting a separate log message for turbo change, because that happens in a different internal game state. Jota is arguing that actually, the turbo change is part of the attack, and so shouldn't be logged separately.

AdmiralJota:
And I think that my question wasn't so much "Why don't we consolidate multiple game_action_log* entries into one log message?" as "Why does Turbo need multiple game_action_log* entries when Mood doesn't?" (which I think may be equivalent to the question of "Why does Turbo trigger during different game states from Mood?").

(Of course, I _don't_ know the internal implementation, so it's entirely possible that I am misunderstanding what my _own_ question really boils down to.)

blackshadowshade:
The extra thing with Turbo (and Fire) is that there is an extra input from the player required to complete the attack, which is not necessary with Mood or Mighty.

The current implementation has BMInterface (see the new line 834) submitting the Turbo values to determine die sizes when the BMGame requires them. The BMGame doesn't actually take the turbo values as an inherent part of the input for the attack, but it pauses to wait for the turbo values to be input if there are turbo dice attacking.

This is similar to the way that we deal with Fire dice, where the BMGame also doesn't take the fired values as an inherent part of the input for the attack. It waits for BMInterface to attempt to change the die values, and then continues with the attack.

As I said, it is probably possible to incorporate the turbo sizes into the inherent data required for the attack, but in my opinion, this would require far more conditional logic to get right, and it's more difficult to debug, because the order of skills applying can be quite convoluted in COMMIT_ATTACK. Separating out the turbo resize (and the fire value change) into a new game state means that it is possible to deal with *only* the turbo resize (or fire value change) when things go wrong.

cgolubi1:
I agree with what i think Shadowshade is saying, namely that the work needed to combine the attack and the turbo resize into a single message without breaking anything, isn't worth it for a first version of Turbo.

Jota: to put this in perspective, the backend of Turbo was initially written on the theory that we'd have attacks containing Turbo dice send the player to a separate screen the way Fire attacks do (and the way Turbo attacks did on the DHS site).  We decided we didn't want to do the UI that way for Turbo, even initially, so Shadowshade has been working on the combined UI, but the backend implementation is still what it is, and that's why the action logs have two entries.

blackshadowshade:
So, I've totally reworked the logging to give the impression that the turbo reroll occurs during the attack. Also, recipe changes for turbo dice are now explicitly logged.

Next step: autofilling the turbo size selectors with the previous values.

AdmiralJota:
Since Turbo lets you re-size a die when you re-roll it, and Chance lets you re-roll a die, then shouldn't Turbo trigger when you use Chance?

irilyth:
> Since Turbo lets you re-size a die when you re-roll it, and Chance lets you re-roll a die, then shouldn't Turbo trigger when you use Chance?

I think the Turbo help text should probably be changed to indicate that you can only re-size it when you use it in an attack; I don't think that, for example, you're supposed to be able to resize Ornery Turbo dice when Ornery triggers, or a Turbo die that's the target of a Trip Attack or Boom Attack.

blackshadowshade:
Are there any chance turbo dice in any of the recipes at the moment?

AdmiralJota:
NoopMan: %t(1,3) D(4) ns(20) z(X) c(Z)!

blackshadowshade:
I'm shifting this discussion to #2156, so we can have a go at it properly there.  :)

blackshadowshade:
So, I've now updated the pull request to improve the logging, reserve space for the turbo selector, and prefill turbo values. This should have addressed pretty much all the blockers.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1269/ (if it passes)

I'm going to leave the whole option/text field question for now, because I'm still not convinced that going with a text field is a better solution.

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1270/ (if it passes)

blackshadowshade:
Huh, found a bug already.

capture

blackshadowshade:
So, I can replicate this issue, and it's something to do with the fact that the reserve turbo die has just been added to the game.

blackshadowshade:
BTW, I'm documenting these bugs so that we have some guidance about which responder tests need to be added.

blackshadowshade:
Okay, the first bug is due to turbo selectors now being autofilled with values. This means that I need to change the logic about when to send turbo values. This will now depend on the visibility of the turbo selector.

blackshadowshade:
So, both bugs should now be addressed.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1272/ (if it passes)

AdmiralJota:
Another example of the first one:

bm1

(Noting it here so I remember to come back to this game to re-test after that fix is deployed...)

blackshadowshade:
Okay, next bug (which I suspected would be there, so I went looking for it):

capture

Here, there is a valid attack if the turbo die is resized to 30, but BMAttackTrip->find_attack() doesn't yet consider the turbo resize.

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1276/ (if it passes)

irilyth:
> So, I've now updated the pull request to improve the logging, reserve space for the turbo selector, and prefill turbo values.

I noticed that the Turbo selector seems to keep the previously selected value now if I select and de-select the die, but not if I make an illegal attack choice (e.g. I add wrong and select too many targets for a Berserk Attack, or the wrong dice of my own for a Skill Attack). It probably shouldn't reset in that situation, especially since the die selections don't reset -- the net effect makes it look like you've still got the selections you had before, but in fact you don't have the Turbo size selection, just the die selections.

blackshadowshade:
Okay, turbo values should now be carried across after a failed attack.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1277/ (if it passes)

cgolubi1:
I think your passing build is actually http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1278/ but it seems to be a passing build.

blackshadowshade:
Okay, here's an interesting one ... http://54.160.177.210/ui/game.html?game=33 is now stuck in a game state that it shouldn't be stuck in.

It's a game between North Carolina [pF(10) o(10) (V)! gt(V) h(V)] and Zomulgustar [t(4) p(5/23)! t(9) t(13) rdD(1) rsz(1) r^(1,1) rBqn(Z)?].

From memory, I was doing some attack that triggered the optional fire support of between 0 and 2 pts, so I reduced the fire die value by 2, then ended up with this current game state.

blackshadowshade:
So, I presume what has happened here is that I've performed a power attack with North Carolina's turbo (V)!, and had the option to use fire overshooting, which I have turned on. The fire page has come up as expected, I've overshot, and then the game has somehow gotten stuck.

I'll try to replicate this at home tonight. If I have any problems, I'll get @cgolubi1 to post the action log entries from that game.

blackshadowshade:
And another oversight: the you-can't-win calculation doesn't take into account the possible change in size of the turbo die yet.

There are two ways to address this: actually add the logic, or say that turbo stops the calculation. I'll have to think about this and work out whether the logic is easy or difficult.

irilyth:
This looks slightly weird:

bm

The big gap between the first two selectors, and the third (fifth?) in particular.

blackshadowshade:
So, it *is* possible to toggle visibility and not reserve space on the turbo subspans, but this leads to a small change in the height of the turbo span (the pink bit) when no turbo die is selected.

cgolubi1:
shadowshade: i can't find your passing jenkins run for the latest commit - link me to it if i'm just missing it, or run it if not.

blackshadowshade:
I was waiting until I sorted out the issue with turbo and fire overshooting, but I'll set it going now.

It'll be http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1279/

blackshadowshade:
Ah, I understand what's going on here. Basically, the game gets to BMGameState::ADJUST_FIRE_DICE and stops there to allow the player to adjust the fire dice. Then, the BMInterfaceGame tries to set the turbo sizes, but this fails because we're not up to BMGameState::COMMIT_ATTACK yet. After this, the selected turbo sizes are not stored anywhere, so the game cannot proceed past BMGameState::CHOOSE_TURBO_SWING due to the missing data.

Hmmm, I'll need to think about this one a bit to sort out how to store the chosen turbo sizes.

blackshadowshade:
Okay, I've resolved the issue with turbo and fire. The solution requires an extra table in the database, so there's a database update file attached: 00091_turbo_01.sql

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1280/ (if it passes)

blackshadowshade:
Okay, I got Jenkins passing. This is ready for deployment, assuming that you're happy with it, @cgolubi1.

cgolubi1:
Hmm, so is this missing a database update file to go with the schema change, or am i misreading?

blackshadowshade:
Indeed, it was because I hadn't committed the new database file to the git repo. That's been added now, and it won't change the Jenkins validity.

cgolubi1:
Okay!  Modulo bugs i haven't found yet in my test rig, i'm now actually running replay tests against this branch.  I'll let those go and look for said test-rig bugs (in addition to regressions in the branch, of course).

I also need to get my test rig into git, which i'll do as a separate set of pulls obviously, starting with the random_ai.py library and working out towards all the helper scripts i use.

cgolubi1:
Any game with CestWhat blows up loading game data after specifying dice --- shadowshade, i created http://54.160.177.210/ui/game.html?game=48 for you to look at, but i'll try to have a replay test in case it's not obvious.  It may be obvious, though.  The exception is:
```
Fatal error: Using $this when not in object context in /var/www/engine/BMAttackTrip.php on line 272
```

blackshadowshade:
The error that is reported is indeed obvious: I'm using $this inside a static function, which is obviously silly.

However, this error has been caused by something else, so I'll have to go digging a little to find out what that is.

blackshadowshade:
Okay, found the error. I was assuming that twin swing dice exposed the properties swingMax and swingMin from the subdice, but they didn't. With the fix that I'm putting in, those properties will now be properly exposed for twin swing dice.

blackshadowshade:
The Jenkins run after this change:

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1282/ (if it passes)

cgolubi1:
Something is weird about the default values provided in the box for CestWhat's `t(T,T)!`.  If i initially set T=4 as in http://54.160.177.210/ui/game.html?game=49 when i click on the turbo die to make an attack, the box shows 8 by default.  (And thus if i don't interact with the box, my setting changes from T=4 to T=8, as happened in game 48.)  If i initially set T=12 as in game 50, when i click on the turbo die to make an attack, the box shows 2, presumably because whatever it was looking for overflowed the array it was looking in or something, though i don't see an obvious JS error in firebug.

blackshadowshade:
So, this bug has a rather obvious cause:

Game.js, line 2615, we have:
var prevVal = die.sides;

This should actually be the swing value in the case of twin swing dice. Now to work out how to fix this ...

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1283/ (if it passes)

cgolubi1:
New jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1284/

cgolubi1:
Diff for bmapi.py:
```
diff --git a/tools/api-client/python/lib/bmapi.py b/tools/api-client/python/lib/bmapi.py
index 38d96ee..4f20d9d 100644
--- a/tools/api-client/python/lib/bmapi.py
+++ b/tools/api-client/python/lib/bmapi.py
@@ -142,7 +142,7 @@ class BMClient():
       args['previousGameId'] = use_prev_game
     return self._make_request(args)
 
-  def submit_turn(self, gameId, attackerIdx, defenderIdx, dieSelectStatus, attackType, roundNumber, timestamp, chat=''):
+  def submit_turn(self, gameId, attackerIdx, defenderIdx, dieSelectStatus, attackType, roundNumber, timestamp, turboVals, chat=''):
     args = {
       'type': 'submitTurn',
       'game': gameId,
@@ -155,6 +155,8 @@ class BMClient():
     }
     for statkey in sorted(dieSelectStatus.keys()):
       args['dieSelectStatus[%s]' % statkey] = dieSelectStatus[statkey]
+    for sizekey in sorted(turboVals.keys()):
+      args['turboVals[%s]' % sizekey] = turboVals[sizekey]
     return self._make_request(args)
 
   def submit_die_values(self, gameId, swingArray, optionArray, roundNumber, timestamp):
```

cgolubi1:
Shadowshade and i talked about the log message.  Right now, other size-changing dice have a log message like:
```
Attacker B(U=30) changed size from 30 to 15 sides, recipe changed from B(U=30) to (U=15), rerolled 29 => 5
```
The message shows both the pre- and post- reroll side-count/recipe, and the pre- and post-reroll value.

Whereas turbo has a two-line message:
```
glassonion performed Power attack using [(X=7)!:6] against [(20):6]; Defender (20) was captured
Turbo die (X=7)! changed size from 7 to 12 sides, recipe changed from (X=7)! to (X=12)!, rolled => 3
```

This looks weird because of the `rolled => 3` business, and also it no longer contains the pre-reroll value.

The issue is that by the time the second log message is rendered, the pre-reroll value is no longer available.  So that's not a big deal: proposed rememdy is to modify the first message to have a third block which is something like "; Attacker (X=7):6! rerolls from 6, changes size due to turbo".  and then remove the "=>" from the second message.

cgolubi1:
Summarizing a second conversation: i complained about the inconsistent return value format of `turboSizeArray`, which is similar to how `optRequestArray` is inconsistent (sometimes an array, sometimes a dict, sometimes keys are strings, sometimes they're integers).  Shadowshade tracked this down to `json_encode()` and has an idea about how to fix it.  But it's not really related to turbo, so he's going to make a separate issue for implementing that fix across the board.

cgolubi1:
Shadowshade and i discussed the question of whether the turbo UI selector should be a pulldown or an entry box.  Highlights:
* Shadowshade and i care about the fact that the `swingRequestArray` API element is sending a min/max pair, and the `turboSizeArray` API element is sending a range.  (In particular, i care about the fact that it's inconsistent.)  Other testers probably don't care about this.
* Turbo size selection is per-die for both swing and option, while between-rounds size selection is per-die for option but per-swing-type for swing.  This explains why it is correct for between-rounds size selection to use different API arrays for swing and option, and reasonable for turbo selection to use a single API array.
* Given that, we agree that the API array should enumerate all values for turbo for both swing and option.
* However, once the API is enumerating all values, the turbo UI can still identify that it's looking at a swing die, turn the range into a min-max, and display a min-max and a fill-in box.

This is the solution we recommend, and afawk everyone involved in this discussion will be okay with that.

Further, we think that the API for `swingRequestArray` should change to match this --- the backend should send a list of values and the UI should change it to min-max for display.  We'll make a separate issue for that.

cgolubi1:
In case it's not implicitly obvious, Shadowshade and i are trying to get Turbo to the point where it can be merged into master, i.e. where the current basic implementation is something we think we can put in the next release.  (After that it will be tested on staging and there will be time to iron out further issues before we actually release to prod.  But we will expect/hope to be able to move forward at that point, adding more pulls rather than reverting in order to get the next site release.)

To that end, i've been working on a bunch of replay-test-related pulls, and i'm pretty close to signing off that the turbo logic hasn't introduced any obvious regressions to non-turbo logic, and that automated play of novel turbo games doesn't blow up (whether or not it's correct).  Barring finding any new problems, i expect to be able to sign off on that within the next day.

We've also identified the selector/text-box issue (above) and the action log wording issue (above), both of which Shadowshade is working.  What else is still undecided and is a blocker?  Do folks have other concerns with throwing this over the fence?

dwvanstone:
I don't have any concerns about putting this up on dev. It all feels intuitive and clear to me as is.

irilyth:
I agree that this seems pretty close to ready. Following up on some old comments I'd made:

- https://github.com/buttonmen-dev/buttonmen/pull/2152#issuecomment-277572030 -- This would probably be better to fix as part of #943, but I think that adding Turbo makes the general #943 problem worse. That's not a reason to hold off on adding Turbo, but it might be a reason to think more about #943 soon.

- https://github.com/buttonmen-dev/buttonmen/pull/2152#issuecomment-280735627 -- I still think this looks a little weird, and that it'd look better if there were some sort of placeholder in the space where turbo selectors will pop up, rather than just blank space. Maybe something that matches what's in the die selection area -- perhaps a grayed-out selector, which becomes un-grayed-out and gets a red border when you select its corresponding die?

I don't think either of those is a blocker; I think the second would be good to fix before it goes to prod, but would be fine to fix after it goes to staging. (Or not to fix at all, if no one else thinks that all the blank space looks weird.)

blackshadowshade:
The turbo swing size selector has been turned into an input field, and the logging has been made more verbose on the topic of the pre-roll value for turbo dice.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1285/ (if it passes)

AdmiralJota:
I agree with irilyth that a greyed out selector would be preferable for Turbo dice that you're not using in the current attack.

cgolubi1:
Here you go - this responderTest reproduces the internal error i found in Turbo + Fire (sorry, Shadowshade:
```php
    /**
     * @depends test_request_savePlayerInfo
     */
    public function test_interface_game_048() {

        // responder003 is the POV player, so if you need to fake
        // login as a different player e.g. to submit an attack, always
        // return to responder003 as soon as you've done so
        $this->game_number = 48;


        $gameId = $this->verify_api_createGame(
            array(6, 1, 6, 5, 12, 11),
            'responder003', 'responder004', 'North Carolina', 'Wiseman', 3
        );

        $expData = $this->generate_init_expected_data_array($gameId, 'responder003', 'responder004', 3, 'SPECIFY_DICE');
        $expData['gameSkillsInfo'] = $this->get_skill_info(array('Fire', 'Ornery', 'Poison', 'Stinger', 'Trip', 'Turbo', 'Weak'));
        $expData['playerDataArray'][0]['button'] = array('name' => 'North Carolina', 'recipe' => 'pF(10) o(10) (V)! gt(V) h(V)', 'originalRecipe' => 'pF(10) o(10) (V)! gt(V) h(V)', 'artFilename' => 'northcarolina.png');
        $expData['playerDataArray'][1]['button'] = array('name' => 'Wiseman', 'recipe' => '(20) (20) (20) (20)', 'originalRecipe' => '(20) (20) (20) (20)', 'artFilename' => 'wiseman.png');
        $expData['playerDataArray'][0]['swingRequestArray'] = array('V' => array(6, 12));
        $expData['playerDataArray'][1]['waitingOnAction'] = FALSE;
        $expData['playerDataArray'][0]['activeDieArray'] = array(
            array('value' => NULL, 'sides' => 10, 'skills' => array('Poison', 'Fire'), 'properties' => array(), 'recipe' => 'pF(10)', 'description' => 'Poison Fire 10-sided die'),
            array('value' => NULL, 'sides' => 10, 'skills' => array('Ornery'), 'properties' => array(), 'recipe' => 'o(10)', 'description' => 'Ornery 10-sided die'),
            array('value' => NULL, 'sides' => NULL, 'skills' => array('Turbo'), 'properties' => array(), 'recipe' => '(V)!', 'description' => 'Turbo V Swing Die'),
            array('value' => NULL, 'sides' => NULL, 'skills' => array('Stinger', 'Trip'), 'properties' => array(), 'recipe' => 'gt(V)', 'description' => 'Stinger Trip V Swing Die'),
            array('value' => NULL, 'sides' => NULL, 'skills' => array('Weak'), 'properties' => array(), 'recipe' => 'h(V)', 'description' => 'Weak V Swing Die'),
        );
        $expData['playerDataArray'][1]['activeDieArray'] = array(
            array('value' => NULL, 'sides' => 20, 'skills' => array(), 'properties' => array(), 'recipe' => '(20)', 'description' => '20-sided die'),
            array('value' => NULL, 'sides' => 20, 'skills' => array(), 'properties' => array(), 'recipe' => '(20)', 'description' => '20-sided die'),
            array('value' => NULL, 'sides' => 20, 'skills' => array(), 'properties' => array(), 'recipe' => '(20)', 'description' => '20-sided die'),
            array('value' => NULL, 'sides' => 20, 'skills' => array(), 'properties' => array(), 'recipe' => '(20)', 'description' => '20-sided die'),
        );
        $expData['gameId'] = $gameId;
        $expData['playerDataArray'][0]['playerId'] = $this->user_ids['responder003'];
        $expData['playerDataArray'][1]['playerId'] = $this->user_ids['responder004'];

        $retval = $this->verify_api_loadGameData($expData, $gameId, 10);

        $this->verify_api_submitDieValues(
            array(8, 2, 1),
            $gameId, 1, array('V' => 8), NULL);

        $expData['activePlayerIdx'] = 0;
        array_unshift($expData['gameActionLog'], array('timestamp' => 'TIMESTAMP', 'player' => 'responder003', 'message' => 'responder003 set swing values: V=8'));
        array_unshift($expData['gameActionLog'], array('timestamp' => 'TIMESTAMP', 'player' => '', 'message' => 'responder003 won initiative for round 1. Initial die values: responder003 rolled [pF(10):6, o(10):1, (V=8)!:8, gt(V=8):2, h(V=8):1], responder004 rolled [(20):6, (20):5, (20):12, (20):11]. responder003 has dice which are not counted for initiative due to die skills: [gt(V=8)].'));
        $expData['gameActionLogCount'] = 2;
        $expData['gameState'] = "START_TURN";
        $expData['playerDataArray'][0]['activeDieArray'][0]['value'] = 6;
        $expData['playerDataArray'][0]['activeDieArray'][1]['value'] = 1;
        $expData['playerDataArray'][0]['activeDieArray'][2]['description'] = "Turbo V Swing Die (with 8 sides)";
        $expData['playerDataArray'][0]['activeDieArray'][2]['sides'] = 8;
        $expData['playerDataArray'][0]['activeDieArray'][2]['value'] = 8;
        $expData['playerDataArray'][0]['activeDieArray'][3]['description'] = "Stinger Trip V Swing Die (with 8 sides)";
        $expData['playerDataArray'][0]['activeDieArray'][3]['sides'] = 8;
        $expData['playerDataArray'][0]['activeDieArray'][3]['value'] = 2;
        $expData['playerDataArray'][0]['activeDieArray'][4]['description'] = "Weak V Swing Die (with 8 sides)";
        $expData['playerDataArray'][0]['activeDieArray'][4]['sides'] = 8;
        $expData['playerDataArray'][0]['activeDieArray'][4]['value'] = 1;
        $expData['playerDataArray'][0]['roundScore'] = 7;
        $expData['playerDataArray'][0]['sideScore'] = -22;
        $expData['playerDataArray'][0]['swingRequestArray'] = array();
        $expData['playerDataArray'][0]['turboSizeArray'] = array("2" => array(6, 7, 8, 9, 10, 11, 12));
        $expData['playerDataArray'][1]['activeDieArray'][0]['value'] = 6;
        $expData['playerDataArray'][1]['activeDieArray'][1]['value'] = 5;
        $expData['playerDataArray'][1]['activeDieArray'][2]['value'] = 12;
        $expData['playerDataArray'][1]['activeDieArray'][3]['value'] = 11;
        $expData['playerDataArray'][1]['roundScore'] = 40;
        $expData['playerDataArray'][1]['sideScore'] = 22;
        $expData['playerWithInitiativeIdx'] = 0;
        $expData['validAttackTypeArray'] = array("Power", "Skill", "Trip");

        $retval = $this->verify_api_loadGameData($expData, $gameId, 10);

        $this->verify_api_submitTurn(
            array(6, 8, 3, 8),
            'responder003 performed Skill attack using [o(10):1,(V=8)!:8,gt(V=8):2,h(V=8):1] against [(20):12]; Defender (20) was captured; Attacker o(10) rerolled 1 => 6; Attacker gt(V=8) rerolled 2 => 8; Attacker h(V=8) changed size from 8 to 6 sides, recipe changed from h(V=8) to h(V=6), rerolled 1 => 3. Turbo die (V=8)! changed size from 8 to 11 sides, recipe changed from (V=8)! to (V=11)!, rolled => 8. ',
            $retval, array(array(0, 1), array(0, 2), array(0, 3), array(0, 4), array(1, 2)),
            $gameId, 1, 'Skill', 0, 1, '', array(2 => 11));

        $expData['activePlayerIdx'] = 1;
        array_unshift($expData['gameActionLog'], array('timestamp' => 'TIMESTAMP', 'player' => 'responder003', 'message' => 'responder003 performed Skill attack using [o(10):1,(V=8)!:8,gt(V=8):2,h(V=8):1] against [(20):12]; Defender (20) was captured; Attacker o(10) rerolled 1 => 6; Attacker gt(V=8) rerolled 2 => 8; Attacker h(V=8) changed size from 8 to 6 sides, recipe changed from h(V=8) to h(V=6), rerolled 1 => 3'));
        array_unshift($expData['gameActionLog'], array('timestamp' => 'TIMESTAMP', 'player' => 'responder003', 'message' => 'Turbo die (V=8)! changed size from 8 to 11 sides, recipe changed from (V=8)! to (V=11)!, rolled => 8'));
        $expData['gameActionLogCount'] = 4;
        $expData['playerDataArray'][0]['activeDieArray'][1]['value'] = 6;
        $expData['playerDataArray'][0]['activeDieArray'][2]['description'] = "Turbo V Swing Die (with 11 sides)";
        $expData['playerDataArray'][0]['activeDieArray'][2]['properties'] = array("HasJustTurboed");
        $expData['playerDataArray'][0]['activeDieArray'][2]['sides'] = 11;
        $expData['playerDataArray'][0]['activeDieArray'][3]['value'] = 8;
        $expData['playerDataArray'][0]['activeDieArray'][4]['description'] = "Weak V Swing Die (with 6 sides)";
        $expData['playerDataArray'][0]['activeDieArray'][4]['properties'] = array("HasJustShrunk");
        $expData['playerDataArray'][0]['activeDieArray'][4]['sides'] = 6;
        $expData['playerDataArray'][0]['activeDieArray'][4]['value'] = 3;
        $expData['playerDataArray'][0]['capturedDieArray'][0]['properties'] = array("WasJustCaptured");
        $expData['playerDataArray'][0]['capturedDieArray'][0]['recipe'] = "(20)";
        $expData['playerDataArray'][0]['capturedDieArray'][0]['sides'] = 20;
        $expData['playerDataArray'][0]['capturedDieArray'][0]['value'] = 12;
        $expData['playerDataArray'][0]['roundScore'] = 27.5;
        $expData['playerDataArray'][0]['sideScore'] = -1.7;
        $expData['playerDataArray'][0]['waitingOnAction'] = false;
        $expData['playerDataArray'][1]['activeDieArray'][2]['value'] = 11;
        array_pop($expData['playerDataArray'][1]['activeDieArray']);
        $expData['playerDataArray'][1]['roundScore'] = 30;
        $expData['playerDataArray'][1]['sideScore'] = 1.7;
        $expData['playerDataArray'][1]['waitingOnAction'] = true;
        $expData['validAttackTypeArray'] = array("Power", "Skill");

        $retval = $this->verify_api_loadGameData($expData, $gameId, 10);

        $_SESSION = $this->mock_test_user_login('responder004');
        $this->verify_api_submitTurn(
            array(11),
            'responder004 performed Power attack using [(20):11] against [gt(V=8):8]; Defender gt(V=8) was captured; Attacker (20) rerolled 11 => 11. ',
            $retval, array(array(1, 2), array(0, 3)),
            $gameId, 1, 'Power', 1, 0, '', array());

        $expData['activePlayerIdx'] = 0;
        array_unshift($expData['gameActionLog'], array('timestamp' => 'TIMESTAMP', 'player' => 'responder004', 'message' => 'responder004 performed Power attack using [(20):11] against [gt(V=8):8]; Defender gt(V=8) was captured; Attacker (20) rerolled 11 => 11'));
        $expData['gameActionLogCount'] = 5;
        $expData['playerDataArray'][0]['activeDieArray'][2]['properties'] = array();
        $expData['playerDataArray'][0]['activeDieArray'][3]['description'] = "Weak V Swing Die (with 6 sides)";
        $expData['playerDataArray'][0]['activeDieArray'][3]['recipe'] = "h(V)";
        $expData['playerDataArray'][0]['activeDieArray'][3]['sides'] = 6;
        $expData['playerDataArray'][0]['activeDieArray'][3]['skills'] = array("Weak");
        $expData['playerDataArray'][0]['activeDieArray'][3]['value'] = 3;
        array_pop($expData['playerDataArray'][0]['activeDieArray']);
        $expData['playerDataArray'][0]['capturedDieArray'][0]['properties'] = array();
        $expData['playerDataArray'][0]['roundScore'] = 23.5;
        $expData['playerDataArray'][0]['sideScore'] = -9.7;
        $expData['playerDataArray'][0]['waitingOnAction'] = true;
        $expData['playerDataArray'][1]['capturedDieArray'][0]['properties'] = array("WasJustCaptured");
        $expData['playerDataArray'][1]['capturedDieArray'][0]['recipe'] = "gt(V)";
        $expData['playerDataArray'][1]['capturedDieArray'][0]['sides'] = 8;
        $expData['playerDataArray'][1]['capturedDieArray'][0]['value'] = 8;
        $expData['playerDataArray'][1]['roundScore'] = 38;
        $expData['playerDataArray'][1]['sideScore'] = 9.7;
        $expData['playerDataArray'][1]['waitingOnAction'] = false;

        $retval = $this->verify_api_loadGameData($expData, $gameId, 10);

        $this->verify_api_submitTurn(
            array(),
            'responder003 chose to perform a Skill attack using [(V=11)!:8] against [(20):11]; responder003 must turn down fire dice to complete this attack. ',
            $retval, array(array(0, 2), array(1, 2)),
            $gameId, 1, 'Skill', 0, 1, '', array(2 => 9));

        array_unshift($expData['gameActionLog'], array('timestamp' => 'TIMESTAMP', 'player' => 'responder003', 'message' => 'responder003 chose to perform a Skill attack using [(V=11)!:8] against [(20):11]; responder003 must turn down fire dice to complete this attack'));
        $expData['gameActionLogCount'] = 6;
        $expData['gameState'] = "ADJUST_FIRE_DICE";
        $expData['playerDataArray'][0]['activeDieArray'][2]['properties'] = array("IsAttacker");
        $expData['playerDataArray'][0]['turboSizeArray'] = array();
        $expData['playerDataArray'][1]['activeDieArray'][2]['properties'] = array("IsAttackTarget");
        $expData['validAttackTypeArray'] = array("Skill");

        $retval = $this->verify_api_loadGameData($expData, $gameId, 10);

        $this->verify_api_adjustFire(
            array(),
            'responder003 chose to abandon this attack and start over. ',
            $retval, $gameId, 1, 'cancel', array(), array());

        array_unshift($expData['gameActionLog'], array('timestamp' => 'TIMESTAMP', 'player' => 'responder003', 'message' => 'responder003 chose to abandon this attack and start over'));
        $expData['gameActionLogCount'] = 7;
        $expData['gameState'] = "START_TURN";
        $expData['playerDataArray'][0]['activeDieArray'][2]['properties'] = array();
        $expData['playerDataArray'][0]['turboSizeArray'] = array("2" => array(6, 7, 8, 9, 10, 11, 12));
        $expData['playerDataArray'][1]['activeDieArray'][2]['properties'] = array();
        $expData['validAttackTypeArray'] = array("Power", "Skill");

        $retval = $this->verify_api_loadGameData($expData, $gameId, 10);

        $this->verify_api_submitTurn(
            array(),
            'responder003 chose to perform a Power attack using [h(V=6):3] against [(20):5]; responder003 must turn down fire dice to complete this attack. ',
            $retval, array(array(0, 3), array(1, 1)),
            $gameId, 1, 'Power', 0, 1, '', array());

        array_unshift($expData['gameActionLog'], array('timestamp' => 'TIMESTAMP', 'player' => 'responder003', 'message' => 'responder003 chose to perform a Power attack using [h(V=6):3] against [(20):5]; responder003 must turn down fire dice to complete this attack'));
        $expData['gameActionLogCount'] = 8;
        $expData['gameState'] = "ADJUST_FIRE_DICE";
        $expData['playerDataArray'][0]['activeDieArray'][3]['properties'] = array("IsAttacker");
        $expData['playerDataArray'][0]['turboSizeArray'] = array();
        $expData['playerDataArray'][1]['activeDieArray'][1]['properties'] = array("IsAttackTarget");
        $expData['validAttackTypeArray'] = array("Power");

        $retval = $this->verify_api_loadGameData($expData, $gameId, 10);

        // BUG encountered while logging game: API adjust_fire_dice(91, turndown, [0], [4], 1, 1491288709) unexpectedly failed: NULL
    }
```
(Shadowshade, let me know if it's not clear what that "BUG" reported API call means you should do.)

cgolubi1:
Also: you have some jshint failures in http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1285/console

src/engine/BMGameAction.php
```diff
@@ -501,8 +501,13 @@ protected function messageAttacker($preAttackDice, $postAttackDice, $attackType
                 $postEventsAttacker,
                 $this->message_recipe_change($attackerInfo, $postInfo)
             );
-            if (!array_key_exists('forceHideDieReroll', $postInfo) ||
-                ($attackType == 'Trip')) {
+            if (array_key_exists('forceHideDieReroll', $postInfo) &&
+                !($attackType == 'Trip')) {
+                $this->message_append(
+                    $postEventsAttacker,
+                    'rerolls from ' . $attackerInfo['value'] . ', changes size due to turbo'
```
cgolubi1: Nit: i think everything else in the log is in past tense ("rerolled", "changed"), so this should be too.

cgolubi1:
Shadowshade: reminder that the UI commits you made on Apr 3rd failed jshint so i can't bring them up on the dev site.  I'd be happy to do so even if you haven't yet fixed the `$turboCache` thing - we can treat those separately.

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1288/ (if it passes)

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1290/ (if it passes)

blackshadowshade:
So, the turbo cache is now being cleared when abandoning a turn needing fire support.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1291/ (if it passes)

blackshadowshade:
@AdmiralJota, would you be able to give us the appropriate CSS for a greyed inactive turbo selector?

AdmiralJota:
For the CSS (which will grey out the text), you probably want to add a rule like: `.disabled_turbo { color: grey; }`. Then you can add/remove that class on the turbo_subspan element with jquery like `turboSubspan.addClass('disabled_turbo');` and `turboSubspan.removeClass('disabled_turbo');`.

To actually disable the control itself (the textbox or dropdown, whatever it ends up being), I believe you'll want to do something like `turboElement.prop("disabled", true);` and `turboElement.prop("disabled", false);`

blackshadowshade:
So, I've changed the selectors to be grey instead of invisible.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1294/ (if it passes)

src/engine/BMGame.php
```diff
@@ -1882,13 +1882,13 @@ public function set_turbo_sizes(array $turboSizeArray) {
             $die = $this->playerArray[$playerIdx]->activeDieArray[$dieIdx];
 
             if (!$die->has_skill('Turbo')) {
-                $this->message = 'Cannot set turbo size for a non-turbo die';
-                return FALSE;
+                // ignore any turbo values supplied for non-turbo dice
```
cgolubi1: Can you say more about why this change is needed?  I don't think random_ai was submitting invalid turbo values?

blackshadowshade: So, there are two changes here.

The second one, where the backend accepts and ignores turbo values from turbo dice that are not being changed, is necessary because of the UI change that leaves the selectors visible but disabled.

The first, where the backend accepts and ignores turbo values from non-turbo dice, is not necessary, but was done for consistency, so that this function consistently accepts irrelevant inputs and ignores them silently. If you want me to reverse this change, I can do so.

blackshadowshade: So, I was thinking about this, and I guess it would also be possible to change the implementation of the form submission to submit only values from non-disabled selectors. That would mean that I could revert both of the backend changes here.

irilyth:
I like the latest changes. One more suggestion (or maybe pair of suggestions): Now that the selectors for non-selected dice are grayed out, I think it'd be nicer to keep the selection area visible all the time; it'd remind you that you have Turbo swing dice. To make it more obvious when you have a decision to make involving them, could we highlight the ones corresponding to selected dice with a red box, just like the selected dice?

blackshadowshade:
@cgolubi1, how's the automated testing of this pull coming along? Are we able to go live with this pull yet?

blackshadowshade:
Form submission now only submits turbo values for non-disabled selectors. I've reverted the changes to the backend validation.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1295/ (if it passes)

deploy/amp/database_initialise.php
```diff
@@ -8,9 +8,9 @@
 // james: uncomment the following lines to also initialise the
 //        primary 'buttonmen' database
 //
-//        recreate_database('buttonmen');
-//        $conn = db_connect('buttonmen');
-//        run_all_sql_files($conn);
+        recreate_database('buttonmen');
+        $conn = db_connect('buttonmen');
+        run_all_sql_files($conn);
```
cgolubi1: Note that you've committed this change here - i don't care, nothing live uses this file, just mentioning it in case you do care and didn't mean to do that, in which case revert it the next time you push a change.

blackshadowshade:
Okay, I've recommented the live database recreation in the AMP script.

Did you want me to rebase this pull request?

blackshadowshade:
This pull request has now been rebased. Fingers crossed, Jenkins should pass:

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1296/ (if it passes)

cgolubi1:
Okay!  I've had 2100 successful new and replayed games (each), and i think it's go time.  Merging turbo into master.  Shadowshade, i'll have a pull up shortly with the random_ai changes needed to support turbo.

cgolubi1:
Hmm, sorry, hang on a sec - i noticed that the currently active test game might be stuck, and i'm looking into that.  It's probably nothing, but i want to double-check.

cgolubi1:
Ahh, the game is IIconfused vs Echo.  That's gonna end well.  :>)  Well, i'm gonna hang out and watch that and double-check that it eventually times out after 1000 turns like we think we have implemented now.

cgolubi1:
I want to see what happens with the IIconfused game, but roundmax behavior has nothing to do with Turbo, so i am in fact merging this.

PR #2155, opened 2017-01-17
Added a script for calculating head-to-head win percentages

irilyth:
http://jenkins.buttonweavers.com:8080/job/buttonmen-irilyth/31/

### DIFF
--- a/tools/api-client/python/head-to-head
+++ b/tools/api-client/python/head-to-head
@@ -1,0 +1,146 @@
+#!/usr/bin/env python
+
+##### head-to-head.py
+#
+# Show head-to-head records between a protagonist (a button or a player)
+# and one or more opponents.
+
+### Imports
+
+# Import stuff from the future.
+
+from __future__ import division
+
+# Import regular stuff.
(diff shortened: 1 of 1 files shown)


cgolubi1:
Hmm, i thought we thought executable script names didn't end with `.py` (or whatever) suffixes.  Anyway the others in this directory don't.

cgolubi1:
Also:

I think your parser should include something like:
```
  parser.add_option(
    "-c", "--config",
    help="config file containing site parameters",
    type=str, default=".bmrc"
  )
```
the way the other scripts do, so the config file doesn't have to be .bmrc.

Other than that, this appears to work, and i'm fine merging it with those changes (add the -c option, make it executable, make it not end in .py), or let's discuss more if you disagree with any of those.

irilyth:
Sure, seems reasonable.

http://jenkins.buttonweavers.com:8080/job/buttonmen-irilyth/33/changes (if it passes)

cgolubi1:
Looks good.

PR #2158, opened 2017-02-26
Implement all current within-gameplay API methods in the python API client

cgolubi1:
* Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/481/ (if it passes)

I'm submitting a PR for this now because turbo changes the API enough that i need to clean up RandomAI's replay functionality a lot in order to replay-test turbo.  This library represents the major part of RandomAI which is a change to existing code files, rather than being new files.  So i'd like to get it reviewed and accepted to get my workspace in sync with the repo.

As a side effect, obviously if anyone else wants to write a bot in python, this should ideally be helpful to them.

### DIFF
--- a/tools/api-client/python/lib/bmapi.py
+++ b/tools/api-client/python/lib/bmapi.py
@@ -19,6 +19,11 @@ class BMAPIResponse():
     self.data = response_dict['data']
     self.message = response_dict['message']
     self.status = response_dict['status']
+    self.rand_vals = None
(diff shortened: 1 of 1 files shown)


blackshadowshade:
As far as I can see, this looks okay. I figure that you'll stumble across problems during your testing if anything is broken here, and it doesn't impact on game code, so I'm happy to accept this.

PR #2166, opened 2017-04-07
In jenkins, run grunt scripts before processing build.xml

cgolubi1:
Technically, we don't config-manage config.xml for each individual jenkins project, so i manually made this change to `/srv/jenkins-jobs/buttonmen-cgolubi1/config.xml`, and then reran jenkins (not on this branch, but it doesn't matter) - http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/487/console - where you can see it ran jshint at the beginning instead of the end.

If this is approved, i'll deploy the template change in the usual way, and reconfigure the existing jobs.

### DIFF
--- a/deploy/vagrant/modules/jenkins/templates/buttonmen_config.xml.erb
+++ b/deploy/vagrant/modules/jenkins/templates/buttonmen_config.xml.erb
@@ -27,13 +27,13 @@
   <triggers class="vector"/>
   <concurrentBuild>false</concurrentBuild>
   <builders>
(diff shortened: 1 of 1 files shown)


cgolubi1:
FWIW, no, i don't actually know why we made this a separate piece of the jenkins config rather than having it controlled by build.xml.  It's possible that that would work too, but i'd have to do more than zero seconds of research to find out.

blackshadowshade:
Yay for running JSHint first! :)

PR #2181, opened 2017-04-30
Actually include SWING_RANGES variable used by trip turbo code

cgolubi1:
Another RandomAI-only fix, which should address a stacktrace exception i found on the site that records new replay games today.

Sorry for the sloppiness.


blackshadowshade:
Fine.

PR #2184, opened 2017-05-01
Added Rush skill

### LINKED ISSUE #2176: Rush dice
Rush dice have been added in the Kickstarter. Our current knowledge about this skill is directly from James Ernest:

> Rush Dice are exactly like Focus Dice but are worth zero points if you keep them. A small handicap for an otherwise strictly better die type.

My suggestion for the recipe encoding of this skill is to use '#'.

blackshadowshade:
Fixes #2176.

This pull request is complete in itself, but has no tests of actual gameplay yet, since at the time of writing, there are no finalised recipes including Rush. It also uses a skill description that will need to be updated when the official skill description is released.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1302/ (if it passes)


blackshadowshade:
Looks like this is on hold, while James Ernest decides whether he wants to use it or not.

PR #2187, opened 2017-05-01
Added new sets: The Core, West Side, Delta, Uptown

blackshadowshade:
Addresses part of #2186.

Requires database updates 02186_kickstarter_01.sql

PR #2194, opened 2017-05-05
Added Rush skill, take 2

### LINKED ISSUE #2176: Rush dice

blackshadowshade:
Fixes #2176.

This pull request is complete in itself, but has no tests of actual gameplay yet. It also uses a skill description that will need to be updated when the official skill description is released. Rush recipes will be added as part of #2187.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1318/


blackshadowshade:
Closing this one too, since the definition of Rush will be changing again.

Issue #2197, opened 2017-05-06
Python API client should be Python 3 compatible


irilyth:
It'd be nice if the Python API client was Python 3 compatible.

irilyth:
Some useful links I found:

- https://wiki.python.org/moin/PortingToPy3k/BilingualQuickRef
- http://python3porting.com/noconv.html
- http://python-future.org/compatible_idioms.html

I might take a stab at this if it's not too impossibly hard.

irilyth:
Some useful links I found:

- https://wiki.python.org/moin/PortingToPy3k/BilingualQuickRef
- http://python3porting.com/noconv.html
- http://python-future.org/compatible_idioms.html

I might take a stab at this if it's not too impossibly hard.

danlangford:
i have been using the python client more and more. i would love to submit some of the enhancements i have added (search_game_history) and am wondering is 2.x compatibility required? Python3 was released in 2008 and should be very easy to obtain on any machine at this point. what are your thoughts? I can double back on my enhancements to insure 2.7 compatibility but it will take time.

blackshadowshade:
The main people using the Python client are @cgolubi1, @irilyth, and @AdmiralJota, so you'll want to coordinate your chat with them. Since any Python changes don't impact the game site directly, such changes should be relatively easy, as long as they don't break what others are doing with the client.

irilyth:
I think I had a branch with some Python 3 conversion, I can dig it up if you might be interested in finishing this off. My recollection is that something in it broke with something else @AdmiralJota was doing as part of Steven Universe playtesting, but maybe now that the SU set is part of the site, we don't need that specific code any more? I don't remember the details, and don't care a lot, but can help if it turns out that I have anything helpful.

danlangford:
thanks. ill wait for @AdmiralJota to weigh in. my concern was mostly about me wanting to send over some PRs that are probably not Python 2 compliant. i dont think at this time @irilyth you need to worry about digging up that old code. thanks

danlangford:
its my opinion that this issue can be closed with #2667 merged. I will work on a PR for what I have to contribute for #2668

irilyth:
I agree that this was resolved by #2885 (which superseded #2667), so let's close this one out, and I'll make a new issue for some follow-on work. Thanks @cgolubi1 and @danlangford!

PR #2198, opened 2017-05-06
Added a bunch of stuff to make a bunch of things Python 3 compatible

irilyth:
http://jenkins.buttonweavers.com:8080/job/buttonmen-irilyth/38/ (if it passes)

Addresses much of #2197.

I've tested a bunch of this stuff by running it with 'python2' and 'python3' on my Ubuntu 12.04 box, which has Python 2.7.3 and Python 3.2.3.

I didn't touch prep_replay_games, replay_loop, test_log_games, or update_replay_games, because I wasn't sure how to test them safely.

### DIFF
--- a/tools/api-client/python/create_game
+++ b/tools/api-client/python/create_game
@@ -90,15 +90,15 @@ def create_games(b, buttons, opponents):
   for opponent in opponents:
     pbutton = random_array_element(buttons)
     obutton = random_array_element(buttons)
-    print "%s (%s) vs. %s (%s)" % (b.username, pbutton, opponent, obutton)
+    print("%s (%s) vs. %s (%s)" % (b.username, pbutton, opponent, obutton))
     time.sleep(3)
-    print b.wrap_create_game(pbutton, obutton, opponent)
+    print(b.wrap_create_game(pbutton, obutton, opponent))
 
 def create_random_game(opponent, b, players=False, usebuttons=False):
   if not players:
     players = lookup_other_players(b)
   if not opponent in players:
-    raise ValueError, "Opponent %s is not known on site" % opponent
+    raise(ValueError, "Opponent %s is not known on site" % opponent)
   buttons = b.wrap_load_button_names()
   while True:
     if usebuttons:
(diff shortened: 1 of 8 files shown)


tools/api-client/python/create_game
```diff
@@ -90,15 +90,15 @@ def create_games(b, buttons, opponents):
...
     time.sleep(3)
-    print b.wrap_create_game(pbutton, obutton, opponent)
+    print (b.wrap_create_game(pbutton, obutton, opponent))
 
 def create_random_game(opponent, b, players=False, usebuttons=False):
   if not players:
     players = lookup_other_players(b)
   if not opponent in players:
-    raise ValueError, "Opponent %s is not known on site" % opponent
+    raise (ValueError, "Opponent %s is not known on site" % opponent)
```
cgolubi1: Hmm, do we like `raise (ValueError` or `raise(ValueError` - i've usually done the latter (and would presumably do the same with print()), on the theory that if it looks like a function, it should be whitespaced like a function.  Whatcha think?

tools/api-client/python/lib/bmapi.py
```diff
@@ -3,19 +3,47 @@
...
+  from cookielib import LWPCookieJar
+
+try:
+  from urllib.parse import urlparse, urlencode
+  from urllib.request import urlopen, Request, HTTPCookieProcessor, build_opener, install_opener
+  from urllib.error import HTTPError
+except ImportError:
+  from urlparse import urlparse
+  from urllib import urlencode
+  from urllib2 import urlopen, Request, HTTPError, HTTPCookieProcessor, build_opener, install_opener
```
cgolubi1: Let's maybe be safe and make this a separate try/except block for each different import, rather than assuming they'll all succeed or fail identically?  Or is this definitely risk-free?

blackshadowshade:
This will need to be rebased to remove conflicts.

irilyth:
> This will need to be rebased to remove conflicts.

Done! With Chaos's whitespace suggestions as well.

irilyth:
> Let's maybe be safe and make this a separate try/except block for each different import, rather than assuming they'll all succeed or fail identically?  Or is this definitely risk-free?

I don't know enougha about the details to say for sure, so I've split 'em up, which seems pretty safe.

irilyth:
http://jenkins.buttonweavers.com:8080/job/buttonmen-irilyth/53/ (if it passes)

blackshadowshade:
You'll probably want to rebase this down to one commit too.  :)

blackshadowshade:
So, as far as I can see, this needs to be rebased, then CircleCI needs to run.

@irilyth, have you set up your CircleCI account to run a build automatically for new commits?

irilyth:
I have, and it's running now!

cgolubi1:
Why didn't i do minimal replay testing with this, given that it's touching one of the python files replay_loop uses?  Why not do so now.

cgolubi1:
Okay, looks fine.  Let's merge this.

PR #2203, opened 2017-05-13
Added floating header for forums

### LINKED ISSUE #2164: Duplicate Mark Thread as Read at top
Hi Devs,

When I am browsing forum posts using next post, I often arrive at threads that I know I don't care about.  For example tournaments I'm not in.  Currently I have to scroll all the way to the bottom to mark that thread as read.  I'd like to be able to mark that thread as read without scrolling.  Could we get a second "Mark thread as read" button at the top of the thread as well as at the bottom of the thread.

Thank you,
Nicole

### LINKED ISSUE #2165: How about reverse-chronological order for forum threads
Hi Devs,

I feel like this one has a low chance of getting approved, but I want to put it out there anyway:

Many of our forum threads are very long, and I hate scrolling.  What if we displayed forum threads in reverse chronological order.

Advantage of reverse-chronological:  New posts right at the top!
Advantage of chronological:  Context of the thread is set at the top.

Thank you,
Nicole

blackshadowshade:
Fixes #2165. Fixes #2164.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1329/


blackshadowshade:
Note that I've just added a stub for the corresponding QUnit test here.

AdmiralJota:
It occurs to me that this should also resolve #2164.

blackshadowshade:
Okay, added fixes statement to the initial comment.

cgolubi1:
I have no idea whether this is a good solution, but i'll load a dev site for it.

irilyth:
Sure: http://transient.dev.buttonweavers.com/

cgolubi1:
I set Nala's account on the dev site active at Jota's request:
```
mysql> update player set status_id=1 where name_ingame='Nala';
Query OK, 1 row affected (0.01 sec)
Rows matched: 1  Changed: 1  Warnings: 0

mysql> 
```

cgolubi1:
Jota thinks this doesn't break anything, but we'd like more eyeballs on the dev site to have confidence before merging it.  Can someone else scroll around on the forums some and try some things?  I think this would be neat to merge for June if we're comfortable with the change.

dwvanstone:
The header on the front page (Button Men Forums) floats too high on the opening page. As soon as I scroll it snaps back to the usual location. I'm on a Mac OS X using latest Safari.

bmforums

blackshadowshade:
So, I can duplicate devious' bug, namely that the header is too high on the main forum page until a scroll happens.

From https://github.com/jeremypeters/floatThead/blob/master/README.md:

> Safari and mobile safari is not fully supported. It might work, or it might not, depending on your markup and safari version.

It seems to be related to the transform translateY property, which I'm seeing in the CSS at 58px.

blackshadowshade:
I've tested this on the Android stock browser and it looks okay.

cgolubi1:
shadowshade: what help do you need in order to resolve the safari bug devious reported?

blackshadowshade:
I don't feel equipped at the moment to deal with this bug. To get me to the stage where I could even think about addressing this bug, I would need to know roughly how (and where in the code) the calculation of the initial vertical offset is calculated.

mkoryak:
if you can get me a jsfiddle reproducing this bug, I can look at it. safari is the devil though, so no promises :)

blackshadowshade:
@mkoryak, many thanks for the offer. I'll get @AdmiralJota on to it, and we'll see if we can get you a minimal example that reproduces this bug.

irilyth:
Firefox 53.0.3 on Ubuntu 16.04, it looks like the header floats lower than it should:

bm

Is there supposed to be a gap between the header and the top of the page?

Also, when I scroll down by a page (with the space bar, or the page-down key), it scrolls the full size of the screen, such that the floating header covers the thing that was just off the bottom of the page before. Here's an attempt to show that... Before I page down:

bm

You can see the very top of "Curabitur etc" at the bottom of the screen. After I page down:

bm

You can see that whole first paragraph of the post is now behind the floating header.

irilyth:
Chrome 58.0.3029.110 on Ubuntu 16.04 seems to have the first problem I describe above (header isn't flush with the page), and not the second (on Chrome, page-down brings what was just below the bottom of the window up to being the first thing after the header, rather than under the header).

irilyth:
Chrome 59.0.3071.125 on Android seems to have the first problem I describe above (header isn't flush with the page), and the second doesn't really apply since there's no "page down" scrolling on the touch screen.

Maybe the header isn't *supposed* to be flush with the top of the screen? I think it's weird that it's not, but what do I know.

AdmiralJota:
As for the header not being flush with the top of the page, that's intentional, since it seemed to me that it would look bad to not have any margin there at all. Do folks disagree?

As for the paging down, that's an interesting one. I'm not sure if I've ever seen a site with a floating header that handled that... hmm.

irilyth:
> As for the paging down, that's an interesting one. I'm not sure if I've ever seen a site with a floating header that handled that... hmm.

I don't know if it's a "floating header" or some other page structure, but in Gmail, for example, if you page down it seems to do the right thing. Twitter too, and https://www.washingtonpost.com/; I didn't find anything immediately that *didn't* work, although I'm sure I've seen sites before that didn't.

cgolubi1:
Incidentally, this pull has a merge conflict now - i'm guessing with the cache change.

mkoryak:
Mention me if you need help and have some place on the internet that I can see the problem

blackshadowshade:
@mkoryak, we have a test site up at http://transient.dev.buttonweavers.com/. If you have trouble signing up because your email provider rejects our confirmation email, let us know and we'll manually activate your account.

We're talking about the main forum page, which you get to when you click the "Forum" button in the header. The floating header appears to be displaying initially at the wrong distance from the top, but only in Safari. I gather that @AdmiralJota has been too busy to work on generating a JSFiddle for this.

mkoryak:
That's understandable, I will take a look on Monday

mkoryak:
You are seeing that spacing because you are using border-spacing in your table to separate the rows.

You are also not explicitly setting the `border-spacing` and using the default value that the browser uses. This is why the spacing value is different in different browsers. I would suggest that you find a value that pleases you and make it an explicit style. 

This fix will work even if you don't make it an explicit rule by getting the value from the element style:

```js
var borderSpacing = parseInt( $('table.floatable').css('border-spacing').split(' ')[1]);
 $('table.floatable').floatThead({
   top: -borderSpacing
});
```

mkoryak:
ok, i lied a little bit,  i am not going to edit my previous comment because it is partially correct. 
 
I noticed that you do have an explicit `border-spacing` set, but the selector is `div.forum > table` which does not apply the floated table headers when it becomes floated.

this breaks the implicit contract required by my plugin, it expects that the ripped out table header in the new table be styled exactly like the original, otherwise its impossible to guarantee that column widths match. I think in your case it didnt break, but I do not recommend doing it. 

You are also explicitly styling the floated header. You are setting `border-spacing` on it, but its too late. By the time it exists the plugin has already calculated it's location based on the original `border-spacing`.

So the implicit style explanation is out. I would recommend not using `border-spacing` to make my plugin happy if it gives you trouble on different browsers. `border-spacing` support was more of an after thought because almost no one uses it anymore. 

Looking at your html, it looks like it will not be trivial to make that change. You will need to put a div inside each table cell and style those, or you can hack around it by adding "spacer" rows between the content :)

mkoryak:
here what im talking about:

```js
$('table.floatable').floatThead('destroy'), $('table.floatable').floatThead('destroy').css({'border-spacing': '0', 'border-collapse': 'seperate'}).find('tbody tr').each(function() { $(this).before("<tr><td style='border:none'></td><td style='border:none'></td>") }).end().floatThead();
```

also I noticed that the location of the header is not correct on load, this is because, i assume you are creating the top nav via javascript. After you modify the DOM and make the nav take up more vertical space you need to tell floatThead to update itself, because its not in the location where it thought it was.

you need to call `$('table.floatable').trigger('reflow')` on the table

blackshadowshade:
So, it sounds like that last line will fix our major bugbear, namely the incorrect initial positioning of the header. The rest can wait for the moment, but @AdmiralJota, if you can suggest where the reflow should occur in the code, I'll implement it, and we can then test it.

blackshadowshade:
I've tried adding this reflow to various portions of the code, including the end of Login.showLoginHeader(), Login.arrangePage(), Forum.arrangePage(), Forum.showPage(), and even the end of forum.html. Nothing seems to work consistently.

mkoryak:
Try it with a settimeout. Or better yet init the plugin with one. It needs
to init after the nav hight is stable

blackshadowshade:
Ha, success!

I added

    setTimeout(function(){ $('table.floatable').trigger('reflow'); }, 2000);

to the end of the success call in forum.html, and chose the delay to be 2000 ms just to see that this is doing the right thing. It always corrects the header position when the header starts out at the wrong place. Now, I guess I just need to play around with the delay times to find out what a reasonable delay time is.

blackshadowshade:
So, I've set the delay to 100 ms, which is short enough to be perceptually "just part of page loading", but appears to be long enough to be after the login header has loaded.

I'm open to comments/suggestions about whether the placed I added the reflow code is really appropriate.

@cgolubi1, if you wouldn't mind redeploying this at your convenience. I've rebased it into one commit because of the conflicts earlier, so you'll either need to go from scratch or do a forced pull.

blackshadowshade:
BTW, thanks so much for your help, @mkoryak!

mkoryak:
Theoretically you should be able to find a place or places in your code to
do this without the delay, but it's probably not worth your time. What is
the worst browser that you support? Does it have MutationObserver support?

blackshadowshade:
From what I've just read, we're probably supporting browsers older than MutationObserver. The idea is to have our game run even on decrepit pieces of hardware like my Mac Powerbook G4 with a PowerPC chip.  :)

mkoryak:
Is this a very popular game?

blackshadowshade:
200-ish users log in per week, roughly.

cgolubi1:
I'll redeploy when i get a chance, probably tomorrow.  Unless anyone objects, i'll tear down and rebuild the dev site.

cgolubi1:
Relaunched the site - the new one is 54.159.47.9 and we'll repoint the DNS when we get a second.

irilyth:
This seems to still have the scrolling bug on Firefox.

blackshadowshade:
@irilyth, can you be specific about which bug you're referring to? If it's the one where the header appears to be lower than you expected, that's currently not a bug in my mind, just a difference of opinion with regards to design.

blackshadowshade:
To be specific, all I addressed was the point that devious mentioned about the header not appearing in the correct place in Safari. I've not attempted to address any of the other points.

blackshadowshade:
From what I see here

http://www.dynamicdrive.com/forums/showthread.php?74649-Javascript-to-control-the-length-of-page-down-up-scrolling

the thing to do is to intercept the PageUp/PageDown keys and control the behaviour explicitly, if we want to take the size of the floating header into account.

irilyth:
Ja, I meant the Firefox-specific thing with page-down; the position of the header is consistent for all of the browsers I've tried, and folks said that its placement was intentional. I still think it looks weird, but if others like it, I don't think it's terrible.

irilyth:
One other random note: The header on the thread page is visually pretty similar to the (non-floating, I believe?) header on the board page, but the board page has a "New thread" button in it, while this one has a "Mark thread as read" button in the same place as the "New thread" button. Do we want a "New thread" button on this header? I think it's fine if we don't, it just struck me because when I went to post a new thread, I happened to be on a thread page, and was looking around for the New thread buton because the header looked so much like the header on the board page that has that button. But that may just be a matter of not being used to it.

blackshadowshade:
I don't think it's an issue to not have a "New Thread" button in the header. I think it's helpful to have someone navigate to the forum that they want to post in before creating a thread, so that they can be sure that they're in the right forum.

blackshadowshade:
So, testing on the newly deployed dev site, I'm still seeing the issue with the header not updating its position. Perhaps this is because of my inherent latency, considering I'm hitting from Australia.

So, I have a suggestion or two:
- we could increase the latency
- we could find out if there's a better place in the code for this reflow (but I don't know where this could be)
- we could add a second reflow with a far greater latency, just in case the first one fails

Any thoughts?

mkoryak:
If window.mutationObserver
   Reflow by observing menu children changes
Else
   Settimeout

irilyth:
I don't know how hard it would be, but we can probably spin up a dev site in ap-southeast-2 if that would help...

-Josh (from my phone)

mkoryak:
one more thing. your header issue can be fixed if you init the plugin with `position: 'absolute'` but you will need to rework your css. You have some css that centers the table, but in that mode the table is wrapped in a div. 

But if you do that, you wont need to reflow at all. The header becomes absolutely positioned inside the wrapping div and will move around "with it"

blackshadowshade:
I tried using mutationObserver, but this either didn't catch any changes or it caused the page to hang. So, I dug into the code again and tried to work out when the reflow actually needs to be called. I think I've found the bit of the code where the reflow needs to happen.

@cgolubi1, I'd like to test this on the dev site, to see if it actually works when there's a significant lag on page loading.

cgolubi1:
Updated the dev site from this branch.

blackshadowshade:
Okay, I've taken a look at this floating header and it now behaves far better. I've only been able to trigger the bad behaviour of the overlap 1 out of 20 times, instead of the 1 out of 4 or 5 times before, and that time it broke was when the page took over a second to load.

One thing that I think I'll fix is the nominal width of the header, since it always wraps on my name. I think that my name is probably one of the longest that posts regularly on the forums.

blackshadowshade:
@dwvanstone, would you mind taking a look at this and seeing if the behaviour of the overlapping floating header is fixed from your perspective too?

dwvanstone:
I'm seeing the same behaviour - most times it's fine, and a few times the header is too high, which is when the page takes a long time to load, as if there's some kind of race condition. I think it's rare enough that I'd be happy releasing it.

mkoryak:
Why does the height of your header change? Would it be possible to give it
some minimum starting height to minimize the badness

dwvanstone:
The height of the header doesn't change - the position of the floating header ("Button Men Forums") sometimes overlaps the "Welcome to ButtonMen" header, as shown in the screenshot from May 21. It used to be repeatable and now it's just an occasional bug (on Safari/Mac). I do not see this issue on Firefox/Linux.

mkoryak:
Is this a bug?

```js
  Forum.scrollTo(Forum.scrollTarget);

  $('table.floatable').trigger('reflow');
```

scrollTo animates for 200 ms, so if any scrolling does happen, it will be happening after the reflow. 

On the other hand, it the plugin should handle scrolling. 

I am looking at the code, maybe ill see something

blackshadowshade:
My rationale for putting the reflow where I chose is to get it happening as late as possible. The issue is not that the header is changing size, but rather that the header may actually be created after quite a lag (depending on the server, the connection, and the browser). Thus, if the reflow occurs before the header has been created, it obviously cannot know about the header. Thus, there is effectively a race condition, but it's a complex one that is also browser-dependent.

My guess is that Safari constructs its pages to a different timeline to other browsers, which means that the header appears later than it would on another browser.

I think that the bug (which is only a display bug) now occurs rarely enough that I'm also happy to accept this behaviour. Once I increase the default width of the header, I'll propose that we go ahead with it, with the possibility of tweaking it in the future.

mkoryak:
I didnt see any async calls for creating the header, but maybe I didnt trace through the code far enough. If you know that the header creation can take a while, why not add a callback and hook into that for floatThead? 

If that option requires too much refactoring, you can also just do `$('body').trigger('headerLoaded')` at the end of loading the header and bind floatThead plugin init to that event (just make sure to also load it on tab change)

cgolubi1:
Skimming the history here: i think dwvanstone said it was okay to release as is; irilyth, you also weighed in, okay with you if we release this?  mkoryak or others, let us know if you have any objections.

mkoryak:
I have no objections, I am not part of this team ;)

irilyth:
I still think that having the floating header not flush with the top of the page looks weird, in the particular sense that it looks mis-aligned. I suppose we could ask more other people for more opinions; or just ignore me, go with it as-is, and if we get people on the forums saying "cool, but it's mis-aligned", then we'll know. :^)

Some sites with floating headers that are flush with the top of the page: Gmail, CNN, Facebook, YouTube. Does anyone have any examples offhand of any sites where the floating header *isn't* flush with the top of the page (and which you think look good)?

blackshadowshade:
For information (and for myself in the future, when I'm making changes), the line causing the offset is this one in Forum.js:

    $('table.floatable').floatThead({ top: 8 });

So, if we decide to change the size of the vertical offset, we simply change the value at the end.


Personally, I don't mind either way, with the floating header flush to the top or offset by a bit.

cgolubi1:
shadowshade: want to go ahead and make that change for now, when you have a moment, so we can try it on the dev site?

blackshadowshade:
Okay, I've made that change to remove the vertical offset of the floating header. I've also re-added the 100 ms delay to try to deal with those bad cases that Devious and I were still seeing after really slow page loads.

cgolubi1:
Updated dev site.

blackshadowshade:
I've taken a look at the dev site with Safari, and now, out of 30 page loads, not a single one ends up with the malpositioning that we were seeing before. Maybe 5 of those had a visible refloat just after page load.

dwvanstone:
I no longer see the bug on Safari. (Well done!)

cgolubi1:
irilyth: can you take a look at this dev site when you get a chance and see if the floating header looks better to you now?

irilyth:
Oops, I thought I had checked it recently, and that the dev site hadn't been updated yet, but it obviously has now. I think it looks better this way; anyone else feel strongly about it?

cgolubi1:
Okay.  Shadowshade, what's needed before this can be merged?  I can see the qunit test is a stub; i'm inclined to let that slide for now because i don't have an immediate test syntax to hand you, but let me know if that sounds wrong to you.

blackshadowshade:
Rebase done.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1370/ (if it passes)

cgolubi1:
Alright, let's merge this.

cgolubi1:
Done.  irilyth, you can reclaim `transient` at your leisure.

PR #2214, opened 2017-05-20
Abstract the die rolls for the RandomBM skills used by respondertest

cgolubi1:
* Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/489/ (if it passes)

This is a first step towards addressing a problem i'm having in regression-testing Rush, namely that all the RandomBM* replay tests fail because the random numbers have changed.  I'll still have to do the work of making RandomAI figure out which die rolls are skill selection and abstract those out when storing games for replay, but if we agree on the approach, i can then do that.

Also this change should be helpful to Shadowshade in adding new non-excluded skills, because it'll mean only one change to responderTest is needed to get all the RandomBM* tests caught up with the new values.

### DIFF
--- a/test/src/api/responderTest.php
+++ b/test/src/api/responderTest.php
@@ -7,6 +7,34 @@ function auth_session_exists() {
     return $dummyUserLoggedIn;
 }
 
+// Dieroller values which may change over time
+
+// RandomBM skills (used by all RandomBM types except RandomBMFixed)
+$RANDOMBM_SKILL_ARRAY = array(
+  'B', // Berserk
+  'b', // Boom
+  'c', // Chance
+  'f', // Focus
+  'I', // Insult
+  'k', // Konstant
+  'M', // Maximum
+  'H', // Mighty
+  'n', // Null
+  'o', // Ornery
+  'p', // Poison
+  'q', // Queer
+  'G', // Rage
+  's', // Shadow
+  'z', // Speed
+  'd', // Stealth
+  'g', // Stinger
+  '^', // TimeAndSpace
+  't', // Trip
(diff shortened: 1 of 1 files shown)


blackshadowshade:
Yes! I love this change, and I'll accept it once Jenkins passes.

test/src/api/responderTest.php
```diff
@@ -7,6 +7,33 @@ function auth_session_exists() {
     return $dummyUserLoggedIn;
 }
 
+// Dieroller values which may change over time
+
+// RandomBM skills (used by all RandomBM types except RandomBMFixed)
+$RANDOMBM_SKILL = array(
```
blackshadowshade: One thought here:

What would you think about doing things a little differently?

    $RANDOMBM_SKILL_ARRAY = array(
        'B', // Berserk
        'b', // Boom
        'c', // Chance
        ..., 
        'h' // Weak
    );
    $RANDOMBM_SKILL = array_flip($RANDOMBM_SKILL_ARRAY);

The reason I'm suggesting this is that any time a new skill is added, there is literally one change, instead of all the numbers after the change needing to be updated.

cgolubi1: Sure, no problem - i didn't want to look for array indices inside each test, but doing one flip at the beginning seems totally reasonable.

cgolubi1:
Done and rebased.  Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/490/ (if it passes)

blackshadowshade:
Great, let's go with that then.

PR #2224, opened 2017-06-26
Force JS script files to cache

### LINKED ISSUE #2220: cache busting for js / css assets, jquery.getScript()
so I'm on a plane right now, with cruddy wifi. natch, this isn't the optimal place to play buttonmen in general, but it's also generally speaking not unreasonable to expect buttonmen to minimize network loads.

right now, the system loader we use is jquery.getScript, which looks like it's appending a cachebuster to the end of each url for all resources, as a query parameter derived from `Date.now()`. 

This is problematic in the extreme when (a) my waterfall has a couple of 8sec response times for these js files, and (b) they're the same files from one click to the next (I'm just clicking next game).

you can [set up jquery](https://api.jquery.com/jquery.getscript/) to not append that cachebuster, and do some manual cache-busting on your own, and that would *tremendously* speed up page load times. (the next thing to do is not reload the page each time, but instead just tear it down and build it back up, so I don't have to re-load login, validate my login, reload everything else, etc etc, when getting the results of an attack).

in my dayjob, I use gulp-rev as part of my build process, which generates an md5 hash of the file's content and renames the file (so main.js becomes main.3434aaabbaa033.js). That way, if there's a rebuild that doesn't change a file, the md5 stays the same. I don't think y'all do the same kind of transpile / build workflow, but it could even be manually revved each time you release a new version, since these files rarely change.

blackshadowshade:
Fixes #2220.

This may require people to do a hard reload of various browser pages if we change the underlying JS.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1346/ (if it passes)

### DIFF
--- a/src/ui/buttons.html
+++ b/src/ui/buttons.html
@@ -1,33 +1,41 @@
 <!DOCTYPE html>
 <!--
-    This document should be written in polyglot HTML5/XHTML5, and therefore
-    should run when served either as text/html or application/xhtml+xml.
+  This document should be written in polyglot HTML5/XHTML5, and therefore
+  should run when served either as text/html or application/xhtml+xml.
 -->
 <html xmlns="http://www.w3.org/1999/xhtml" lang="en" xml:lang="en">
 <head>
-    <meta charset="utf-8" />
-    <title>Button Men Online</title>
-    <link rel="stylesheet" href="js/extern/jquery-ui.css" />
-    <link rel="stylesheet" href="gui.css" />
-    <link id="favicon" rel="shortcut icon" type="image/x-icon" href="/favicon.ico" />
-    <meta http-equiv="Cache-control" content="no-cache" />
+  <meta charset="utf-8" />
+  <title>Button Men Online</title>
+  <link rel="stylesheet" href="js/extern/jquery-ui.css" />
+  <link rel="stylesheet" href="gui.css" />
+  <link id="favicon" rel="shortcut icon" type="image/x-icon" href="/favicon.ico" />
+  <meta http-equiv="Cache-control" content="no-cache" />
 </head>
 
 <body>
-    <noscript>
-        Unfortunately, javascript is required to use this site.
-        Please enable javascript and reload the page.
-    </noscript>
-    <script src="js/extern/jquery-1.9.0.min.js"></script>
-    <script>
-        $(document).ready(function() {
-          $.getScript('js/Loader.js', function() {
-            Loader.loadScripts([ 'js/Buttons.js' ], function() {
-                Login.showLoginHeader(Buttons);
-            });
+  <noscript>
+    Unfortunately, javascript is required to use this site.
+    Please enable javascript and reload the page.
+  </noscript>
+  <script src="js/extern/jquery-1.9.0.min.js"></script>
+  <script>
+    $(document).ready(function() {
+      $.ajax({
+        url: 'js/Loader.js',
+        dataType: 'script',
+        cache: true,
+        success: function() {
+          Loader.loadScripts([
+            'js/Buttons.js'
+          ],
+          function() {
+            Login.showLoginHeader(Buttons);
           });
-        });
-    </script>
+        }
+      });
+    });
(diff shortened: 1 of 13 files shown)


src/ui/buttons.html
```diff
@@ -21,11 +21,16 @@
...
-                Login.showLoginHeader(Buttons);
+            $.ajax({
+                url: 'js/Loader.js',
+                dataType: 'script',
+                cache: true,
+                success: function() {
+                    Loader.loadScripts([ 'js/Buttons.js' ], function() {
+                        Login.showLoginHeader(Buttons);
+                    });
+                }
```
cgolubi1: Just a note: all of these change the horizontal whitespace --- our current standard is 2 char indents for JS.

src/ui/prefs.html
```diff
@@ -22,15 +22,21 @@
...
+            $.ajax({
+                url: 'js/Loader.js',
+                dataType: 'script',
+                cache: true,
+                success: function() {
+                    Loader.loadScripts([
+                        'js/extern/jquery-ui.min.js',
+                        'js/extern/spectrum.js',
+                        'js/UserPrefs.js'
+                    ],
```
cgolubi1: Nit: i actually like this style, where each loaded script is on its own line, better than the style in the other files where the loaded script is inline with Loader.loadScripts.  I think it's easier to read which scripts each file is loading at a glance, which is useful.  Mind changing the others to match this?

cgolubi1:
Okay, dev site is up (see https://github.com/buttonmen-dev/buttonmen/wiki/Dev-site-index )

src/ui/buttons.html
blackshadowshade: Ah, interesting. I was trying to work out what the correct indentation was, since the HTML is indented with 4 spaces, but the JS was indented with 2 spaces.

Any thoughts on this? I thought it was odd to have mixed indentation.

src/ui/prefs.html
blackshadowshade: Sure, happy to do this.

blackshadowshade:
Whitespace has been updated. I'll run a Jenkins on this when I rebase it into one commit.

blackshadowshade:
So, I don't know exactly what needs testing here, but I've been able to play a move, view a game that I'm not a part of, view a forum thread and create a new one, view a button page, and create a game. I can log in and log out.

cgolubi1:
I think the main testing someone needs to do is to load the page with firebug or equiv, and verify that when you browse to multiple game pages or whatnot, not as many files are loaded.

blackshadowshade:
That's not too difficult to do by watching the network tab on Firefox Developer Tools. The resources that are cached return a 304 instead of a 200, and I can verify that I'm seeing this. I'd love someone else to take a look though.

cgolubi1:
shadowshade: and you see different ratios of cached to not-cached on other dev sites or on prod?

I'll try to make a few minutes to test this too, or, barring that, if at least one other person can load the dev site and verify that nothing seems broken to them, we can call that good enough.

blackshadowshade:
So, what I'm seeing on the live site is this:

capture

The difference is in the 200s in the middle not being cached any more.

cgolubi1:
Cool.  That seems like it's doing what it's supposed to, then.  Can any other testers load this dev site to make sure they don't see any obvious visual regressions?

dwvanstone:
I've wandered around the site and could see no visual regressions.

cgolubi1:
Thanks, dwvanstone.  That's two ayes and no nays, so i think we are good to go here.

shadowshade: want to rebase this down to one commit?

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1349/ (if it passes)

cgolubi1:
Looks good:

```
$ diff -r --brief TEST-buttonmen-2220_cache_js TEST-buttonmen-2220_cache_js.multicommit | grep -v /.git/
$ 
```

PR #2228, opened 2017-07-03
Added a .gitignore file in the Python API client directory to ignore pyc files

irilyth:
http://jenkins.buttonweavers.com:8080/job/buttonmen-irilyth/45/ (if it passes)


blackshadowshade:
@irilyth, I'm not certain about checking in .gitignore files. I seem to recall a while back that we weren't going to do this, since different people have different files that they ignore, similar to how different people use different editors and editor settings, but we don't check those in.

blackshadowshade:
Oh, the other thing here is that different people have .gitignore files at different levels. I use a global one at work, and a local one at home.

cgolubi1:
Hmm.  I may be misremembering, but if i'm not, what happened before is that someone suggested a .gitignore which ignored .vagrant directories, which breaks my workflow in which i use git diff to figure out whether i have vagrant instances launched in various workspaces.

So my feeling about .gitignore in general is that i don't have a problem with them unless they break someone's workflow.  So i feel like the acceptance criteria should be the regular devs saying "yep, this seems fine with my workflow".

I don't have a strong preference about local vs global.  I feel like global would have the advantage that it lets us encode information about our expectations about what the code tree looks like, all in one place.  So i guess slight preference for global, again, unless it breaks someone's workflow.

Thoughts?

irilyth:
I don't have any opinions at all between global and local .gitignore files.

blackshadowshade:
So, I'm currently looking at my work checked-out copy of the BM archive. I have a .gitignore at the level of the root buttonmen folder, which basically just ignores /src/api/dummy_data. I'm pretty sure that I have a similar one at home.

I've been reading up on multiple .gitignore files, and have learned that .gitignore files apply recursively to all subdirectories (with each set of rules supplementing and supplanting previous rules as necessary). For me, that means that multiple files won't cause bad conflicts, which was one of my concerns. Also, considering that I'll probably never contribute to this Python client directory, I withdraw my objection in this case.

cgolubi1:
Merging based on all the prior discussion.

PR #2230, opened 2017-07-07
Added the ability to show multiple players' or buttons' records, to the 'record' script.

irilyth:
http://jenkins.buttonweavers.com:8080/job/buttonmen-irilyth/49/ (if it passes)

### DIFF
--- a/tools/api-client/python/record
+++ b/tools/api-client/python/record
@@ -2,7 +2,7 @@
 
 ##### record.py
 #
-# Show overall record of a button or a player.
+# Show overall record of one or more buttons or players.
 
 ### Imports
 
@@ -30,22 +30,26 @@ def parse_arguments():
 
   # Set up the main parser and subparser.
 
-  parser = argparse.ArgumentParser(description="Show record of a button or a player.")
+  parser = argparse.ArgumentParser(description="Show overall record of one or more buttons or players.")
 
   subparsers = parser.add_subparsers(dest='type', title="Type of thing to show",
                                      help="more help: record <subcommand> --help")
 
   # Set up the button subparser.
 
-  parser_button = subparsers.add_parser('button', help="show record of a button")
+  parser_button = subparsers.add_parser('button', help="show button record(s)")
 
-  parser_button.add_argument('name', help="button to show")
+  parser_button.add_argument('names',
+                             nargs = '+',
+                             help="button(s) to show ('all' for all buttons)")
 
   # Set up the player subparser.
 
-  parser_player = subparsers.add_parser('player', help="show record of a player")
+  parser_player = subparsers.add_parser('player', help="show player record(s)")
 
-  parser_player.add_argument('name', help="player to show")
+  parser_player.add_argument('names',
+                             nargs = '+',
+                             help="player(s) to show ('all' for all players)")
 
   # Add general optional arguments.
 
@@ -116,6 +120,35 @@ if not bmconn.verify_login():
   print "ERROR: Could not log in to {0}".format(args.site)
   sys.exit(1)
 
-# Print the overall record of the thing.
+# If the name is 'all', do all the things.
 
-print_record (bmconn, args.type, args.name)
+if args.names == ['all']:
+
+  # Prompt for confirmation first.
+
+  print "WARNING: Getting records for all {}s will take a while and slow down the site. Are you sure?".format(args.type)
+  confirm = raw_input("Type 'yes, really' to continue, anything else to abort: ")
+  if confirm != "yes, really":
+    print "You typed \"{}\", aborting now.".format(confirm)
+    sys.exit(5)
+  
+  # If we're still here, get all the things.
+
+  if args.type == 'button':
+    names = bmconn.wrap_load_button_names()
+  elif args.type == 'player':
+    names = bmconn.wrap_load_player_names()
+  else:
+    print "ERROR: I don't know how print all things of this type ({})".format(args.type)
+    sys.exit(1)
+
+# Otherwise, just do the specified things.
+
+else:
+
+  names = args.names
+
+# Print the overall record of the things.
+
+for name in names:
+  print_record (bmconn, args.type, name)


blackshadowshade:
I really don't know if the 'all' feature here is something we should be supporting against a live database, considering how many queries will be running. I'm fine with this sort of script running over a limited number of buttons on a live database, for example, when play-testing a new set. 

Naturally, I have no problems with this happening against a static database, for example, based on a backup.

I'm not going to block this pull request by saying that you shouldn't make this change, because there's nothing to stop you using this code locally without checking it into GitHub. I guess I'm just asking you to reconsider the 'all' feature.

cgolubi1:
wrt shadowshade's concern: what if you put sleep statements in the loop when querying multiple records?  That would provide a way to do this expensive thing that people probably want to do, while putting the expense on wall-clock time rather than slowing down the site.

irilyth:
Sure, a sleep sounds fine. How long?

cgolubi1:
I dunno, if each query takes on the order of 5s, sleeping for a couple of seconds (2-3) between queries is going to slow down the script by less than an order of magnitude, and that way if there is some other query that's getting stuck behind this one, it'll give it a chance.  But maybe that's wrong-headed, as someone said somewhere, we don't think there's a queue of backlogged requests.

blackshadowshade:
If I'm correct, a sleep of 3 seconds should make the querying over all buttons extend to a few hours, so we are in fact increasing things by an order of magnitude.

I'd be okay with Chaos's suggestion of 3 seconds delay, which should improve not only database performance but also Apache/PHP performance.

cgolubi1:
Hmm, i don't think increasing by an order of magnitude is good.

Sorry i haven't yet settled on what i think about this, but: at the end of the day, the problem here is that we are not throttling individual players' ability to make expensive calls.  So as shadowshade noted in his first comment, if people don't query all buttons using this script, they could easily query all buttons not using this script.  If this script parallelized calls, that would be one thing.  But it doesn't, it waits for each call to return before starting the next one.  So it can only be causing one site call at a time.

Given that, and given the plethora of warnings it prints for running in ALL mode, i think we should merge it as is with no sleep.  The warnings mean that well-meaning players will either not slam the site, or at least ask questions.  If we get not-well-meaning players, we'll need to implement server-side throttling anyway.

shadowshade: what do you think of that argument?

cgolubi1:
Oh, i meant to ask:

> I'd be okay with Chaos's suggestion of 3 seconds delay, which should improve not only database performance but also Apache/PHP performance.

Do we actually think this is true, or do we think it's exactly the same impact (sporadic pauses while the server serves each one of these requests, which a given player is either behind or not behind) over a longer time?

blackshadowshade:
Is this something that we can test on a dev site (with admittedly worse specs)?

cgolubi1:
If we think it's important, i guess i can bring up a dev site and load an older copy of the prod database onto it.

That's a bit of overhead, but i don't mind, as long as someone can lay out a test they'd run on that site which they think would tell us something useful about how site performance seems to interactive players while this script is running.

blackshadowshade:
I guess what we're looking to do is test the average page load time for a normal user looking at the history page and loading a normal game, while the script is running as compared to a baseline.

cgolubi1:
How will you test "average"?  How will you tell if it's better or worse if we put sleeps in the script?

blackshadowshade:
I would suggest running 10 page loads of the history page showing 'all history for irilyth', and 10 page loads of an active game. Timing could be done with a stopwatch, since we don't need a huge amount of precision. What we're looking for here is a gross difference in load times (e.g., 20% increase), something noticeable.

irilyth:
I think we can just test this in prod if we want. I did that one time; if our subjective user experience is terrible when we go to test while it's running, we can just instantly kill it.

I won't be around late tonight, but could help with this at some point in the 20:00 Eastern timeframe, or later on Thursday.

blackshadowshade:
Okay, sounds good. I brought up the idea of testing on a dev instance so that we could have a stable baseline on page loads, but it should be fine if you can get a stable baseline.

cgolubi1:
irilyth: any reason you can't run this test yourself and report on the results?  I can try to get to it this weekend if no one else can, but i think basically anyone has the necessary tools.

irilyth:
> I would suggest running 10 page loads of the history page showing 'all history for irilyth', and 10 page loads of an active game. Timing could be done with a stopwatch, since we don't need a huge amount of precision. What we're looking for here is a gross difference in load times (e.g., 20% increase), something noticeable.

Ok, sure.

For the game page, I used http://www.buttonweavers.com/ui/game.html?game=25256, which has a long chat log, and for the history page, http://www.buttonweavers.com/ui/history.html#!playerNameA=irilyth. For each, I did a shift-reload each time, to avoid the cache as much as possible. I did ten tests before running the script (./record button all), the game page first and then the history page, then ran the script and did ten tests again, then killed the script and did ten tests again. I reloaded the page ten times in a row, with a lap timer on a stopwatch app on my phone, hitting "lap" on the timer and "shift-ctrl-r" on my keyboard at approximately the same time.

Game page results:

before | during | after
------ | ------ | -----
3.36 | 4.99 | 2.77
2.61 | 4.30 | 3.57
3.13 | 4.00 | 3.33
3.28 | 4.36 | 3.26
2.95 | 3.71 | 3.39
3.40 | 3.97 | 3.63
2.86 | 3.58 | 3.21
3.16 | 3.12 | 3.03
2.69 | 3.80 | 2.81
2.87 | 3.15 | 2.99

Totals: 30.31 before, 38.98 during, 31.99 after.

History page results:

before | during | after
------ | ------ | -----
7.41  | 10.81 | 7.31
6.93  | 10.76 | 6.82
6.29  | 12.96 | 7.29
6.21  | 8.62  | 7.47
5.88  | 11.21 | 6.97
7.00  | 10.78 | 6.62
5.67  | 11.96 | 8.14
6.53  | 10.33 | 6.50
6.20  | 11.24 | 7.45
12.68 | 9.82  | 6.45

Totals: 70.80 before, 108.49 during, 71.02 after.

This was all between around 12:30 and 01:00 Eastern.

blackshadowshade:
So, there's about a 25% increase in page load times while the script is running. That's pretty significant.

Could you try running the same sort of test with a 2 sec sleep between each query?

cgolubi1:
Neat (ish).  Thanks for assembling that, and my additional request is also for the stat shadowshade asked for (what difference does it make to the impact if there's a 2-second sleep between queries).  Remind me: do we know approximately how long the script takes to run (without sleeps)?  It sounds like you killed it rather than letting it run to completion during your test yesterday.

irilyth:
Where did I talk about run time... Ah, it was on IRC, and it took about 15 minutes to run the last time I tried it. I killed it this time when I was done testing.

I can do another round with sleeps, of various lengths, but it takes like half an hour to do each round of testing, is tedious and physically fiddly, and I don't think it'll tell us anything useful.

The situation here isn't that the site is suddenly getting hundreds of simultaneous queries, such that spreading them out with a sleep between them would improve performance. The situation is that the site is getting one request at a time, one after the other, for 15 minutes, about 600 of them. If we put in a 2-second sleep, then it'll be getting one request at a time, one after the other, for about 35 minutes. During that time, if someone loads a page while the site's processing one of thse queries, it'll be 25% slower; if they load a page while it's sleeping, it won't be.

It seems better to me to have the site be 25% slower for 100% of the page hits for 15 minutes, than for it to be 25% slower for 40% of the page hits for 35 minutes.

But, I can do it, just in case I'm wrong, and somehow the series of queries is causing more load when run back to back than when there's a sleep between them.

If anyone has any suggestions about how to automate the tests, I can do it sooner. If not, I can do it by hand when I have an hour or two to spend on it, probably in a week or two.

cgolubi1:
shadowshade: can you weigh in on that?  For my part, irilyth's explanation of the impact seems correct to me.  Do you think it's likely that the impact would be something different, so that it's worth running the experiment?

As i said a long time ago, i think we should accept this pull.  The tool has a lot of warnings around it, and well-meaning people will respect those warnings, and if we get non-well-meaning people, we are going to need to deal with the actual underlying performance problem, whether or not we accept this pull.

irilyth:
Oh, the other thing I was going to mention is that with only ten samples, there's likely to be a lot of variance: It wouldn't be at all unlikely for seven of the tests to hit during a sleep, or during a query, and thus to get very different overall results.

As a longer-term idea, it might be interesting to put together some fully automated performance tests / metric-gathering / something, but that's probably a separate project.

blackshadowshade:
I guess the thing that doesn't make sense to me is the "about 600 queries" bit. I would have thought that a comprehensive matchup query would require far more queries. If the number of queries is true, then each query is taking in the order of 1.5 seconds, which means that a 2 second pause should improve performance significantly for other requests.

Why? Well, we know that page loads are in the order of 2 - 7 seconds normally, and there are multiple simultaneous requests required for each load (excluding the improvements that we'll get by caching). It seem likely that a number of these requests will complete more quickly during the 2 second pause, since the requests will occur over a period of more than 2 seconds.

cgolubi1:
shadowshade: it's iterating over either players or buttons.  So 600 queries sounds about right to me.

And the point about the delay is, the number of queries is the number of queries.  So if they take place over a longer period of time, they'll slow down a smaller fraction of a larger number of requests.

I think what i'm waiting for here is either:
* shadowshade to agree that we can merge the pull as is
* irilyth to agree to run an experiment with sleeps

and as long as neither of those things happens, i'm going to call this PR indefinitely blocked.

Note that if irilyth *does* run the experiment and it looks like the sleeps don't help, i'm going to take that as a reason to merge without the sleeps.

blackshadowshade:
I guess I'm uncomfortable about the large impact this script has on load times, especially if it could potentially be run by more than one person simultaneously---and that's the whole point of adding the python client to the Github repository, so that it is theoretically accessible and usable by anyone who wants to use it.

I would ideally like the impact on load times to be less than 20%, since this is roughly within the range of the normal baseline fluctuations seen in the testing, and thus shouldn't impact on the subjective sluggishness of the site to a casual user.

cgolubi1:
I mean, at the end of the day, the problem isn't the client, the problem is the server.  As i've said a couple of times before, i'd rather players have access to a script that warns them via a prompt that they're going to have an impact on the site, so they might choose not to run it often, than that players just stumble over the API and write their own loops without thinking about it.

But i've said that a couple of times and haven't gotten a response to it, so i assume it's not a compelling argument.  Like i said, calling this one blocked; if there's anything specific either of y'all would like me to do to try to help unblock it, let me know.

irilyth:
There's almost no chance that anyone other than me will ever run this, so I'll just run it out of my clone, and we can stop wasting time on it.

blackshadowshade:
I'm actively looking at providing the button vs button matchup stats in a different way, so in the medium term, I'm hoping that the 'all' feature shouldn't ever need to be run.

PR #2237, opened 2017-07-25
Make randomAI able to do replay testing now that we have isChatPrivate in the API response

cgolubi1:
Without this change, `replay_loop` blows up on `prep_replay_games`.  I should do replay-testing on my own pulls which change the API; that was my bad.
* Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/496/ (if it passes)


blackshadowshade:
Okay, looks good.

PR #2210, opened 2017-05-16
Added Rush skill, take 3

blackshadowshade:
Fixes #2176.

This pull request implements the version of Rush most recently proposed by James Ernest.

This requires a responder test to lock in the behaviour of Rush dice.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1333/

PR #2243, opened 2017-08-05
Modify bm_rand() handling to use a separate function for skill-selection random numbers

cgolubi1:
* Unblock regression testing for #2210 
* Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/500/ (if it passes)

### DIFF
--- a/deploy/amp/phpunit_bootstrap.php
+++ b/deploy/amp/phpunit_bootstrap.php
@@ -5,6 +5,7 @@
 
 // Setup bm_rand() override for unrandom tests
 $BM_RAND_VALS = array();
+$BM_SKILL_RAND_VALS = array();
 $BM_RAND_REQUIRE_OVERRIDE = FALSE;
 
 function bm_rand($min = FALSE, $max = FALSE) {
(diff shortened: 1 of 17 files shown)


blackshadowshade:
It seems to me that there might be missing function parameters to bm_skill_rand()?

cgolubi1:
Huh, you're right.  It doesn't end up mattering, because we never call the test functions without `$BM_RAND_REQUIRE_OVERRIDE` so they always exit before that point.  But when you're right, you're right.  :>)  Will fix.

cgolubi1:
I made that fix, but i think this isn't passing jenkins, so i'll keep working on that, and don't worry about it until then.

cgolubi1:
Okay, sorry about all that - i've fixed the unit tests so Jenkins now passes - http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/502/ - and since no one has stood up a dev site for this, i went ahead and rebased back down to one commit.  (The only change from the previous commit was the fixing of the `BMBtn*` unit tests to use the new die-roller-override syntax.)  This is ready to review.

blackshadowshade:
Okay, this looks fine to me. After looking specifically at the changes to BMBtnSkillRandomBM.php and the bootstrap files, I'm pretty sure that the changes should leave live functionality untouched, and that's the most important thing.

Issue #1784, opened 2015-11-04
Allow third-party game creation


irilyth:
Now that we have a way to accept or reject games, allow players to create games between two other players, which the two other players would both be prompted to accept.

(My perhaps-obvious use case for this is Mad Robin and UBFC; this would let my scripts create matches via the API, with all the right parameters, rather than having to use forum posts to ask players to create their own, and sometimes getting the parameters wrong.)

danlangford:
another use case would be introductions. my friend starts playing and maybe there are a few players that have been particularly helpful and kind to me i could introduce my friend to them by starting a game between them. i think its worth discussing if this should be an opt-in feature via the preferences. "Allow third parties to create games with you vs other players"

irilyth:
#1783 might cover the "allow third parties etc" preferences.

Hurtay:
As long as we don't have tournaments implememted (which probably would be a lot of programming to do and therefore won't come anytime soon) I think this third party games are a brilliant idea. Imho this issue is something comletely different than #1783. 
Since everyone can reject games he doesn't want I'm not even sure if it's necessary to put it into the player's preferences but for being on the safe side a box to tick (or better: untick) is the perfect solution.

Hurtay:
This is my BfotM for February 2016 (sorry for labeling as late as I do, it's my first label at all and I wasn't sure how to do it).

blackshadowshade:
Just an update on this: I'll take a look at this once the action logging has been translated across to the new methodology, since we'll most probably want to log who created the game.

AdmiralJota:
Ooh. It'd be nice to log that for regular games, too. I'd like to see log entries for "X created an open game", "Y joined the game", "X challenged Y", etc. Should we make another ticket for that?

blackshadowshade:
Yeah, let's make that a new issue. I'll be forced to address it when we do this issue, but I might address parts of it beforehand too.

irilyth:
In the course of testing this, I found two bugs with the current game-creation API, #2218 and #2219; meanwhile, this change introduces a new bug, whereby I can create a game that specifies only the player's button, but not the player, opponent, or opponent's button. This is a regression from master, which currently says "ValueError: Failed to call createGame, got: Game create failed because you must be the first player."

I think that until/unless we want to support open games with neither player specified, the API should require you to set 'player' at a minimum -- it doesn't have to be yourself any more, but it has to be *someone*.

blackshadowshade:
[...]

PR #2244, opened 2017-08-14
Added third-party creator to game page

blackshadowshade:
Addresses part of #1784.

Currently, what I've done here is add the creator info to the response sent from the backend, then if the creator is not one of the two players in the game, I've shown the creator on the game screen. There's no change to the Overview screen yet.

If we want any changes to the Overview screen, I'd suggest that we do that in a separate pull request.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1357/ (if it passes)

### DIFF
--- a/src/engine/BMGame.php
+++ b/src/engine/BMGame.php
@@ -3598,6 +3598,7 @@ class BMGame {
             'validAttackTypeArray'       => $this->get_validAttackTypeArray(),
             'gameSkillsInfo'             => $this->get_gameSkillsInfo(),
             'playerDataArray'            => $this->get_playerDataArray($requestingPlayerIdx),
+            'creatorDataArray'           => array('creatorId' => $this->creatorId),
         );
         return $dataArray;
     }
--- a/src/engine/BMInterface.php
+++ b/src/engine/BMInterface.php
@@ -199,6 +199,9 @@ class BMInterface {
             $data['playerDataArray'][0]['playerColor'] = $gameColors['playerA'];
             $data['playerDataArray'][1]['playerColor'] = $gameColors['playerB'];
 
+            $data['creatorDataArray']['creatorName'] =
+                $this->get_player_name_from_id($data['creatorDataArray']['creatorId']);
+
             return $data;
         }
         return NULL;
--- a/src/ui/js/Api.js
+++ b/src/ui/js/Api.js
@@ -514,6 +514,9 @@ var Api = (function () {
                          data.playerDataArray[my.game.opponentIdx],
                          my.game.opponentIdx);
 
+    my.game.creatorId = data.creatorDataArray.creatorId;
+    my.game.creatorName = data.creatorDataArray.creatorName;
+
     my.game.pendingGameCount = data.pendingGameCount;
 
     return true;
--- a/src/ui/js/Game.js
+++ b/src/ui/js/Game.js
@@ -1748,6 +1748,14 @@ Game.pageAddGameHeader = function(action_desc) {
   } else {
     gameTitle += 'Round #' + Api.game.roundNumber;
   }
+
+  // add creator if the game has been created by a third party
+  if (('CHOOSE_JOIN_GAME' == Api.game.gameState) &&
+      (Api.game.creatorId != Api.game.player.playerId) &&
+      (Api.game.creatorId != Api.game.opponent.playerId)) {
+    gameTitle += Game.SPACE_BULLET + 'Created by ' + Api.game.creatorName;
+  }
+
   $('title').html(gameTitle + ' &mdash; Button Men Online');
 
   Game.page.append(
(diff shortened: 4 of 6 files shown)


cgolubi1:
Jenkins failed - looks like you need to add your API changes to the python API too.

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1358/ (if it passes)

cgolubi1:
Jenkins passes now.  Let's have a dev site.

cgolubi1:
Dev site is at 54.221.87.113 - irilyth, want to allocate a CNAME?

blackshadowshade:
@irilyth, when you get a chance, would you mind creating some third-party games here, please? We'll want a variety of player combinations, so that we can look at both games where we're involved and games where we're not involved.

Also, try creating a game where the game creator is the second player, and another player is the first player.

irilyth:
Since the ticket didn't seem to link to it yet: This is being tested at http://blip.dev.buttonweavers.com/

One immediate thing, that probably isn't related to this change: When I tried to create a game against a player who doesn't exist, it succeeded, but now http://blip.dev.buttonweavers.com/ui/game.html?game=50 says "The game hasn't started yet." ... Ha, ok, it created it as an open game. That seems wrong, but probably isn't related to this change.

I ran it again, and it created some games, 51 - 56; and when I look at the ones that iriclone is in and irilyth isn't, I see the "Created by irilyth" bit at the top (and not in the ones that irilyth is in).

Do we actually want this at the top of every game page? It's probably fine, just wasn't sure what the expected/intended effect was.

blackshadowshade:
I think it's important to have it at the top of every game page, because we want to be able to know at any point in time that a game has been created by someone else, especially when I get automated tournaments up and running.

cgolubi1:
IMO the ideal (not talking about "what would i veto or not" just saying if i could pick anything at all) would be to have the game creation including the creator be the first thing in the activity log.

cgolubi1:
So, as a playtester, what i dislike about this (and why i think this is kind of worse-than-nothing as-is), is that i don't like the creator having the same prominence as every other piece of information about the game, during all of gameplay.  I don't care about the creator while i'm playing the game, and because of where the creator name sits, i think it's going to be easy for people to mistake the creator for the opponent and get confused while playing.

Here's a thought: why not include that extra segment in the header, or somewhere else prominent, *only during the phase where you're deciding whether or not to accept the game*?  Because IMO when you're deciding whether to accept the game is when you really might care where it came from.  Then after the game is accepted, drop that segment from the header, and at that point people can rely on a solution like #2252 if they need to look up the information.

blackshadowshade:
I've updated this to only apply when players are deciding whether to join a game or not.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1371/ (if it passes)

cgolubi1:
That addresses my objection, and i'm happy to see this merged.  irilyth, you commented on this PR before.  What do you think of the current state on `blip`?

cgolubi1:
Oh, i should have said: i updated the dev site.  :>)

irilyth:
Seems good to me! I think it doesn't stand out all that much, but if blackshadowshade is happy, it's fine by me.

blackshadowshade:
I think this compromise makes it possible to easily determine whether a newly created game to be accepted has been created by a third party, and thus, we can potentially defuse the potential for malicious game creation. I'd be open to making the creator more prominent, for example, bolding it, but I think the current solution may be enough.

Personally, I have the problem that I have auto-accept games set, so I won't be seeing the message at all, since I skip this game phase.  XD

I'd also be open to us having some sort of visual clue on the overview page for third-party created games, which could double as a sign of automatically-generated tournament games when we have that implemented.

cgolubi1:
We're having this conversation on a bunch of different issues right now, so let me summarize where i think we are:
* This pull (#2244) adds creator info to the game header during the game acceptance phase - i think no one objects to that
* Pull #2252 adds a creator action log entry to new games - i think no one objects to it either
* People would like to see the creator action log entry on older games, so we'd need to figure out how to backfit that.  I think we currently think that can be a separate pull from #2252, and that it's not a blocker for accepting #2252.  Let me know if anyone disagrees
* There's discussion about whether we should do more, e.g. shadowshade just asked if games created by third parties should be marked on the overview somehow.  I think we don't have general consensus around any concrete actions for that, and will need to discuss more.

cgolubi1:
Shadowshade, meanwhile i think you can rebase this down to one commit, because i think it is going to be accepted as-is.

blackshadowshade:
Rebased.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1374/ (if it passes)

cgolubi1:
Okay, this looks unchanged in the rebase, so i'm going to merge it.

PR #2245, opened 2017-08-18
Initial check-in of stats script

blackshadowshade:
Addresses part of #614.

So, I thought that it might be a good idea for me to submit the current version of my stats script to the repository, firstly so it doesn't get lost, and secondly so that we can start moving forward on getting stats working on the live site.

I don't anticipate any problems with Jenkins since I'm only submitting one file to the tools directory, but here it is anyway:

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1360/ (if it passes)


cgolubi1:
Fine with me.  You'll eventually want to put a README in this directory so people know what it is, but i don't think that's worth holding up getting it cached somewhere.

PR #2249, opened 2017-08-29
Fix replay_loop regression

cgolubi1:
* Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/504/ (if it passes, which it should, since jenkins doesn't use replay_loop)


blackshadowshade:
Looks okay, Python only, Jenkins passes. Merging.

PR #2252, opened 2017-08-31
Add an action_log entry at game creation time

cgolubi1:
* Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/507/ (if it passes)
* Relevant to #1784

### DIFF
--- a/deploy/database/schema.game.sql
+++ b/deploy/database/schema.game.sql
@@ -76,6 +76,13 @@ CREATE TABLE game_action_log (
     INDEX (game_id)
 );
 
+CREATE TABLE game_action_log_type_create_game (
+    id                 INTEGER UNSIGNED AUTO_INCREMENT PRIMARY KEY,
+    action_log_id      INTEGER UNSIGNED NOT NULL,
+    creator_id         SMALLINT UNSIGNED NOT NULL,
+    INDEX (action_log_id)
+);
+
 CREATE TABLE game_action_log_type_end_draw (
     id                 INTEGER UNSIGNED AUTO_INCREMENT PRIMARY KEY,
     action_log_id      INTEGER UNSIGNED NOT NULL,
--- a/deploy/database/updates/01784_game_creation_action_log.sql
+++ b/deploy/database/updates/01784_game_creation_action_log.sql
@@ -1,0 +1,6 @@
+CREATE TABLE game_action_log_type_create_game (
+    id                 INTEGER UNSIGNED AUTO_INCREMENT PRIMARY KEY,
+    action_log_id      INTEGER UNSIGNED NOT NULL,
+    creator_id         SMALLINT UNSIGNED NOT NULL,
+    INDEX (action_log_id)
+);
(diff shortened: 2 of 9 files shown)


cgolubi1:
I'm submitting this for two reasons.  One is that it's something i've been feeling the lack of as a player for a long time --- it's often nice to know when a game was created and which player created it.  The other is that it's the alternative i suggested in my comment to #2244 - it's a way to get the information about who created a game available in the page in a consistent way, without making it prominent and confusing players.

Note that i'd like this to be considered for merging regardless of what we decide about #2244 because i think it's independently useful.  But i also think it might make #2244 unnecessary.

cgolubi1:
Shadowshade - take a look at the code, and if it looks generally okay, i can make a dev site.

blackshadowshade:
Just flicking through the code, I note that there is no attempt to go back and update existing games to add a log entry for them. We can certainly add that in a separate pull request. What do you think about the idea?

(I'll take a closer look at the code soon.)

cgolubi1:
Mm.  I'd be inclined not to bother, because we don't know what time older games were created (do we?), and we don't necessarily know who created them.  Or did you backfill the creator field when you added it to the DB?  (I don't think i checked.)

I don't object if we can think of a good way to get the information, but i do think it can be a separate pull, since this code definitely won't cause a problem if backfilled action-log entries don't exist.

blackshadowshade:
I've checked the live database, and we have creator_id and start_time filled for the first thirty games, so I presume that we have this information for all games.

blackshadowshade:
Okay, this looks good to me. Let's get a dev site up.

I've tested a little locally and the chat log seems to work both for standard games and third-party-created games. However, the most important thing for me, namely, seeing whether a third-party has created a game for me, doesn't stand out, and so I'd argue that something like #2244 is still necessary.

cgolubi1:
Dev site created - irilyth, at your leisure can you hook up `transient=54.92.148.192`?

irilyth:
> Dev site created - irilyth, at your leisure can you hook up `transient=54.92.148.192`?

Done!

irilyth:
> I've checked the live database, and we have creator_id and start_time filled for the first thirty games, so I presume that we have this information for all games.

It doesn't seem to be showing for all games, like http://transient.dev.buttonweavers.com/ui/game.html?game=32 -- is that surprising?

(It does seem to work for new games, like http://transient.dev.buttonweavers.com/ui/game.html?game=50 .)

blackshadowshade:
Yeah, that's expected. We'll need to run some SQL to retrofit this solution to already-existing games.

cgolubi1:
The outstanding question here is whether it's fine to push this without the retrofit to older games, or whether we need that as part of this pull.  If folks say it's fine to have it just for new games, shadowshade, you can go ahead and merge it - it's already one commit.

irilyth:
Since I mentioned this in passing: I think this is fine as-is. I only mentioned older games because there was some talk about it, and I wasn't sure what was expected -- I actually don't much care if older games get updated or not.

blackshadowshade:
Okay, merging this. Chaos, I figure it's probably an easy job to write SQL for a retrofit, so we should probably do this for completeness.

PR #2256, opened 2017-09-19
Simple symlink autoversioning for BM JS and CSS files

### LINKED ISSUE #772: Force browsers to update cached Javascript files on redeploy
Currently, if Javascript files are changed on the site, browsers need to be refreshed to force them to update their cached versions.

Based on the discussion in #763, it may be possible to use a redirection trick to force browsers to reload the Javascript files.

### LINKED ISSUE #2255: no known way to force a JS cache refresh in firefox
After the cache change merged in #2224 i am not aware of any way to get firefox tabs to refresh the UI when the javascript changes.  When i change the code and do a reload or a shift-reload, nothing changes.  Closing the tab and opening a new tab, again, no impact (i still see the old JS).  I haven't yet restarted firefox because that's kind of a timesink for me, though i'll try to check the next time i restart it for some other reason.  This is on firefox 55.0.3 on a mac.

A standard page reload does the right thing in Chrome.

This is a release blocker unless we can demonstrate somehow that no one other than me sees this - we need to be able to give people a method for updating the JS on the site when we do a release, and it seems that our standing recommendation of shift-reload no longer works under firefox.

cgolubi1:
* Fixes #2255 
* Fixes #772
* Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/509/

### DIFF
--- a/deploy/vagrant/manifests/init.pp
+++ b/deploy/vagrant/manifests/init.pp
@@ -27,6 +27,7 @@ node default {
       $puppet_hostname = "sandbox.buttonweavers.com"
     }
   }
+  $puppet_timestamp = generate('/bin/date', '+%s')
 
   # Generic node configuration
   include "apt::client"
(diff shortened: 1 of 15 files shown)


blackshadowshade:
Okay, this looks good to me. Let's give this a go!

PR #2261, opened 2017-09-24
Pass the version cookie to Loader.js and use it when loading default BM modules

### LINKED ISSUE #2260: Game showing up as "Created by undefined" in game accept UI
I created http://staging.buttonweavers.com/ui/game.html?game=107 by clicking "Change first player", selecting irilyth, selecting blackshadowshade as the second player, and choosing Random button for both players.

The action log correctly knows that i created the game.  The header bar shows:
```
Game #107  •  irilyth (Magistra) vs. blackshadowshade (bluebec)  •  New Game  •  Created by undefined
```

Diagnose and fix.

cgolubi1:
* Fixes #2260
* Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/514/

I didn't do significant testing that this fixes the particular problem, because that's hard to setup.  But i tested that the pages still seem to load, which is the important part.  Given that it doesn't break page loads, it *should* fix the problem (given that the version cookie business helped for other JS files), and the implementation is such that it shouldn't have any impact on your plan for keeping things working on your windows stack - the sed stuff didn't change at all, so your existing set of symlinks should still be just fine.

### DIFF
--- a/src/ui/active_players.html
+++ b/src/ui/active_players.html
@@ -30,7 +30,8 @@
           ],
           function() {
             Login.showLoginHeader(ActivePlayers);
-          });
+          },
+          'VERSION_COOKIE');
         }
       });
     });
--- a/src/ui/buttons.html
+++ b/src/ui/buttons.html
@@ -31,7 +31,8 @@
           ],
           function() {
             Login.showLoginHeader(Buttons);
-          });
+          },
+          'VERSION_COOKIE');
         }
       });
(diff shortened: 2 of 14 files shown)


blackshadowshade:
Okay, let's give this a go. It's a different date to the last deploy, so when you redeploy staging, we should now see everything working off the newly versioned JS files.

PR #2264, opened 2017-09-26
Fix bug in (hackish) handling of trip turbo range in RandomAI

cgolubi1:
* Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/515/ (if it passes)

I went to check in on my long-running replay site and found it had stopped its loop due to this bug, which it encountered in CestWest vs Wyoming.  The problem was immediately reproducible.  With this fix, it successfully played over 20 CestWest vs Wyoming games without blocking.


blackshadowshade:
Okay, looks fine to me. Python only, Jenkins passes.

PR #2265, opened 2017-09-29
creatorDataArray is a key that will not change during replay

cgolubi1:
* Jenkins: http://jenkins.buttonweavers.com:8080/job/buttonmen-cgolubi1/516/ (if it passes)

Shadowshade: sorry for the inconvenience.  It would be fair to ask how i can possibly accept so many pulls that break replay testing without noticing.  I think i'm just not good at remembering that API changes that obviously aren't game logic regressions, nevertheless impact replay testing.  Anyway, this is all in service of getting to the point where i can do replay testing for things in the queue that actually might impact game logic, so help me keep churning through these, and thanks in advance.

### DIFF
--- a/tools/api-client/python/lib/random_ai.py
+++ b/tools/api-client/python/lib/random_ai.py
@@ -37,6 +37,7 @@ STRING_KEYS = [
 ]
 
 UNUSED_DURING_AUTOPLAY_KEYS = [
+  'creatorDataArray',
   'gameChatEditable',
   'gameChatLog',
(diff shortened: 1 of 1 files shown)


cgolubi1:
This passes and only touches RandomAI, so i'm going to go ahead and merge it - shadowshade, if you see any problem with that at all when you're next online, let me know and i'll revert.

blackshadowshade:
Looks fine to me.

PR #2268, opened 2017-10-11
Fixed bug with fire decline turndown and turbo

### LINKED ISSUE #2258: Game ends up in a bad game state during fire overshooting
Original post: http://www.buttonweavers.com/ui/forum.html#!threadId=569&postId=10615

> In game 28247 I attempted to overshoot with my poison fire die so that it would be low enough to be captured. I did something wrong and got an error message, and instead of carefully reading the message to figure out exactly what I needed to do, I attempted to do a quick fix and re-submitted my move. I did this a couple times before noticing that I needed to choose between a couple action choices at the bottom of the screen. I tried to choose one of the actions, but I guess it was too late and I got the following error message:

> "Can't figure out what action to take next"

blackshadowshade:
Fixes #2258.

There's no easy fix to the database to get these games running again with the correct logging.

I propose that we set the status of these games to BROKEN (status = 5), which will hide them from the Overview screen.

If that is considered to be too drastic, we could set the game_state in the 'game' table manually back to 40 (START_TURN), and toggle the is_awaiting_action values in the 'game_player_map' table for both players. That will get the games going, but the logging will be incorrect for the one broken attack.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1378/

### DIFF
--- a/src/engine/BMInterfaceGame.php
+++ b/src/engine/BMInterfaceGame.php
@@ -1320,7 +1320,10 @@ class BMInterfaceGame extends BMInterface {
                         $argArray['fireValueArray'][$dieIdx] = $dieValueArray[$tempIdx];
                     }
                     break;
-                case 'no_turndown':  // fallthrough to allow multiple cases with the same logic
+                case 'no_turndown':
+                    $argArray['dieIdxArray'] = $dieIdxArray;
+                    $argArray['dieValueArray'] = $dieValueArray;
+                    break;
                 case 'cancel':
                     $argArray['dieIdxArray'] = $dieIdxArray;
                     $argArray['dieValueArray'] = $dieValueArray;
@@ -1356,7 +1359,6 @@ class BMInterfaceGame extends BMInterface {
                 'Caught exception in BMInterface::adjust_fire: ' .
                 $e->getMessage()
             );
-            var_dump($e->getMessage());
             $this->set_message('Internal error while adjusting fire dice');
             return FALSE;
         }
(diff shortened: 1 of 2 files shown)


cgolubi1:
This looks straightforward and you added a test.  I'm just going to replay a few games, then i'll merge this.  Sorry it got lost in my inbox.

cgolubi1:
My primary complaint looking at this (and, yes, i am working on replaying - sorry, i basically always find a problem with my test rig's state whenever i go back to it.  Obviously i need monitoring of some sort.  `:>P`) is that i don't see the bug described in words anywhere.  I think a comment at the top of the new responderTest listing the set of moves in the test which are important to reproduce the former-bug, and what the desired behavior is, would be ideal.

I'll merge without it if need be, but it would be better for future development if we had something written down that said what the bug had been.

blackshadowshade:
So, let me describe the bug here so that you know what you're replaying. We can add that to the responder test too:

- the player needs to have both a fire die, and a turbo die with value not at max, but with a value at least as large as an opponent's targetable die
- the player needs to have fire overshooting turned on
- the player needs to perform a power attack with the turbo die targeting the target die
- the player needs to decline fire overshooting

Then the bug should hit, which is caused by the turbo die resize being deleted as soon as the player declines fire overshooting.

cgolubi1:
Okay, gotcha.  Given that, i think your test is very straightforward.  I've now replayed on the order of 1000 games with no errors, and i think that's good enough.  Merging.

cgolubi1:
BTW, don't delete your branch.  Because i'm going to do a pull of just this branch into staging, to try to do a point release of this fix.

cgolubi1:
Hmm, well, i said that, but in fact there's not that much that's scary in master right now, so let's not do anything weird --- let's just kick off a normal release.

PR #2271, opened 2017-10-19
Changed colour of can't win message

### LINKED ISSUE #2147: Please change the surrender game text to a color other than red
Can the surrender game text be a color other than red?  Plain-old black?  I feel like the game is yelling at me, not suggesting at me.

blackshadowshade:
Fixes #2147.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1382/

### DIFF
--- a/src/ui/gui.css
+++ b/src/ui/gui.css
@@ -104,7 +104,7 @@ div.battle_mat_opponent_overlay {
 }
 
 div.cantWin {
-  color: red;
+  color: blue;
   font-weight: bold;
   margin-bottom: 10px;
 }


cgolubi1:
Sorry this got lost in my inbox.  This needs to be a dev site, or at least, people need to weigh in on whether they think the new color is better.  Standing one up now.

cgolubi1:
irilyth, can you sync up 54.198.218.1 as `temporary`?

And folks in general, can we get some quick testing on this dev site --- all shadowshade did was change the color of "You can't win", so is it better or worse?

irilyth:
Done!

What's an easy way to test this? Any suggestions for a matchup that's easy to manipulate into a you-can't-win state?

cgolubi1:
Hmm, how about Bar Mitzvah Boy vs Lab Rat or maybe better any of the Soldiers without 20-siders, like Avis, Changeling, etc?

blackshadowshade:
So, testers, what do you think about the new colour?

dwvanstone:
It's been too hard to create a game to get to that state. Is there a user/login with a game I can use to view the new colour? Or post a screenshot?

dwvanstone:
Or do people tend to test with two users so they can play out a game?

irilyth:
I generally test stuff like this with two users, so I can quickly get games into the desired state.

Here's a screenshot:

bm

I'd suggested orange, and I think that would look better, but I think the blue is ok if folks prefer that.

AdmiralJota:
I suppose this is probably already a lost battle at this point, but I still prefer the original red. Partly because it's traditional (it's known as "the red text", after all), and partly because it's strongly attention-getting: it lets you know there's something you really ought to be paying attention to here.

blackshadowshade:
No, I don't think it's a lost battle. I think that the colour is certainly not set yet.

I chose the blue because I didn't really want to introduce a new colour that we don't use elsewhere. However, am I'm hearing correctly that both irilyth and Jota don't think this blue stands out enough?

irilyth:
I don't have particularly strong feelings about it, and I think the blue stands out fine; I mostly suggested orange because it says "warning" to me, while blue looks more like "link".

cgolubi1:
I don't think there's a tearing rush on this.  Let's not merge it.  Instead, let's give orange a shot --- maybe we'll get consensus on orange, and if we don't, we can figure out what to do next.

AdmiralJota:
I wonder if there's a good shade between orange and red that A) looks different enough from the usual error message red that it won't be confused, but B) still conveys the feel of a red alert-y kind of message?

Maybe something around #E02810?

irilyth:
The original request on #2147 specifically didn't want red because it felt "like the game is yelling at me", so something that feels like an alert may not be the intended effect.

AdmiralJota:
I suppose we need to answer that question, then: *do* we want it to feel like an alert or not? I would vote for yes.

irilyth:
Might it make sense to discuss this more on #2147, so the original requester will see it?

nkmcalli:
Why does the surrender suggestion message warrant attention getting?  It's not that big a deal.  For example, I never surrender no matter what the message says.  You can also just make the text black.

Red should really be reserved for serious situations where data loss can occur or big mistakes can be made.  If at all.  And remember, in other cultures, red means something different anyway.  For example, red can mean good luck.

nkmcalli:
I like dwvanstones comments on the other thread, and agree with the black text suggestion there.

blackshadowshade:
I'm withdrawing this pull request because there is no simple agreement on the colour that the "I can't win" message should be.

cgolubi1:
Hokay.  I'll take down the dev site then; if i manage to catch up on your other pulls this month, i can try to help shepherd a conversation about this, but i don't want to do so at the expense of other stuff you're trying to get pushed.

blackshadowshade:
Sounds right. I think that having the colour as a user preference is actually the *right* solution to this, even though it was mentioned by @AdmiralJota a little flippantly.

PR #2274, opened 2017-10-19
Added forum markup to link to other forum threads

### LINKED ISSUE #1905: Create forum syntax to link other forum
let's add [thread=172] to the forum/chat syntax to more easily route people to new or old conversations. Like "we digress let's talk about that over here X" or "ya we had that discussion here X go review the arguments" 

It would allow me in game chat to say "go here X to learn about my tournament" or in a forum say "go here X to pick up a new UBFC fighter"

Also If we combo that with the ability to Pin or Sticky a post we can make a table of contents at the top of each forum: i.e. Adoptions, Golf, UBFC, how to help Dev

blackshadowshade:
Fixes #1905.

The current solution only allows linking to forum threads, not posts. To allow links to posts, the simplest way would be to not escape the parameter, but I'm uncertain whether we should be allowing unescaped parameters here.

Edit: the new solution allows linking to forum posts too.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1395/ (if it passes)

### DIFF
--- a/src/ui/js/Env.js
+++ b/src/ui/js/Env.js
@@ -353,6 +353,19 @@ Env.applyBbCodeToHtml = function(htmlToParse) {
       'closingHtml': '</a>',
       'escapeParameter': true,
     },
+    'forum': {
+      'isAtomic': false,
+      'isLink': true,
+      'openingHtml':
+        '<a class="chatForumLink" href="forum.html#!threadId=###">' +
+        'Forum thread: ',
+      'alternateOpeningHtml':
+        '<a class="chatForumLink" href="forum.html#!threadId=###&postId=$$$">' +
+        'Forum thread: ',
+      'closingHtml': '</a>',
+      'escapeParameter': true,
+      'multipleParameters': true,
+    },
     '[': {
       'isAtomic': true,
       'openingHtml': '[',
@@ -360,7 +373,7 @@ Env.applyBbCodeToHtml = function(htmlToParse) {
   };
 
   var outputHtml = '';
-  var tagStack = [ ];
+  var tagStack = [];
 
   // We want to build a pattern that we can use to identify any single
   // BB code start tag
@@ -423,8 +436,23 @@ Env.applyBbCodeToHtml = function(htmlToParse) {
         tagName = tagStack.pop();
         outputHtml += replacements[tagName].closingHtml;
       } else {
+        var secondTagParameter = '';
+        if (replacements[tagName].multipleParameters) {
+          var parameters = tagParameter.split(',');
+
+          tagParameter = parameters[0];
+          if (parameters.length > 1) {
+            secondTagParameter = parameters[1];
+          }
+        }
+
         var htmlOpeningTag;
-        if (tagParameter && replacements[tagName].alternateOpeningHtml) {
+        // Some replacements have optional parameters and produce different
+        // opening HTML depending on whether or not all the parameters were
+        // provided.
+        if (replacements[tagName].alternateOpeningHtml &&
+            ((replacements[tagName].multipleParameters && secondTagParameter) ||
+            (!replacements[tagName].multipleParameters && tagParameter))) {
           htmlOpeningTag = replacements[tagName].alternateOpeningHtml;
         } else {
           htmlOpeningTag = replacements[tagName].openingHtml;
@@ -434,14 +462,23 @@ Env.applyBbCodeToHtml = function(htmlToParse) {
           // We need to HTML decode the parameter before we URI encode it,
           // and the easiest way is to pretend we're going to render it
           var tempDiv = $('<div>');
+
           tempDiv.html(tagParameter);
           var decodedTagParameter = tempDiv.text();
           htmlOpeningTag = htmlOpeningTag.replace(
             '###',
             encodeURIComponent(decodedTagParameter)
           );
+
+          tempDiv.html(secondTagParameter);
+          var decodedSecondTagParameter = tempDiv.text();
+          htmlOpeningTag = htmlOpeningTag.replace(
+            '$$$',
+            encodeURIComponent(decodedSecondTagParameter)
+          );
         } else {
           htmlOpeningTag = htmlOpeningTag.replace('###', tagParameter);
+          htmlOpeningTag = htmlOpeningTag.replace('$$$', secondTagParameter);
         }
         outputHtml += htmlOpeningTag;
         if (!replacements[tagName].isAtomic) {
@@ -503,8 +540,8 @@ Env.buildProfileLink = function(playerName, textOnly) {
 
 // Utility function to build a vacation image object
 Env.buildVacationImage = function(size) {
-  var image = (size=='large') ? 'vacation22.png' : 'vacation16.png';
-  return  $('<img>', {
+  var image = (size == 'large') ? 'vacation22.png' : 'vacation16.png';
+  return $('<img>', {
     'src': Env.ui_root + 'images/' + image,
     'class': 'playerFlag',
     'title': 'On Vacation'
@@ -564,7 +601,7 @@ Env.toggleSpoiler = function() {
 // function (not counting the callback argument, which will be handled by
 // callAsyncInParallel).
 Env.callAsyncInParallel = function(functions, finalCallback) {
-  var completedFunctions = [ ];
+  var completedFunctions = [];
 
   // First, initialize all the function statuses before we start executing
   // anything
@@ -581,7 +618,7 @@ Env.callAsyncInParallel = function(functions, finalCallback) {
     if (typeof(value) == 'function') {
       functionDetails = {
         'func': value,
-        'args': [ ],
+        'args': [],
       };
     } else {
       functionDetails = value;
--- a/src/ui/gui.css
+++ b/src/ui/gui.css
@@ -855,6 +855,11 @@ div.forum div.splitRight {
     width: 40%;
 }
 
+div.forum div.splitRight span.postTag {
+  color: #e0e0e0;
+  margin-right: 2em;
+}
+
 /* END: Forum*/
 
 /* BB Code (OK technically also forum but chat too) */
--- a/src/ui/js/Forum.js
+++ b/src/ui/js/Forum.js
@@ -826,6 +826,8 @@ Forum.buildPostRow = function(post) {
     'text': postDates,
   }));
   var buttonHolder = $('<div>', { 'class': 'splitRight', });
+  var postTag = '[forum=' + Api.forum_thread.threadId + ',' + post.postId + ']';
+  buttonHolder.append($('<span>', { 'class': 'postTag', 'text': postTag, }));
   if (post.posterName == Login.player && !post.deleted) {
     var editButton = $('<input>', {
       'type': 'button',
@@ -913,6 +915,12 @@ Forum.buildHelp = function() {
   helpDiv.append($('<div>', {
     'class': 'help',
     'html': '[set=Soldiers]: <a href="buttons.html?set=Soldiers">Soldiers</a>',
+  }));
+  helpDiv.append($('<div>', {
+    'class': 'help',
+    'html': '[forum=1,6]text[/forum]: ' +
+            '<a href="forum.html#!threadId=1&postId=6">' +
+            'Forum thread: text</a>',
   }));
   helpDiv.append($('<div>', {
     'class': 'help',


blackshadowshade:
Also, I'm aware that the appearance of the link as "Forum thread 357" is not really ideal, but anything more would require some of Jota's magic, I think.  :)

AdmiralJota:
Thoughts on both of those points:

One way to handle linking to individual posts would be to add another property to the replacements objects to indicate that a replacement allows for multiple parameters. So we might define it as something like:

    'forum': {
      'isAtomic': true,
      'isLink': true,
      'openingHtml':
          '<a class="chatForumLink" href="forum.html#!threadId=###&postId=$$$">' +
          'Forum thread ',
      'closingHtml': '</a>',
      'escapeParameter': true,
      'multipleParameters': true,
    },

And then use it like `[forum:357,5825]` to generate the URL http://buttonweavers.com/ui/forum.html#!threadId=357&postId=5825. That *would* require additional magic, but I can look into doing that if you think it would be worthwhile.

For the second point about the appearance of the link: one option there would just be to set 'isAtomic' to false. Then the user would be expected to use both opening and closing tags, and it could display whatever was in between them as the link text (presumably keeping the "Forum thread: " prefix in the opening tag).

The downsides to this would be that the user could make up whatever text they wanted -- e.g., lying about what thread it linked to -- and everyone would need to include the text by hand. Dunno if those are acceptable downsides; I think it might be better than just bare internal ID numbers, though.

blackshadowshade:
For the first point, I think that we should theoretically be able to deal with both syntaxes, i.e., [forum=357] and [forum=357,5825].

For the second, I'm leaning towards setting isAtomic to false, because we really want the writer to specify what they're linking to.

AdmiralJota:
Does this do the trick?

https://gist.github.com/AdmiralJota/4956978412318049278989a3968cfe14

blackshadowshade:
At line 456, you're missing an open parenthesis. Where should it go?

blackshadowshade:
To get the code running, I put an extra close bracket right at the end of the conditional.

So, I put in [forum=1]test[/forum] and the whole page stops loading. The error is 

    TypeError: undefined is not an object (evaluating 'replacements[tagName].multipleParameters')

No matter what should happen with a syntax error, an error that stops an entire forum page from loading is rather catastrophic and should be caught and dealt with.

AdmiralJota:
Indeed. 

The parenthesis does belong at the end of the conditional.

Do you happen to have the line number where that error occurred?

blackshadowshade:
So, let me tell you what I'm getting during testing.

[forum=14] works, and gives the following link:

    <a class="chatForumLink" href="forum.html#!threadId=14">Forum thread: </a>

[forum=14][/forum] fails, giving the following error in Firefox Developer Edition:

    TypeError: replacements[tagName] is undefined[Learn More] jquery-1.9.0.min.js%20line%201%20%3E%20eval:434:1

but I think the line number here is probably not useful.

AdmiralJota:
Aha, I think I see the problem. I've updated the gist at https://gist.github.com/AdmiralJota/4956978412318049278989a3968cfe14

If that doesn't solve it, I'll fire up a local copy of the site and investigate it this weekend.

blackshadowshade:
Okay, that looks much better! I'll update the pull request now.

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1395/ (if it passes)

src/ui/js/Env.js
```diff
@@ -625,4 +662,4 @@ Env.validateUrl = function(url) {
   // This also doesn't verify that the URL points to a safe page, but that is
   // outside of the scope of this function.
   return url;
-};
+};
```
cgolubi1: Looks like you lost the trailing whitespace on Env.js.  If i'm reading the diff right, please fix that.

cgolubi1:
General comment: if you're willing to put in separate PRs for whitespace-only changes when you touch new files, it would make PRs like this one a lot easier to read.  I know that relies on me getting to those quickly to not slow you down too much, so i understand if you can't, but if you can and flag things as "whitespace only" and they pass Jenkins, we should be able to just merge them.

cgolubi1:
Anyway, fix the trailing newline and rebase for conciseness, and i'll stand up a dev site for this.

blackshadowshade:
Thanks for the heads-up about the whitespace changes, I thought I'd reverted them, but apparently they got through again.

A lot of these whitespace changes have either been caused by Jota's formatter, or by the fact that I've copied-and-pasted out of a gist. I'm not certain which. Most if not all of the whitespace changes are inconsistent with our standard format, so I'll revert them.

blackshadowshade:
Rebased.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1399/ (if it passes)

cgolubi1:
Thanks, that's much more legible.  Putting up a dev site now.

cgolubi1:
Site is up at 54.166.151.254 - irilyth, can you add a 'blip' alias pointing to that?

irilyth:
> Site is up at 54.166.151.254 - irilyth, can you add a 'blip' alias pointing to that?

Done!

cgolubi1:
So this seems to work at a glance based on shadowshade's dev site testing.  Does it need to do anything other than seem to work and pass the tests?  I vote we merge it.  Anyone think otherwise?

dwvanstone:
I just took a look at this. The links do go someplace. How would I know if the test link "threadId=1&postId=19" is correctly going to post ID 19? How can a user use this to create a link to a specific post?

blackshadowshade:
If you hover over the star to the left of a user's name (to the left of a post), you get the threadId and postId.

blackshadowshade:
So, testers, are we all good with this feature?

irilyth:
Hm, this seems basically fine, but I was slightly surprised at first to have to put in text and a closing [/forum]; but I guess that's the right way to do this. You could potentially allow just [forumthread=6] and have it automatically fill in the thread name as the text, but, meh.

I think it's ok as is if everyone else does.

irilyth:
One other thought is that it'd be slightly simpler if the syntax matched what's in the URL, so if I was looking at http://blip.dev.buttonweavers.com/ui/forum.html#!threadId=4&postId=28, I could copy and paste the "threadId=4&postId=28" part, and not have to edit that a bunch. In the way that I can copy the "button=IIconfused" part out of http://blip.dev.buttonweavers.com/ui/buttons.html?button=IIconfused, the "game=50" part out of http://blip.dev.buttonweavers.com/ui/game.html?game=50, etc. This feels more cumbersome to use.

blackshadowshade:
What do other testers think about the syntax? What are you expecting to have to put in?

dwvanstone:
I like the brevity of typing as it stands now.   [forum=1,19]TEXT[/forum]  seems clear enough. I think having to type out [forum=1,threadID=19] would be extra typing. Plus I wouldn't want us to mix the markup code with whatever the code actually uses in the URL (allowing us to change the URL as needed in the future without changing the markup code).

irilyth:
I also want to minimize extra typing; my thinking is that I'm mostly going to be copying and pasting out of the URL, and so if I have to type a bunch of stuff to delete a bunch of stuff from the URL that isn't in the markup link, that's extra typing. :^ )

dwvanstone:
What URL will you be copying from? I mean, when you're trying to link to a particular post you won't have its postID in the URL.

dwvanstone:
> If you hover over the star to the left of a user's name (to the left of a post), you get the threadId and postId.

It's not intuitive to me that I have to convert "threadId" to "forum=". I would expect there to be forumID, threadID, and a postID for every post.

Also, if I'm on a mobile device, I can't hover over the star to find a particular postID.

irilyth:
If I click on the star next to a post, the ID shows up in the URL bar, and I can then copy the whole "threadId=4&postId=28" business out of there.

dwvanstone:
[However, I don't have a strong opinion about this issue and am happy to let it go out as is. I just think it may need some explanation in the forum as to how to make it work.]

irilyth:
> I would expect there to be forumID, threadID, and a postID for every post.

Thread IDs are unique across forums; there is only one thread ID 4, so you don't need to specify the forum at all.

Forums do have IDs, but those show up as boardId, if you click on a link to a forum (and not a thread or a post). They don't show up at all when you click on a thread or a post.

dwvanstone:
You can click on those stars? Cool! That solves the mobile issue too.

dwvanstone:
> Thread IDs are unique across forums; there is only one thread ID 4, so you don't need to specify the forum at all.

The forum example looks like this: 

[forum=1,6]text[/forum]: Forum thread: text

I need to put two numbers after the equal sign. My best guess would be that the first is the forum id (1=Misc Chatting, 2=Gameplay, etc.) and the second is the thread id. (As a casual user I wouldn't know that thread IDs are unique across forums.)

As a superuser feature this is great, but it's not clear what the numbers mean or how to get them for the newcomer. 

I wasn't clear what the proposal was when copying out of a URL. What would be an example markup?

irilyth:
> As a superuser feature this is great, but it's not clear what the numbers mean or how to get them for the newcomer.

Agreed.

> I wasn't clear what the proposal was when copying out of a URL. What would be an example markup?

Mm, good question. My intution would be something like `[forum threadId=4]` for a thread, or `[forum threadId=4&postId=28]` for a post, which is conveniently just like what's in the URL; but, is not much like the other things we have now, which are more like either `[thing]` or `[thing=stuff]`. Hmm.

AdmiralJota:
For intuitiveness, how about if the tag were called "thread" instead of "forum", so you could call it as `[thread=4]` or `[thread=4,28]`?

(By the way, if I remember correctly, the second parameter is optional: you don't need to specify a specific post in the thread unless you want to.)

blackshadowshade:
Yes, that's correct, the second parameter is optional.

dwvanstone:
I think [forum=] is clearer than [thread=], since the other tags are things like [game=], [wiki=], and [issue=]. It wouldn't be clear what a "thread" is without the context. 

I think this is good as it's going to get, but I recommend a paragraph in a forum explaining how it works for the curious newcomer.

blackshadowshade:
Do people think there should be an additional hint at the side of the page for links to forum threads, showing the simpler

    [forum=2]

syntax?

irilyth:
I'm pretty happy with `[forum threadId=4]` for a thread, or `[forum threadId=4&postId=28]` for a post, FWIW.

cgolubi1:
I think the next step here is for someone to summarize all the feedback and suggest a path forward.  If no one else has done that by the time i check in next, i'll do so, but it looks like a bunch of people are participating, and i'm comfortable with whatever the consensus is.

blackshadowshade:
I would be okay with @irilyth's suggestion. Now, let's see whether it's possible. @AdmiralJota?

AdmiralJota:
I don't think it would really make sense to implement, to be honest. The syntax for forum markup is basically [tag], [tag=value] or [tag=value,value]. It would require a lot of special-casing to try to incorporate the HTTP URL query string syntax into it.

irilyth:
Hm, ok; how about [forum=threadId=4] for a thread, or [forum=threadId=4&postId=28] for a post? Sort of funny-looking, but basically everything after the equals sign is just the value of the tag.

blackshadowshade:
What about [forumthreadId=4] and [forumthreadId=4,postId=28]?

AdmiralJota:
That might be a bit more doable, but would it really solve irilyth's problem?

AdmiralJota:
A different thought: what if each forum post had a bit of invisible (e.g., white-on-white) in it somewhere that contained the actual forum markup for that post? E.g.:

image

irilyth:
Ooh, that's clever. Maybe make it very light gray, so you might find it even if you didn't know it was there, but it'd still be pretty unobtrusive?

irilyth:
Either way, that now has my vote.

blackshadowshade:
That's a nice idea, @AdmiralJota! I'd be fine with that, if you can tell me what changes are necessary to get it working.

AdmiralJota:
I think these should do the trick: https://gist.github.com/AdmiralJota/49f1e4e37294151cc0598aa43a88ed6b

If you want the grey to be lighter or darker, you can just replace the `#e0e0e0;` with a different color (like `#f0f0f0;` or `#d0d0d0;`).

blackshadowshade:
I've added @AdmiralJota's changes without doing any testing myself.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1466/ (if it passes)

cgolubi1:
This is on my plate to update the dev site when i have a moment.

cgolubi1:
Okay, i think i updated this dev site.  Check it out.

irilyth:
I like the look of it. One thing about the behavior of it: In Chrome, if I triple-click on it, it selects the markup text (good) but also the "Posted at 2017-12-10 21:19:58." text (bad). If that text isn't in its own cell (or div of whatever), would that be easy to add?

AdmiralJota:
> I like the look of it. One thing about the behavior of it: In Chrome, if I triple-click on it, it selects the markup text (good) but also the "Posted at 2017-12-10 21:19:58." text (bad). If that text isn't in its own cell (or div of whatever), would that be easy to add?

Hmm, that's odd: the text *is* in its own div:

    <div class="postFooter">
      <div class="splitLeft">Posted at 2017-12-10 21:19:58. </div>
      <div class="splitRight">
        <span class="postTag">[forum=4,28]</span>
        <input type="button" value="Quote">
      </div>
  </div>

I'm not sure why Chrome would be doing that. In Firefox, triple-clicking works as expected.

irilyth:
Hm, drat! I confirmed that it works fine with Firefox even in my Linux VM, and does not work on Chrome even under Mac OS. Bah.

blackshadowshade:
Unfortunately, triple click behaviour is not standardised across browsers (obviously!). If we really cared about it, we could write a listener for a triple click event.

(see https://css-tricks.com/snippets/jquery/triple-click-event/ or https://stackoverflow.com/questions/6480060/how-do-i-listen-for-triple-clicks-in-javascript)

However, if there's a simpler solution, that would be preferable.

I suggest that someone saves a minimal static HTML version of the relevant part of the forum page so that we can tweak the HTML manually and test triple clicks in Chrome on this.

blackshadowshade:
@irilyth, do you see the triple-click behaviour as being a blocker for this pull request? Or are you happy to have this accepted as-is, with the possibility of tweaking this in the future?

irilyth:
Given that we don't have an easy fix for the triple-click behavior, I'm fine with this going out as-is.

blackshadowshade:
I'll open an issue regarding triple-click behaviour on the forum links.

cgolubi1:
Where are we here?  The last comment was from irilyth on 28 Feb saying that the triple-click behavior isn't a blocker.  Do we have votes in favor of merging?  Do we have any outstanding votes against merging?

irilyth:
I'm still fine with this going out as-is. :^)

cgolubi1:
Thanks, irilyth.  No other complaints, so i'm going to go ahead and merge this.

PR #2275, opened 2017-10-21
Removed 'changed size due to turbo' message

### LINKED ISSUE #2217: Log says Turbo die changed size when it didn't
In http://www.buttonweavers.com/ui/game.html?game=24235 I made an attack with a Turbo (Z=8)! die, and typed nothing in the box for a changed Turbo size.  The game correctly kept it at 8 sides, but the log reported it this way: 

> jimmosk performed Power attack using [(Z=8)!:6] against [(X=20):5]; Defender (X=20) was captured; Attacker (Z=8)! rerolled from 6, changed size due to turbo

> Turbo die (Z=8)! remained the same size, rolled 5

(because the log is reverse chronological order, that second message is actually above the first, but I switched the order to read more clearly here)

Can that spurious "changed size due to turbo" be removed?

blackshadowshade:
Fixes #2217.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1386/ (if it passes)

### DIFF
--- a/src/engine/BMGameAction.php
+++ b/src/engine/BMGameAction.php
@@ -516,7 +516,7 @@ class BMGameAction {
                 !($attackType == 'Trip')) {
                 $this->message_append(
                     $postEventsAttacker,
-                    'rerolled from ' . $attackerInfo['value'] . ', changed size due to turbo'
+                    'rerolled from ' . $attackerInfo['value']
                 );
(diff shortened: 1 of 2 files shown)


cgolubi1:
I actually don't see any reason not to just merge this.

PR #2279, opened 2017-10-28
Added functionality for Gordo and Guillermo

### LINKED ISSUE #265: Unique die skill needed
Unique: u
I think unique is similar to weak, in that it was a skill created to handle rules that were simply spelled out on the button.

---

(Implemented, but not found listed in the original site captures, nor in the randomdice buttonmen roster)

Unique dice were created as a way to handle Guillermo's special rules[my best guess, not verified].  The unique skill applies when a button recipe has two or more swing dice [can it also be applied to option dice?].  Swing selection will fail if any two of the unique dice were assigned the same size.

Questions: Can Unique be assigned to dice other than swing [or option]?  example 'u(10) u(X)' meaning the X can be given any value other than 10
Can Unique be assigned to Mood Swing?  to Turbo?

blackshadowshade:
Fixes #265.

This pull request enables Gordo and Guillermo. It enforces the uniqueness of all die sizes for Gordo, and just swing values for Guillermo. It also stops players from accepting an aux die for Gordo that is a swing die with a type already in Gordo's recipe.

I've tested this with Gordo vs Guillermo, Gordo vs King Arthur, and Gordo vs Lancelot.

In addition, I've removed an unused argument in Giant's button special, which I noticed when I was using it as a template for these button specials.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1398/ (if it passes)

### DIFF
--- a/src/engine/BMGame.php
+++ b/src/engine/BMGame.php
@@ -910,6 +910,27 @@ class BMGame {
     protected function set_swing_values() {
         foreach ($this->playerArray as $player) {
             if (!$player->waitingOnAction) {
+                // check for unique swing values
+                if (!self::are_unique_swing_values_valid($player)) {
+                    $hookResult = $player->button->run_hooks('unique_swing_setting_fail_message', array());
+
+                    if (is_array($hookResult) &&
+                        array_key_exists('BMBtnSkill'.$player->button->name, $hookResult) &&
+                        array_key_exists(
+                            'unique_swing_setting_fail_message',
+                            $hookResult['BMBtnSkill'.$player->button->name]
+                        )
+                       ) {
+                        $this->message = $hookResult['BMBtnSkill'.$player->button->name]
+                                                    ['unique_swing_setting_fail_message'];
+                    } else {
+                        $this->message = 'The swing values chosen are not unique';
+                    }
+                    $player->swingValueArray = array();
+                    $player->waitingOnAction = TRUE;
+                    return;
+                }
+
                 // apply swing values
                 foreach ($player->activeDieArray as $die) {
                     if (isset($die->swingType)) {
@@ -925,6 +946,25 @@ class BMGame {
                 }
             }
         }
+    }
+
+    /**
+     * Are unique swing values valid?
+     *
+     * @param BMPlayer $player
+     * @return bool
+     */
+    protected static function are_unique_swing_values_valid(BMPlayer $player) {
+        $hookResult = $player->button->run_hooks(
+            __FUNCTION__,
+            array(
+                'activeDieArray' => $player->activeDieArray,
+                'swingValueArray' => $player->swingValueArray
+            )
+        );
+
+        return (!isset($hookResult['BMBtnSkill'.$player->button->name][__FUNCTION__]) ||
+                $hookResult['BMBtnSkill'.$player->button->name][__FUNCTION__]);
     }
 
     /**
@@ -2776,7 +2816,7 @@ class BMGame {
     protected static function is_button_slow($button) {
         $hookResult = $button->run_hooks(
             __FUNCTION__,
-            array('name' => $button->name)
+            array()
         );
 
         $isSlow = isset($hookResult['BMBtnSkill'.$button->name]['is_button_slow']) &&
--- a/src/engine/BMBtnSkillGordo.php
+++ b/src/engine/BMBtnSkillGordo.php
@@ -1,0 +1,83 @@
+<?php
+/**
+ * BMBtnSkillGordo: Code specific to Gordo
+ *
+ * @author: james
+ */
+
+/**
+ * This class currently supports the special skills of Gordo
+ */
+class BMBtnSkillGordo extends BMBtnSkillUniqueSwing {
+    /**
+     * An array containing the names of functions run by
+     * BMCanHaveSkill->run_hooks()
+     *
+     * @var array
+     */
+    public static $hooked_methods = array(
+        'are_unique_swing_values_valid',
+        'can_button_add_this_aux_die',
+        'unique_swing_setting_fail_message'
+    );
+
+    /**
+     * Hooked method applied when checking if a button has unique swing
+     *
+     * @param array $args
+     * @return array
+     */
+    public static function are_unique_swing_values_valid(array $args) {
+        $parent_result = parent::are_unique_swing_values_valid($args);
+
+        if (!$parent_result[__FUNCTION__]) {
+            return array(__FUNCTION__ => FALSE);
+        }
+
+        foreach ($args['activeDieArray'] as $die) {
+            if (!($die instanceof BMDieSwing)) {
+                if (in_array($die->max, $args['swingValueArray'])) {
+                    return array(__FUNCTION__ => FALSE);
+                }
+            }
+        }
+
+        return array(__FUNCTION__ => TRUE);
+    }
+
+    /**
+     * Hooked method applied when checking if a button restricts decisions on
+     * auxiliary dice
+     *
+     * @param array $args
+     * @return array
+     */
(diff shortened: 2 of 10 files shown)


blackshadowshade:
Heads up for @Hurtay that Gordo should be available soon.  :)

cgolubi1:
This didn't pass Jenkins.

blackshadowshade:
http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1400/ (if it passes)

cgolubi1:
Okay, let me see if i can just use the replay test engine without modification to look for regressions here.  This seems like a straightforward application of replay testing.

cgolubi1:
Oh, hmm, i seem to have gotten confused about this --- i tried replay testing, and it failed because it doesn't know how to deal with Gordo wanting unique swing values.  But in fact what i should be doing is temporarily skipping Gordo and Guillermo, and seeing if the rest of the replay testing works.

Then before merging i'll need to make the client support those buttons.

cgolubi1:
Okay, i replayed about 4000 games without Gordo and Guillermo, and saw no problems.  Next step is to make RandomAI able to support buttons with the unique skill.

blackshadowshade:
Would you be able to put up a dev site for this so that the testers can get onto it?

cgolubi1:
Sure, will do.

cgolubi1:
irilyth: can you sync up 'temporary' to 54.167.234.232 in route53?

irilyth:
> irilyth: can you sync up 'temporary' to 54.167.234.232 in route53?

Done!

dwvanstone:
This is great! I've done some testing up on 54.167.234.232.

This is the only issue I've found:

 Gordo v. Michigan, accepting auxiliary dice. When selecting swing dice, Gordo has (V) (W) (X) (Y) (Z) (6). If I select V=6, W=7, X=8, Y=9, Z=10, it gives me an error: "The swing values chosen are not unique." Perhaps a clearer error would be "The swing values must be unique and not be the same size as any other dice." (I.e., it should be a different error from when Guillermo has non-unique swing dice.)

blackshadowshade:
After a great deal of consideration of how to reword the message, with the proviso that it not be very long (and thus take up vertical real estate on small screens), I've changed the message to 

"Cannot have multiple dice with the same size"

Are you okay with that, @dwvanstone?

@cgolubi1, I've got a Jenkins run for this:

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1438/ (if it passes)

However, I've neither rebased nor merged master. Once we're done with testing, I'll rebase this and run a fresh Jenkins run.

cgolubi1:
Sounds good - thanks for not rebasing - i'll update the dev site.

src/engine/BMGame.php
```diff
@@ -912,7 +912,11 @@ protected function set_swing_values() {
             if (!$player->waitingOnAction) {
                 // check for unique swing values
                 if (!self::are_unique_swing_values_valid($player)) {
-                    $this->message = 'The swing values chosen are not unique';
+                    if ('Gordo' == $player->button->name) {
```
cgolubi1: Hmm, wait... i'm not going to update the dev site for now, because i'm confused about how this fix is relevant, and i think we can talk about it on this issue just as easily.  Does this have anything to do with whether the button is Gordo?  I thought the problem was that the opposing button had an auxiliary die.  Why would Guillermo not have the same problem?

cgolubi1: Oh, i see what you're saying from the original statement of the pull --- the rules are actually different for Gordo vs Guillermo.  Well, okay, i will update the dev site then.  Is there really no better way to identify buttons that work like Gordo, though?  Grepping by button name deep in the backend seems really clunky.

dwvanstone:
> "Cannot have multiple dice with the same size"

Yes, that is perfect. Thanks!

src/engine/BMGame.php
blackshadowshade: Last night, I was thinking that it would be okay, it's only a once-off. However, thinking about it now, I think I agree that it's a bit clunky. I'll see if I can implement some sort of hooked function to move the logic to the BtnSkill classes.

blackshadowshade:
Right, that logic has been shifted into a button hook.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1439/ (if it passes)

cgolubi1:
Thanks.  I'll update the dev site (should be a noop) and the replay site i'm working on.

cgolubi1:
Worth noting is something i realised when i was making the change for #2305 --- unless i'm missing something, there's no API way to find out what button skills a particular button has (you can tell which button skills are in the game, but the API doesn't report which button has the skills).  So API clients/bot players are going to have to do what RandomAI does here, and look at the player's button name.  I suppose i could also check if Gordo or Guillermo is in the button skills just for fun, but that seems like extra work for no benefit, since the button skills are named after the buttons.

cgolubi1:
Anyway.  I'm going to okay this for merging once my replay site playes a few hundred Gordo or Guillermo games without falling over, which is what it's trying right now.  I'm taking dwvanstone's comments above as approval, so i think we're good to go here.

cgolubi1:
```
Starting game 24
  File "./test_log_games", line 71, in <module>
    c.log_test_game(i, button1, button2)
  File "lib/random_ai.py", line 1545, in log_test_game
    self.next_game_action()
  File "lib/random_ai.py", line 1533, in next_game_action
    elif state == 'CHOOSE_AUXILIARY_DICE': self._game_action_choose_auxiliary_dice()
  File "lib/random_ai.py", line 1406, in _game_action_choose_auxiliary_dice
    return self._game_action_choose_auxiliary_dice_player(chosen_client, chosen_array)
  File "lib/random_ai.py", line 1105, in _game_action_choose_auxiliary_dice_player  
    retval and retval.message or "NULL"))
  File "lib/random_ai.py", line 1444, in bug
    raise ValueError, message
ValueError: API choose_auxiliary_dice(24, add, 5) unexpectedly failed: This button cannot add this auxiliary die
```

The game is Lancelot vs Gordo.  How do we feel about that?  I feel that the behavior is okay --- i tried this game in the UI (see screencap) and it doesn't look that unclean.  But (1) obviously i'll have to make another change to RandomAI to work with this behavior, and (2) on the PR side, i think Gordo's flavor text should make it clear that this is something that can happen, so that when players see this error, they can mouseover the button special description and get a hint about what happened.

cgolubi1:
Okay, i have a random_ai.py modification in a working directory that has played 100 games of Gordo vs. Auxiliary without blowing up, and wound up choosing the Aux in a small handful of them (spot-check: Gordo vs. Jota with the +(12), which sounds legitimate).  I'll go back to full replay testing with that change, and assuming it looks good, i'll commit it.

I'm still requesting the modification to the button skill description clarifying that the Gordo player is required to decline an auxiliary die which would have the same size as another die.

blackshadowshade:
I've updated the description. I've also rebased this because I'm about to head off on holidays, and I didn't want to hold up accepting this pull request.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1442/ (if it passes)

cgolubi1:
LGTM.  I'll check in on my replay site.

cgolubi1:
It's doing great; i'll submit a PR for my random_ai change.

cgolubi1:
Passes, and the only change is the mentioned description one:
```
79c79,81
<         return 'None of Gordo\'s dice can be the same size.';
---
>         return 'None of Gordo\'s dice can be the same size. ' .
>                'Note that a player playing Gordo is required to decline an auxiliary die ' .
>                'that would force two of Gordo\'s dice to be the same size';
```

PR #2280, opened 2017-10-31
Added subdieArray to API response for captured and out of play dice

### LINKED ISSUE #2188: Captured twin swing that have split asymmetrically do not display correctly
In the image below, I have a captured (X=3,X=4) die, caused by splitting a (X=7,X=7) via radioactive. However, it's showing as a (X=3.5,X=3.5) die because the backend is not supplying the subdie sizes to the front-end for captured dice.

Active die:
>value	4
>sides	5
>skills	[]
>properties	[1]
>0	"Twin"
>recipe	"(X,X)"
>description	"Twin X Swing Die (with 3 and 2 sides)"
>subdieArray	[2]
>0	
>sides	3
>value	3
>1	
>sides	2
>value	1

Captured die:
>value	4
>sides	7
>recipe	"(X,X)"
>properties	[2]
>0	"WasJustCaptured"
>1	"Twin"

The database entry for the die is correct:
>| 5290 |        4 |                 3 |      36 |         2 | (X,X)   |        1 |     4 |          7 | WasJustCaptured;Twin__{"sides":[3,4],"values":[1,3]} |

Obviously, the thing missing is the subdieArray.

capture

blackshadowshade:
Fixes #2188.

I've tested this with GreatWolf (with the twin reserve die set to Y=3) and Anti-Llama, and the non-symmetric captured twin dice show up correctly.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1401/ (if it passes)


cgolubi1:
A default replay site won't Just Work here, but let's see how close i can get to verifying lack of unrelated regressions, and maybe we'll call that good enough.

cgolubi1:
Oh, FYI, this needs to be rebased - i went to try to setup a new replay site, and noticed the merge conflict.

blackshadowshade:
Thanks for the heads up. I've rebased it, and we'll see what Jenkins thinks:

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1436/ (if it passes)

blackshadowshade:
I'm aware of the responderTest conflict here and will deal with it ASAP.

blackshadowshade:
Okay, this has been rebased.

cgolubi1:
I think i'm in better shape to actually deal with the replay testing on this; relaunching the replay site now.

cgolubi1:
First attempt at an update script to check correctness of subdieArray values.  It's not going to handle every case (in particular it won't DTRT if we have multiple identical twin dice in terms of recipe/sides/value), but it might be enough for the cases we hit in replay testing.

```
#!/usr/bin/python
##### update_replay_games
# be made in random_ai.py.

import sys
sys.path.append('lib')

import random_ai

def dice_seem_identical(die, olddie):
  for key in ['recipe', 'sides', 'value']:
    if die[key] != olddie[key]: return False
  return True

def guess_subdie_array(die, old_active_dice, old_captured_dice):
  # TODO: if necessary, also diff old active dice vs current active dice
  # to look for ones that are now missing
  for olddie in old_captured_dice + old_active_dice:
    if dice_seem_identical(die, olddie):
      return olddie['subdieArray']
  raise ValueError("No old die found for: %s among captured=%s, active=%s" % (die, old_active_dice, old_captured_dice))

# Custom classes
  #
    for i in 0, 1:
      for die in data['playerDataArray'][i]['capturedDieArray']:
        if not ',' in die['recipe']: continue
        if not die['sides']: continue
        die['subdieArray'] = guess_subdie_array(die, self.old_active_die_arrays[i], self.old_captured_die_arrays[i])
    self.old_active_die_arrays = [ data['playerDataArray'][1]['activeDieArray'], data['playerDataArray'][0]['activeDieArray'], ]
    self.old_captured_die_arrays = [ data['playerDataArray'][0]['capturedDieArray'], data['playerDataArray'][1]['capturedDieArray'], ]
    return data

# Initial setup
if len(sys.argv) < 2:

DIRNAME = sys.argv[1]

  d.translate_game(gamepath)
```

blackshadowshade:
I'll be interested to see how this update script deals with the change to the API. From my casual read of it, it doesn't seem to be doing enough---but perhaps it's smarter than I'm giving it credit for.  :)

cgolubi1:
Oh, it breaks on trip cases just like you'd expect.  `:>)`  I'll keep working on it.

cgolubi1:
This actually isn't doing badly --- 7 failures in 8800 games.  Checking out the details now.

cgolubi1:
Hmm, it seems like this worked fine.  The failures were bad guesses about which subdie should work how, and i think they were cases where the button had multiple dice with similar recipes.  Anyway, the ones i spot-checked, your code did the right thing (i.e. the die breakdown carries over correctly from the breakdown on the active die which was captured).  IMO, that's all good enough, and i'm going to go ahead and merge this.

cgolubi1:
Oh, wait, before i merge this, a question: does this actually resolve #2188, or does it just provide the information to the UI so that the UI could be changed to resolve #2188?  I guess i'm surprised that this by itself would fix that issue, because there's no UI change here.

PR #2278, opened 2017-10-22
Addressed remaining PHPDoc warnings

blackshadowshade:
Fixes #1130.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1392/ (if it passes)

PR #2270, opened 2017-10-19
Added spaces to button names before parentheses

blackshadowshade:
Fixes #2196.

Requires database update 02196_button_name_spacing_01.sql

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1381/

PR #2269, opened 2017-10-19
Added description of berserk/turbo interaction

blackshadowshade:
Fixes #2239.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1380/

Issue #2267, opened 2017-10-11
Plan 2017-11 release


cgolubi1:
Do all testing and deployments for the 2017-11 buttonweavers.com prod release.

cgolubi1:
Okay, with #2284 afaik things look good for the November release.  Anyone seen any issues on staging?  I'm planning to push to prod tomorrow.

dwvanstone:
I did kick the tires a bit and see no issues.

PR #2282, opened 2017-11-02
Push changes from master to staging on 2017-11-03

cgolubi1:
This update brings these approved pulls to the staging branch:

* User-visible:
    * #2269: add description of turbo/berserk interaction
    * #2270: add spaces to button names before parentheses
    * #2268: fix bug with fire decline turndown and turbo
* Ideally not user-visible:
    * #2264, #2265: RandomAI fixes
    * #2278: address phpdoc warnings

Database updates:
```
02196_button_name_spacing_01.sql
```

Once this pull is merged, i will load the changes on staging.buttonweavers.com as part of #2267.


cgolubi1:
I'm short on time right now, but i'll try to merge this and get it on staging this evening, so we can do an abbreviated sanity-check of it and then go to prod.

cgolubi1:
Okay, this is merged.  Testers, please check out the staging site.  Since the big change is the fire/turbo bugfix, which shouldn't be super visible, i don't expect there's much to see, but kick the tires a bit.

PR #2284, opened 2017-11-03
Updated button art names to match new button names

### LINKED ISSUE #2283: regression: button art not displayed due to #2270
I realised on the staging site that some of the buttons whose names were changed in #2270 had art.  Button names shouldn't matter for anything in the database, but button art is loaded based on the button name not its ID.  So we'll need to update those files.

blackshadowshade:
Fixes #2283.

I'm pretty sure this won't need a new Jenkins run.


cgolubi1:
I'll go ahead and merge this, but fwiw my hesitation was mostly because i had to think for awhile about whether it might need a jenkins run for any reason (e.g. because responder tests return art filenames) (and i'm still not 100% sure it doesn't, but, if it did, we'll fix it when the next ones start failing).  My future vote is, if we don't have to guess, we don't have to guess wrong.

blackshadowshade:
Okay, adding a Jenkins run for completeness.  :)

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1403/ (and hopefully it passes)

cgolubi1:
Thanks for feeding my paranoia there.  :>)

PR #2291, opened 2017-11-09
Fixed bug with weak dice tripping twin dice

### LINKED ISSUE #2289: logic bug in case of a weak trip 2-sided die attacking only twin dice
Replay testing found something useful for once: when a button whose only trip die is `ht(2)` attacks a button which has only twin dice remaining, `validAttackTypeArray` is reported as containing `Trip`.  This is a bug; there's no legal trip attack.

blackshadowshade:
Fixes #2289.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1410/

### DIFF
--- a/src/engine/BMAttackTrip.php
+++ b/src/engine/BMAttackTrip.php
@@ -104,7 +104,9 @@ class BMAttackTrip extends BMAttack {
             return FALSE;
         }
 
-        $isDieLargeEnough = ($attacker->max >= $defender->min);
+        $attackerPostTripMax = self::post_trip_roll_max($attacker);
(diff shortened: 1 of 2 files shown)


cgolubi1:
Unless i'm missing something, i can regression test this with a default replay site.  Trying that now.

cgolubi1:
There's some sort of skill attack problem with Dead Guy that i've run into on a couple of replay sites so far; i'm sure it was caused by that skill attack fix from last month.  I'll figure that out, but i'm pretty sure it's a false positive for *this* replay site.  Other than that, it replayed 2800 games with no errors, so that's probably good enough.

Once the merge freeze is lifted, i'll merge this.

PR #2293, opened 2017-11-10
Changed wording of dropdown select text for cancelling during fire turndown

### LINKED ISSUE #1709: Fire overshooting cancel option text incorrect
When I use a power attack with optional Fire dice, the third option in the dropdown message is incorrect.

The first option says: "Submit attack without turning down fire dice"
The third option says:  "Don't turn down fire dice (cancelling the attack in progress)"

The third option implies that not turning down the fire die will cancel the attack.

blackshadowshade:
Fixes #1709.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1412/


cgolubi1:
I don't have a problem with this.  Let's just merge it.  It's easy enough to change again if people don't like it.

PR #2308, opened 2017-12-25
Finish RandomAI support for Gordo by supporting mandatory auxiliary die decline

cgolubi1:
* Needed for regression testing of #2279
* No jenkins because the only change is to randomAI


blackshadowshade:
Looks good. Merging.

PR #1924, opened 2016-01-15
Adding a help link

blackshadowshade:
This starts to address part of #137.

Currently, the pull request is incomplete, and I'd like input (and help) before I push this through to completion. I'll attach a screenshot as the first response so that it's easier for people to start commenting.

@cgolubi1, is this going in roughly the right direction, and what sort of testing should we be doing here? @AdmiralJota, how would you make this look nicer? Everyone, what do you think about the text (which comes directly from #1149)?

Of course, there is also text there that will probably be commented out (because the pages don't already exist). I'm happy to also accept comments about obvious stuff like that.  :)

PR #2310, opened 2017-12-28
Added logic to consider stinger dice in fire attack guidance

### LINKED ISSUE #1716: Fire dice & stinger dice: message incorrect
If I use a Fire die and a stinger die is part of a skill attack the message is incorrect. Here's an example from the production site:
screen shot 2015-09-20 at 9 33 13 am

It says "Turn down Fire dice by a total of 2 ..."  However since a stinger die is involved in the attack I can turn down the Fire dice anywhere from 2 through 5 points. 

The code did allow me to turn it down by 4 points so I believe the back end correctly allows it; I think only the message is incorrect.

blackshadowshade:
Fixes #1716.

I've tested this with wranklepig vs wranklepig.

As part of the logic, I've included what should work with a button with twin stinger and a fire die, but I haven't been able to test this because there is no such button at the moment.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1443/ (if it passes)

### DIFF
--- a/src/ui/js/Game.js
+++ b/src/ui/js/Game.js
@@ -984,10 +984,24 @@ Game.actionAdjustFireDiceActive = function() {
 
   var attackerSum = 0;
   var attackerDiffFromMax = 0;
+  var stingerMax = 0;
+  var stingerMin = 0;
+  var hasStingerDice = false;
   $.each(Api.game.player.activeDieArray, function(i, die) {
     if (die.properties.indexOf('IsAttacker') >= 0) {
       attackerDiffFromMax += die.sides - die.value;
       attackerSum += die.value;
+
+      if (die.skills.indexOf('Stinger') >= 0) {
+        hasStingerDice = true;
+
+        if (die.properties.indexOf('subdieArray') >= 0) {
+          stingerMin += die.properties.subdieArray.length;
+        } else {
+          stingerMin += 1;
+        }
+        stingerMax += die.value;
+      }
     }
   });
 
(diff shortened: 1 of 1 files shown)


cgolubi1:
Is there a responder test that tests this case?  If so, we might want to consider making a QUnit test.

Regardless, it seems like i can stand up a dev site for this.

cgolubi1:
irilyth: can you connect temporary 54.224.152.182 in route53 when you get a second?

blackshadowshade:
As far as I can see, I don't think there's a responder test for this. It's probably a good idea to add a minimal test of wranklepig vs wranklepig, although we *could* expand responder test 008, which has our favourite button, BlackOmega.

irilyth:
> irilyth: can you connect temporary 54.224.152.182 in route53 when you get a second?

Done!

blackshadowshade:
So, testers, what do you think of this feature?

dwvanstone:
I'll try to take a look at this later this week.

dwvanstone:
The issue of not being asked to use Fire die is still there. I tried playing wranklepig vs. wranklepig. 

I had these buttons:

Fo(13) showing 8
gc(7) showing 4
nt(5) showing 1

I selected the gc(7) and nt(5) and performed a skill attack against a die showing 5. I believe I should have been allowed to lower my Fire die in order to make that attack, but I wasn't allowed to lower the Fire die.

dwvanstone:
Am I correct that the text is now "Your turn to complete an attack by adjusting fire dice" for all cases using the Fire dice? I think that will work for me, a seasoned Fire die player, but I worry that will make it harder to learn this skill. But it does fix the bug.

blackshadowshade:
Okay, I hadn't realised that there are two issues here: the message and the possible overshooting on a skill attack with stinger. I've only fixed the message here. I'll spawn a second issue for the overshooting problem.

blackshadowshade:
#2328 opened.

blackshadowshade:
As for the text

> Your turn to complete an attack by adjusting fire dice

there's no change here. Do you think the text should change when the minimum turndown is zero?

dwvanstone:
> Do you think the text should change when the minimum turndown is zero?

I think I'm getting myself confused about the various cases. Is there another site (besides production) which has the old behavior that I can test on?

blackshadowshade:
Try staging.

blackshadowshade:
Also, regarding long-term documentation (see #1924), I can imagine a section devoted to fire-related issues. Alternatively, if someone felt like writing an article about this for the BM blog (http://buttonweavers.blogspot.com), I think this would be a great place for it in the interim.

dwvanstone:
Great. I have found the changed text and it looks perfect.

cgolubi1:
This is a simple change, it's UI-only, it passes Jenkins, and dwvanstone says it's an improvement.  Merging it.

PR #2324, opened 2018-01-22
Retain chat on change of chat privacy

### LINKED ISSUE #2296: Changing chat settings should preserve chat textbox
I'd prefer it if hitting "Prefer private chat" or "Prefer public chat" wouldn't clear out the Chat textbox.

The use case was that I was in a game with a public chat, and as I was responding to a chat I decided partway that this was might be better as a private chat. And so I hit "Prefer private chat" which deleted all of the content I had typed in. Hitting the Back button did not bring back the chat contents, so I had to start again.

Thanks!

blackshadowshade:
Fixes #2296.

http://jenkins.buttonweavers.com:8080/job/buttonmen-blackshadowshade/1461/ (if it passes)

### DIFF
--- a/src/ui/js/Game.js
+++ b/src/ui/js/Game.js
@@ -64,10 +64,22 @@ Game.showLoggedInPage = function() {
 };
 
 // Redraw the page after a previous action succeeded: to do this,
-// clear all activity variables set by the previous invocation
+// clear all activity variables set by the previous invocation except for
+// items that are currently being actively cached
 Game.redrawGamePageSuccess = function() {
-  Game.activity = {};
+  Game.activity = Game.getActivityAfterApiSuccess();
+
   Game.showLoggedInPage();
+};
+
+Game.getActivityAfterApiSuccess = function() {
+  var activityAfterApiSuccess = {};
+
+  if (('cacheAcrossCall' in Game.activity) && Game.activity.cacheAcrossCall) {
+    activityAfterApiSuccess.chat = Game.activity.chat;
+  }
+
+  return activityAfterApiSuccess;
 };
 
 // Redraw the page after a previous action failed: to do this,
@@ -1718,6 +1730,10 @@ Game.formToggleChatVisibility = function(e) {
     'game': Game.game,
     'private': !(Api.game.player.isChatPrivate),
   };
+
+  // Store the game chat in recent activity
+  Game.activity.chat = $('#game_chat').val();
+  Game.activity.cacheAcrossCall = true;
 
   Api.apiFormPost(
     formargs,
@@ -3617,7 +3633,7 @@ Game.chatBox = function(hidden) {
     'maxlength': Game.GAME_CHAT_MAX_LENGTH,
   });
 
-  // Add previous chat contents from a rejected turn submission if any
+  // Add previous chat contents that have been actively cached
   if ('chat' in Game.activity) {
     chatarea.val(Game.activity.chat);
   }
--- a/test/src/ui/js/test_Game.js
+++ b/test/src/ui/js/test_Game.js
@@ -102,12 +102,57 @@ test("test_Game.redrawGamePageSuccess", function(assert) {
   $.ajaxSetup({ async: false });
   BMTestUtils.GameType = 'frasquito_wiseman_specifydice';
   Game.activity.chat = "Some chat text";
+  Game.activity.blip = "generic member test";
   Game.redrawGamePageSuccess();
   var item = document.getElementById('game_page');
   assert.equal(item.nodeName, "DIV",
         "#game_page is a div after redrawGamePageSuccess() is called");
   assert.deepEqual(Game.activity, {},
         "Game.activity is cleared by redrawGamePageSuccess()");
+  $.ajaxSetup({ async: true });
+});
+
+// Use stop()/start() because the AJAX-using operation needs to
+// finish before its results can be tested
+test("test_Game.redrawGamePageSuccess_cacheChat", function(assert) {
+  $.ajaxSetup({ async: false });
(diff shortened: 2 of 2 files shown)


src/ui/js/Game.js
```diff
@@ -64,9 +64,21 @@ Game.showLoggedInPage = function() {
 };
 
 // Redraw the page after a previous action succeeded: to do this,
-// clear all activity variables set by the previous invocation
+// clear all activity variables set by the previous invocation except for
+// chat that is currently being actively cached in Game.activity.chat
```
cgolubi1: At the risk of being picky: i think it would be tidier to make a function for this, like:
```
Game.activity = Game.getActivityAfterApiSuccess();
```

Basically i think futzing around with the chat variable is a little low-level for this function.

My less nitpicky worry is: is this redraw function *only* invoked after chat privacy is changed, or is it invoked after any API function is invoked?  If the latter, won't it retain chat on a successful attack in which you submit your chat?  Did you test that?

blackshadowshade: I tested after a successful attack, and I didn't notice any visible differences because:
- the chat box is not visible after successful attack, and
- if the chat is to be altered, the chat box that becomes visible is filled with the current value of chat anyway

I guess the question is whether we ever really want to clear the chat variable in Activity, considering all the cases in the past where we've really wanted to keep it.

cgolubi1: Hmm, what if your opponent has to pass?  Anyway, maybe i should just put up a dev site, since it sounds like we might get bogged down in parsing possible options by memory otherwise.  I'll do that when i get a sec.

blackshadowshade: Okay, good pick. I tested with the case where the opponent autopasses and yes, this is one case where we probably don't want to maintain previous chat.

So, based on the fact that there is some sense that the default should still be to clear the chat variable, how would you feel about me using a new Game.activity.chatCache variable to explicitly show that we want to carry the old chat across? Or would you prefer some other approach?

cgolubi1: Sorry, missed that this ended with a question for me.  As in the variable would be a boolean, and it would be false most of the time, but calling an API like change chat privacy would set it to true?  Seems reasonable to me --- if that's the approach, i would make the name more generic, like `Game.activity.cacheAcrossCall`, in case there's other activity it turns out we want to cache.

blackshadowshade: Okay, sounds good. I'll take another look at this soon.

blackshadowshade:
Okay, I've reworked this pull request to address the issues and I've also rebased it back into one commit.

Unfortunately, I still can't run a Jenkins run, so that will have to wait until Jenkins can see Github again.

cgolubi1:
I tried blindly upgrading the "git" and "git client" plugins in jenkins to see if that would help.  Not sure why i thought that was a good idea, but (spoiler alert) it wasn't.  This is going to take more doing; i'll try to block off some time for it tomorrow, and sorry for the bother.

src/ui/js/Game.js
```diff
@@ -1719,6 +1731,10 @@ Game.formToggleChatVisibility = function(e) {
     'private': !(Api.game.player.isChatPrivate),
   };
 
+  // Store the game chat in recent activity
+  Game.activity.chat = $('#game_chat').val();
+  Game.activity.cacheAcrossCall = true;
```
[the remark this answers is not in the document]
cgolubi1: Just to clarify, the thing i really object to is the new parameter `Game.activity.cacheAcrossCall` - it introduces the implication that it's doing something, i.e. that defining it at around the time you define the chat is somehow what *causes* the chat to be the thing that's cached.  But in actual point of fact, hardcoding `chat` as the parameter to be saved in `Game.getActivityAfterApiSuccess()` is what causes chat to be the thing that's cached.  I think introducing the new parameter will actually make this code harder for us to understand in the future, so i'd rather not unless i'm missing something it's doing.

cgolubi1:
Tagging incomplete; but remove tag if i'm misunderstanding stuff.

src/ui/js/Game.js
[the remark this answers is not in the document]
blackshadowshade: So, how do we handle the case where we don't want to cache the chat?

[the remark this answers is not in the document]
cgolubi1: Ahh, i see what you're doing, and now i understand what i want you to do differently.  `:>)`

So right now the code is:

```
  Api.apiFormPost(
    {
      type: 'submitTurn',
      game: Game.game,
      attackerIdx: Api.game.playerIdx,
      defenderIdx: Api.game.opponentIdx,
      dieSelectStatus: Game.activity.dieSelectStatus,
      attackType: Game.activity.attackType,
      turboVals: Game.activity.turboVals,
      chat: Game.activity.chat,
      roundNumber: Api.game.roundNumber,
      timestamp: Api.game.timestamp,
    },
    { 'ok': { 'type': 'server', }, 'notok': { 'type': 'server', }, },
    '#game_action_button',
    Game.redrawGamePageSuccess,
    Game.redrawGamePageFailure
  );
```

The only difference between `Game.redrawGamePageSuccess()` and `Game.redrawGamePageFailure()` is that `Game.redrawGamePageSuccess()` caches chat.  So you could make a new function, `Game.redrawGamePageCachingChat()` that hangs onto the chat.  (But, in fact, you probably don't want to just hang onto the chat.  You want to hang onto *all* activity, right?  Like, if someone selected some dice before toggling chat, you don't want to unselect them.  So you really want exactly the behavior of `Game.redrawGamePageFailure()` in the success case too.  So you could simply call that function in both cases, or if you don't like calling a function called Failure in the success case on comprehensibility grounds, you could make a second function that does the same thing.

blackshadowshade:
Okay, closing this one in favour of #2396.

PR #2326, opened 2018-01-30
Add Rush support to RandomAI

cgolubi1:
* Used for testing #2210
* Should be a noop until Rush is merged (and random_ai doesn't have unit tests)


blackshadowshade:
Okay, looks plausible. Let's give this a go.

Issue #440, opened 2014-01-15
Upgrade the version of JSHint that we are using


blackshadowshade:
Our Javascript linter, JSHint, is currently at version 1.1.0. The current version of JSHint has changed to 2.4.X, and there are major changes, as indicated by the change of the major version number.

It would be good to update this sometime.

Issue #439, opened 2014-01-15
Upgrade the version of JQuery that we are using


blackshadowshade:
Since we started working with JQuery, the version numbers have changed quite considerably. It would probably be good to update, if only to accommodate new browser versions.

PR #2337, opened 2018-03-12
Add an initial v2.0 circleci config

cgolubi1:
* CircleCI: https://circleci.com/gh/cgolubi1/buttonmen/60

I think this is a working CircleCI configuration that has enough of the core functionality that we can adopt it and move forward, unblocking development work.  The only arguably important thing that's missing is JS code coverage, which i'd like to have, but which will take some thinking to get working on a new site, and it doesn't strike me as worth holding up the world over.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,69 @@
+version: 2
+jobs:
+  build:
+    machine:
+      image: circleci/classic:latest
+    steps:
+      - checkout
+      - run:
+          name: Symlink /buttonmen
+          command: sudo ln -s /home/circleci/project /buttonmen
+      - run:
+          name: Make sure log dir exists
+          command: mkdir -p build/logs
+      - run:
+          name: Bootstrap puppet
+          command: sudo bash deploy/vagrant/bootstrap.sh
+      - run:
+          name: Run puppet
+          command: sudo puppet apply --modulepath=/buttonmen/deploy/vagrant/modules /buttonmen/deploy/circleci/manifests/init.pp
+      - run:
+          name: Audit line termination of all files in the tree to match Button Men standard
+          command: python deploy/jenkins/audit_newlines.py
+      - run:
+          name: Audit layout of .php files in the tree to match Button Men standard
+          command: php ./deploy/jenkins/audit_php_files.php .
+      - run:
+          name: Audit test coverage of .js files in the tree to match Button Men standard
+          command: python ./deploy/jenkins/audit_js_unit_test_coverage.py
+      - run:
+          name: Run grunt JS lint tests
+          command: sh ./util/grunt/jenkins-grunt.sh
+      - run:
+          name: Generate checkstyle.xml using PHP_CodeSniffer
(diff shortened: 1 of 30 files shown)


cgolubi1:
If we like this in general, i'll write some documentation about how individual contributors should get themselves setup with CircleCI --- it's quite easy if you have a github account.

blackshadowshade:
I've taken a look at the CircleCI output, and just based on the duration of 15 minutes, I think that this is looking very positive. The output looks quite different, and it'll probably be useful to set up the Test Summary section once everything is running right. However, the PHPUnit output is familiar, so I had no trouble with that. Also positive is the possibility of running multiple runs in parallel.

I'm looking forward to seeing JSHint and QUnit running.

One question based on something I've been thinking about: How much work will it be to change the script to work with PHP 7.2 and the newest PHPUnit (and theoretically under the equivalent of Ubuntu 16)? I ask because I'm looking at getting the site up and running under PHP 7 with an updated version of Ubuntu, and I'm making good progress.

cgolubi1:
Changing versions will be trivial with this setup --- since we're building the site from scratch and using our vagrant config to install stuff, we just install different stuff, push to the branch, and see if it passes.

cgolubi1:
Also worth noting: i didn't configure logging yet, but i did configure artifact upload, so you can poke around in the "Artifacts" tab to find out why your build failed, and the autogenerated PHPdoc shows up there too.

blackshadowshade:
Huh. I thought we'd gotten our code down to no PHPDoc errors. I presume that the settings must be different to what we had under Jenkins.

blackshadowshade:
I've taken a closer look at those PHPDoc errors, and I don't exactly understand what's going on here. Perhaps the syntax has changed.

For example, there's one error in engine/BMGameState.php that says

    Argument $gameState is missing from the Docblock of as_string

but the codebase currently has

    /**
     * Convert numerical game state into a string
     *
     * @param int $gameState
     * @return string
     */
    public static function as_string($gameState) {

Anyway, for the moment, this is a minor issue, and it's far more important that we get the JSHint and QUnit tests running.

blackshadowshade:
I'm going to link #439 and #440 here, because I think this CircleCI concept is a simple way of solving both these issues.

cgolubi1:
Okay, indeed, i downgraded to phpdoc-2.7.0 in puppet, and now phpdoc is saying much more sane things: https://circleci.com/gh/cgolubi1/buttonmen/54

(I know that's not the most important problem, but it's the one i could solve in 10 minutes.  :>)

blackshadowshade:
Hey, progress is progress!  ;)

cgolubi1:
Per the updated description, assuming this passes, i think we should call it good enough to unblock the world.

Shadowshade, if you like this, merge it, and i'll update our developer docs to explain how to get setup on CircleCI and you can test them.

cgolubi1:
I updated https://github.com/buttonmen-dev/buttonmen/wiki/Developer-guide:-getting-started

blackshadowshade:
Okay, this looks fine as a starting point, and I agree, let's get things unblocked. We can tidy up the many small things later.

Issue #1167, opened 2014-08-07
Upgrade to PHP 5.5 or higher


AdmiralJota:
We're apparently currently using PHP 5.3.10. We should upgrade to a newer version at some point. (This would particularly help with #884.)

blackshadowshade:
I'm choosing this as my BotM for April.

blackshadowshade:
Looking at https://github.com/sebastianbergmann/phpunit/wiki/Release-Announcement-for-PHPUnit-4.7.0, we should probably consider upgrading to PHP 5.6 or higher when we actually do this upgrade.

blackshadowshade:
Okay, adding my September BotM to this.

blackshadowshade:
Just noting for myself that the solution to #1728 that was implemented currently installs PHPUnit 4.8.26, according to Chaos, which is great.

blackshadowshade:
So, on this theme, I've been using with PHP 5.6.30-12 on stats.dev, and have hit the following issues that will need to be addressed when we upgrade our version of Ubuntu (as noted in #2208):

1. /etc/mime.types needs to be checked to see whether the PHP types there have been commented out.
2. PHP-MySQL needs to be upgraded
    
    sudo apt-get install php5.6-mysql

3. The .conf files for Apache need to be updated, because they have been rearranged:

    sudo cp /etc/apache/conf.d/buttonmen /etc/apache/conf.enabled/buttonmen.conf

4. mbstring needs to be installed

    sudo apt-get install php5.6-mbstring

5. The call to crypt() needs to be changed to

    ':password' => password_hash($password, PASSWORD_DEFAULT),

This list is not complete yet, since it's still not possible to create new users, and I'll update the list when I track down the last few issues.

blackshadowshade:
Just a thought that we should actually be aiming for PHP 7 when we upgrade, not PHP 5.6.

A migration guide from PHP 5.6 to PHP 7.0 is here: http://php.net/manual/en/migration70.php

blackshadowshade:
I've updated my stats.dev instance to PHP 7.2, following the instructions here: https://askubuntu.com/questions/109404/how-do-i-install-different-upgrade-or-downgrade-php-version-in-still-supported

I'll be testing PHP 7.2 in parallel on the stats.dev server and my new MAMP 4.0 distribution.

blackshadowshade:
One thing that has changed with PHP 7 is the way it handles indirect access to variables.

[...]

PR #2338, opened 2018-03-23
Added logic to migrate up to PHP 7

### LINKED ISSUE #884: Upgrade from crypt() to password_hash()?
The documentation for the PHP crypt() function (http://docs.php.net/manual/en/function.crypt.php) suggests that password_hash() is preferable to crypt() and is compatible with existing hashes. It looks like upgrading to it might be as simple as just calling that function instead of crypt() and increasing player.password_hashed to 255 characters. We should consider if we want to switch over.

blackshadowshade:
Addresses the codebase part of #1167. Fixes #884.

The changes here fall into two categories:

1. Indirect expressions are handled differently in PHP 7.

See http://php.net/manual/en/migration70.incompatible.php for details.

2. New functions for password hashing and verification were added in PHP 5.5, and the use of a manual salt was deprecated in PHP 7.0.

See http://php.net/manual/en/function.password-hash.php for details.

I've added code here to convert short hashes (length 34) to long hashes (length 60) on successful login. I've also added a database update to increase the hash field sizes to 255, on the basis of advice from the PHP doc page for password_hash.

I've tested the following with my MAMP stack under PHP 7.1.13: logging in, logging out, creating a new user, and validating a new user. Game loading and play looks unchanged.

### DIFF
--- a/deploy/database/schema.player.sql
+++ b/deploy/database/schema.player.sql
@@ -6,7 +6,7 @@ CREATE TABLE player_status (
 CREATE TABLE player (
     id                  SMALLINT UNSIGNED AUTO_INCREMENT PRIMARY KEY,
     name_ingame         VARCHAR(25) UNIQUE NOT NULL,
-    password_hashed     VARCHAR(128),
+    password_hashed     VARCHAR(255),
     name_irl            VARCHAR(40) NOT NULL,
     email               VARCHAR(254),
     is_email_public     BOOLEAN DEFAULT 0 NOT NULL,
@@ -51,7 +51,7 @@ CREATE TABLE player (
 CREATE TABLE player_auth (
     id         INT UNSIGNED AUTO_INCREMENT PRIMARY KEY,
     player_id  SMALLINT UNSIGNED,
-    auth_key   VARCHAR(253) UNIQUE NOT NULL,
+    auth_key   VARCHAR(255) UNIQUE NOT NULL,
     login_time TIMESTAMP DEFAULT CURRENT_TIMESTAMP
 );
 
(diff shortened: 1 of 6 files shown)


blackshadowshade:
So, apart from the file conflict here, I'll need help customising the CircleCI config to run with PHP 7.2, and using the Xenial Xerus release of Ubuntu.

cgolubi1:
Otherwise, i don't have much of a problem with this.  It seems like a good use case for standing up a dev site, so that we can see if there are any issues with existing users/logins under PHP 5.

cgolubi1:
irilyth: can you sync up fleeting = 54.91.8.63?

irilyth:
> irilyth: can you sync up fleeting = 54.91.8.63?

Done!

blackshadowshade:
Yes, sure, there's a way to test the conversion of hashes. The obvious way is to deploy a site running PHP 7.* with an old database (as we currently do), and then have a number of testers try to log in. They should all be able to log in successfully, which is the first part of the test. The second part of the test is to compare the hashes before and after this login step, and check that they've changed from short to long hashes.

The third part of the test is for a new user to create a new account. They should also be able to login successfully, and the hash in the database should be a long hash, both before and after login.

cgolubi1:
Okay, well, at least i logged in and logged out on the new site, created and verified a new user, and checked the DB to see that the player and player_auth table entries still seem to have a similar hash format.

If other players can check out fleeting.dev.buttonweavers.com and see if we've clearly broken anything related to logins or auth, that'd be neat.  If this isn't harmful, i'm inclined to merge it.

irilyth:
I was able to log in and out of fleeting.dev.buttonweavers.com as my various testing users, and it all seemed to work fine.

cgolubi1:
Okay, i think if this stuff works, it's pretty harmless, and if it doesn't work, we'll find out and deal with it on staging testing.  So i'm going to suggest we merge it and deal with PHP 7.2 testing under a different pull unless you (Shadowshade) don't think that's a good idea.

blackshadowshade:
Agreed. Let's merge this, then deal with potential PHP 7.2 problems when we have a PHP 7.2 test site.

Issue #2347, opened 2018-04-17
Fix TL status of buttons


blackshadowshade:
In http://www.buttonweavers.com/ui/forum.html#!threadId=42&postId=12420, it was noted that ConMan is not TL.

http://www.zatrolene-hry.cz/dialog/download-file?id_soubor=871 (which has been cached in https://github.com/buttonmen-dev/buttonmen/blob/master/notes/database/button-men-pravidla-cz.pdf) has a list of TL status for many of the buttons. The TL status on the site should be checked against this.

blackshadowshade:
Differences according to this list:

Max is not TL
Bridge and Tunnel buttons are not TL (Agent Orange, Huitzilopochtli, Lisa, Nethershadow, Phuong, Uncle Scratchy)
Button Brains buttons are not TL
Apples and Green Apple are not TL
Diceland buttons are not TL
Dead Guy is TL !!!!!
Carson is not TL
Most of the Metamorphers are not TL (Daisy, Jingjing, Spencer, Talus)
Count Repugsive is TL
Angora is not TL
Amara Wintersword is not TL
Captain Bingo is not TL
Oni is not TL
Spite is not TL
Super Germ is not TL
Dr. Speculo is not TL
Harry Pudding is not TL
MothMan is not TL
Assailer is not TL
Harbinger is not TL
ConMan is not TL
Pikathulhu is not TL
Ulthar is not TL

blackshadowshade:
One obvious question is whether this list is up to date or not. It seems to be based on www.randomdice.com, and it has the www.buttonmen.dhs.org site referenced, so it's quite possible that it accurately reflects what was on the old site.

blackshadowshade:
Naturally, any changes here that agree with the official site https://beatpeopleup.cheapass.com/tournament-list/ should definitely be adopted. That means, for example, that Angora will definitely end up being not TL.

irilyth:
As a random other question: Where did the TL status for buttons on the new site come from in the first place? How did we get so many of them wrong, or at least "at odds with this other list"?

blackshadowshade:
I'm not certain that we can answer that definitively, considering that Fog was the one who did most of the work on compiling the information on the buttons.

irilyth:
Cassidy Werner <cheapasscassidy@gmail.com> confirms that ConMan is not TL.

[...]

PR #2348, opened 2018-04-17
Added file with TL status of buttons

blackshadowshade:
In this pull request, I'm simply adding a PDF file to the notes folder. The file has a table which has the TL status of buttons.

I'm caching this file so that we have something to refer to for #2347 if the original file gets taken down.


cgolubi1:
I agree!  That is a PDF which has been added to the notes folder.

PR #2353, opened 2018-04-23
Fixed bug with skill attack containing more than two zero-sided dice

### LINKED ISSUE #2350: skill sum bug triggered by Dead Guy
When Dead Guy attempts to make a skill attack including more than two 0-sided dice, the attack fails with 'Attacking die values do not sum up to target die value.'

blackshadowshade:
Fixes #2350.

### DIFF
--- a/src/engine/BMAttackSkill.php
+++ b/src/engine/BMAttackSkill.php
@@ -55,6 +55,8 @@ class BMAttackSkill extends BMAttack {
      */
     protected $hitTable = NULL;
 
+    const MAX_ZEROS = 2;
+
     /**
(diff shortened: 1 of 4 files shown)


blackshadowshade:
I've fixed the bug. I've also made sure that the zeros that are being stripped originate from zero-sided dice that do not have any skills.

cgolubi1:
I would love to merge this, because the bug is basically only hurting me.  So let's get a replay site running for it.

cgolubi1:
This has replayed 8300 games and the same number of novel games without exception.  Let's merge it.

PR #2355, opened 2018-04-28
Fixed TL status

### LINKED ISSUE #2347: Fix TL status of buttons

blackshadowshade:
Fixes #2347.


cgolubi1:
Database change, so let me do an update-vs-rebuild test on this.

cgolubi1:
Looks fine.  So people who want to have opinions about this can see all the changes in https://github.com/blackshadowshade/buttonmen/blob/93a544c71a96893d8889c4dad38fa699d9c23456/deploy/database/updates/02347_tl_status_01.sql

I'm not super stressed about any of the TL downgrades, but i guess i don't see why it's advantageous for us for Dead Guy to be TL.  Nobody likes Dead Guy, and if we ever implement "random TL buttons", people will be annoyed if Dead Guy shows up in those games.  My vote is that we are the captains of our souls and all that, and we can reject that change if we don't like it.  (And i don't like it.)

Disagreements?  Do other folks think either (a) that it is objectively advantageous for Dead Guy to be a TL button, or (b) that we are obligated to make Dead Guy TL even if it's not advantageous for us?

irilyth:
My inclination would be to keep it TL, but also keep the thing that excludes Dead Guy from random versus random matches in general.

cgolubi1:
Note that right now Dead Guy isn't TL (on our site) - this PR modifies it to make it TL.

But that's not really important, obviously, cause we want to wind up with the correct state.  Can you say more words about why your inclination is to make it TL?  My inclination is that TL buttons should be fairly balanced and cheese-averse, because that's what TL players are looking for.  I don't see a big difference between the use case of "give me a random TL button" and the use case of "i want to see my stats among games with TL buttons".

I think it makes sense to defer to other sources of information about what buttons should be TL in general, but i also think we can override with common sense.  And i think common sense says that Dead Guy is not a button where there's probably been a lot of data from in-person games (like, people are not playing serious BM games with one player playing Dead Guy), and that Dead Guy is transparently cheesy.  So IMO that goes against the spirit of TL and we should use our judgment and mark it not-TL.

I will note that i find my argument above very compelling, but it's not *that* important to me.  So if other folks disagree, i'll go along with it.  I just want to at least discuss why briefly.

irilyth:
I think that TL is an official Cheapass Games designation, and that we should synch with it.

I think we should use the hypothetical "balanced" label, or the ability to limit random vs random games by stats, or something like that, to get the "give me a fun random game" feature.

AdmiralJota:
Yeah, I agree with irilyth here. It's not up to us to decide whether or not a button should be TL. "Tournament Legal" doesn't mean "the buttonweavers team thinks this button is fairly balanced and cheese-averse"; It means "Cheapass Games has decided that this button can be legally used in official Cheapass-sponsored tournaments".

It *is* up to us to decide what we *do* with that information, though: we're not obligated to even track TL status on the site if we don't want to. I would be perfectly happy to have a separate "balanced" tag that's entirely independent of whether or not a button is TL, and to base most of our random button selection logic off of that rather than the TL status.

cgolubi1:
Mm, fair enough.  If we think the list mentioned in #2347 is canonical, then let's roll with it.

blackshadowshade:
Sounds like I should put a little bit of work into getting a 'balanced' tag working with our random game generation, at least as a demonstrator.

cgolubi1:
Okay, i finally compared your list here to your list on #2347, and i don't think you changed the TL status of the Button Brains buttons?

blackshadowshade:
Okay, I've updated the Button Brains TL status.

cgolubi1:
LGTM.  Rebase to one commit and i'll merge it.

blackshadowshade:
Rebased.

blackshadowshade:
Well, CircleCI failed spuriously because it couldn't find Overview.
(see https://circleci.com/gh/blackshadowshade/buttonmen/74)

I've set it to build again. Let's hope that fixes it.

blackshadowshade:
And another spurious failure because CircleCI can't find OpenGames.
(see https://circleci.com/gh/blackshadowshade/buttonmen/78)

cgolubi1:
Yeah, i got a spurious OpenGames failure myself today.  More debugging needed.  :>/  Anyway, let's merge this.

PR #2359, opened 2018-05-05
Added Jota's page that converts XLS-pasted tables into forum tables

### LINKED ISSUE #2356: Allow tables in forum posts
Our users often try to use tables in their forum posts, and we should make it possible for them to do this in a way that actually looks like a table.

Blargh also mentioned that it would be nice to be able to cut and paste the table. For that, we'd need to know what sort of formats to support.

blackshadowshade:
Fixes #2356.

So, I'm just adding Jota's HTML/JQuery page that converts tables from Excel into a [code] block that can be posted into our forums.

I figure that other external tools (like Bagel's button filter) might also fit into the /external folder.

### DIFF
--- a/src/ui/tools/table_converter.html
+++ b/src/ui/tools/table_converter.html
@@ -1,0 +1,60 @@
+<html>
+  <head>
+    <script src="https://ajax.googleapis.com/ajax/libs/jquery/3.3.1/jquery.min.js"></script>
+    <script>
+      function process()
+      {
+        var lengths = [];
+
+        var lines = $('#input').val().split('\n');
+        for (var i = 0; i < lines.length; i++)
+        {
+          var fields = lines[i].split('\t');
+
+          for (var j = 0; j < fields.length; j++)
+          {
+  		    if (lengths[j] == null || lengths[j] < fields[j].length)
+		    {
+		      lengths[j] = fields[j].length;
+            }
+          }
+        }
+
+        var out = "[code]";
+		for (i = 0; i < lines.length; i++)
+		{
+          if (i != 0)
+          {
+            out += '\n';
+          }
+
+		  var fields = lines[i].split('\t');
+
+          for (var j = 0; j < fields.length; j++)
+          {
+            out += fields[j];
+
+            var padding = lengths[j] - fields[j].length + 2;
+            while (padding > 0)
+            {
+              out += ' ';
+              padding--;
+            }
+          }
+		}
+		out += "[/code]";
+
+		$('#output').text(out);
+      }
+    </script>
+  </head>
+  <body>
+    <h2>Convert MS Excel tables into forum tables</h2>
+    <div>Instructions: Copy table from MS Excel, paste it into the box, then click Go.</div>
+    <br>
+    <textarea id="input"></textarea>
+    <input type="button" id="go" onclick="process();" value="Go"></input>
+    <br><br>
+    <textarea id="output" style="border: 1px solid black; padding: 5px;margin: 5px;font-family: monospace; white-space: pre;"></textarea>
+  </body>
+</html>


irilyth:
Does this actually take a .xls file as input? If not, it might make sense to name it based on what it does actually take as input. (At a glance, it looks like maybe it's tab-delimited text?)

(It also doesn't matter a whole lot what it's called, so if no one else cares much, feel free to ignore me.)

AdmiralJota:
Yeah, it takes line/tab delimited text, since that's what you get when you copy a group of cells from Excel and paste it into a textbox on a web page.

irilyth:
Ah, it's for copy and paste from a spreadsheet screen, not for uploading a file or whatever; that makes sense.

If we wanted to give it an accurately descriptive name, we could call it table-converter-tab-to-spaces.html or something. Or just table-converter.html, maybe we'll update it in the future to convert between multiple table formats. :^ )

blackshadowshade:
So, two things:
- I'm happy to change the name to table_converter.html
- I think it would be useful to have a title and instructions on the HTML sheet itself. Anyone feel like doing some wordsmithing here?

cgolubi1:
Let's stand up a dev site and see if this is loadable, shall we?

Meanwhile, i want to take a brief moment to discuss naming, since we haven't had this kind of content on the website before.  You opted for /external/, presumably to indicate that it's not part of the main body of stuff.

Can i countersuggest /tools/, for consistency with the top-level branch where we're keeping the api client, stats stuff, etc?

blackshadowshade:
I considered 'tools' and wondered if it would be confusing in the future, since we would then have ambiguity when referring to the tools folder. However, if you'd prefer 'tools', I'm okay with that.

cgolubi1:
My theory is that it's one thing (user-contributed tools) some of which are displayed on the website and some of which aren't.  We don't get confused between tools/ and test/tools/, after all.

blackshadowshade:
Okay, changes made. The dev site for this can be updated.

cgolubi1:
Created a dev site --- irilyth, when you have a chance can you sync up: transient 54.167.251.230

irilyth:
```transient.dev.buttonweavers.com has address 54.167.251.230```
Done!

cgolubi1:
So one note is that this would be our first JS that people who aren't logged into the site can use (other than the "new player" and "login" code, of course).  I think that's fine because JS runs browser-side, but does anyone have qualms we should think about?

blackshadowshade:
I have no problems with this particular case.

I think that if any future tool hits the database, we should think extra carefully about it.

cgolubi1:
> I think that if any future tool hits the database, we should think extra carefully about it.

Disagree - i think that if any future tool hits the database, we should disallow unauthenticated access to it with no thought whatsoever.  :>)

But, indeed, i'm not going to veto this.  Have folks checked out the actual functionality?  Does it do what we want?

cgolubi1:
It's at http://transient.dev.buttonweavers.com/ui/tools/table_converter.html for those who want to try it out but don't want to hunt for it.

cgolubi1:
Still waiting for someone except the submitter to weigh in on this solution.  Since it's not part of the main site, the bar is pretty low, but i'd like to at least hear from Jota that the online converter does what it was designed to do.

If no one says anything in a couple more days, i'll go ahead and merge this since it is low-risk, but i haven't been able to actually test making it generate a table (i don't have an excel table handy and the input format isn't documented), so i can't +1 that it works.

AdmiralJota:
It looks like it works in some browsers but not others. That is, it produces sensible output in any case, but copying and pasting the output may or may not preserve whitespace correctly depending on what browser/version you're using.

We might be able to fix that by changing the output area from a div to a textarea. Should we do that?

blackshadowshade:
I'm happy to try that.

blackshadowshade:
The div has been changed to a textarea. @cgolubi1, if you wouldn't mind updating the test site, please.

blackshadowshade:
Oh, because I rebased before, I've rebased this again into one commit. You'll need to do a fresh deploy.

cgolubi1:
Okay, no worries.  If you happen to remember not to rebase during dev-site testing, that's great.  Rebuilding now.

cgolubi1:
Actually, i'm going to cheat on this to save the trouble of repointing the DNS, and just copy the new HTML file into place.

cgolubi1:
Okay, should be updated.  Check it out.

cgolubi1:
Next step: AdmiralJota should report on whether http://transient.dev.buttonweavers.com/ui/tools/table_converter.html looks good to merge now.

AdmiralJota:
The </div> after <textarea> should be </textarea>.

Other than that, it seems to work fine.

cgolubi1:
Aight.  I think we're good to go here.  Thanks, all.

PR #2363, opened 2018-05-13
Don't verify SSL cert when downloading phpunit

### LINKED ISSUE #2362: replay/dev sites can't download phpunit because they don't like the cert issuer
On new replay sites (and dev sites, but it only really matters for replay sites), vagrant provision is failing with:

```
err: /Stage[main]/Php::Base::Feature::Phpunit/Exec[php_wget_phpunit]/returns: change from notrun to 0 failed: /usr/bin/wget --no-verbose -O /etc/php5/deploy-includes/phpunit.phar https://phar.phpunit.de/phpunit-old.phar returned 5 instead of one of [0] at /tmp/vagrant-puppet-1/modules-0/php/manifests/init.pp:30
```

and a zero-length phpunit.phar is created.

cgolubi1:
* Fixes #2362 
* Testing: applied the change when standing up a replay site and verified that puppet succeeded.

### DIFF
--- a/deploy/vagrant/modules/php/manifests/init.pp
+++ b/deploy/vagrant/modules/php/manifests/init.pp
@@ -24,7 +24,7 @@ class php::base {
 class php::base::feature::phpunit {
   exec {
     "php_wget_phpunit":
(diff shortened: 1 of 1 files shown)


blackshadowshade:
Let's do this.

Issue #2342, opened 2018-04-07
Tweak our CircleCI setup to get it "just so"


blackshadowshade:
So, this is the issue where we put our wishlist of tweaks and changes to the CircleCI setup.

blackshadowshade:
On a PHPCS failure, the email from CircleCI suggests "For more detailed failure reports, use JUnit XML formatting." and has a link to https://circleci.com/docs/1.0/test-metadata/

If we were able to specify this format for the PHPCS output, then I'm guessing that the failure message would show the contents of the XML file, meaning that I don't need to go digging through Artifacts to find the errors.

cgolubi1:
One thing you should keep an eye out for is that we have a CircleCI 2.0 config.  They seem to be mid-migration from 1.0 to 2.0, and i didn't see any reason not to pick the newer one, which seemed like it had a more flexible config language.  But as a rule that means if you find a 1.0 doc for how to do something, you probably need to look for a 2.0 doc to do the same thing, because they totally rewrote the config language.

blackshadowshade:
It would be useful to be able to follow someone else's builds in my control panel.

blackshadowshade:
Our QUnit runs seem to be rather inconsistent, sometimes failing because QUnit cannot seem to find all of the source during one run, and then passing on the next run when it does manage to find all the source.

cgolubi1:
So... i'm really suspicious: the files QUnit needs are being rsynced immediately before the QUnit tests run, and as you say, the failure mode looks consistent with the files not being there yet.  Obviously, rsync is exiting successfully.  But i wonder if the files are being copied to some kind of eventually consistent data store.  In which case the fix would be to verify that the files exist before succeeding the step.  Let me experiment with that.

blackshadowshade:
I've rebased #2311, #2314, and #2355. The first two have explicit UI changes, while the last does not. #2314 has never passed testing before, the others have (but only intermittently). We'll see whether these all pass now.

PR #2364, opened 2018-05-13
Add a CircleCI step which tries to verify that the dummy API files have landed before running QUnit

cgolubi1:
* Partially addresses #2342 
* I can't prove this works or that i've diagnosed the problem correctly --- my test CircleCI run passed on the first try, and we know we're getting spurious QUnit failures pretty often in CircleCI (i'd say at least 1/5 of the time?), so if we merge this and don't see more spurious failures, or if we merge this and ever see it need to sleep once before continuing, we should feel very confident in it.  And if we merge this and do see another spurious failure, we can go back to the drawing board.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,6 +39,9 @@ jobs:
           name: Install dummy API files created by PHPUnit into test webserver
           command: sudo rsync -av src/api/dummy_data/ /var/www/api/dummy_data/
       - run:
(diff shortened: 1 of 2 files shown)


blackshadowshade:
I'm happy to try this. I'll rebase my pull requests (when I have time) and then I'll rerun them all, including the one I haven't been able to get to pass yet. We'll see if this makes a difference.

PR #2367, opened 2018-06-06
Add Oregon to responderTest's skill descriptions

cgolubi1:
* Assuming it passes, should fix a couple of replay test errors i found in my default replay loop


blackshadowshade:
Looks good to me. Merging.

PR #2373, opened 2018-06-27
Enable btn_special for Oregon in database

### LINKED ISSUE #2333: Oregon has the unique die skill
In #2279, we implemented the unique die skill for Gordo and Guillermo.

Oregon also has the unique die skill, so it should be trivial to enable this button by giving it the same button-level skill.

blackshadowshade:
Fixes #2333 better.


cgolubi1:
Sigh.  Well, i didn't catch this either, so here we are.  Let's merge it and deploy it.

PR #2381, opened 2018-07-17
Get phpunit from a version stashed on a local webserver to work around changes in phar.phpunit.de's SSL accessibility

cgolubi1:
* Fixes download of phpunit.phar failing yet again

This isn't a great solution, but the version of phpunit.phar we're downloading is over a year old anyway, so we're not really missing updates, and i'd rather pick a workaround that keeps us moving for now and focus on upgrading php / ubuntu, than sink a lot of time into debugging why the handshake is unhappy.

### DIFF
--- a/deploy/vagrant/modules/php/manifests/init.pp
+++ b/deploy/vagrant/modules/php/manifests/init.pp
@@ -23,8 +23,9 @@ class php::base {
 
 class php::base::feature::phpunit {
   exec {
+    # TODO: once we upgrade PHP, start getting the latest version of phpunit again
     "php_wget_phpunit":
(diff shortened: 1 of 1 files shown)


cgolubi1:
FWIW, it seems very likely that the answer is our ubuntu version, because this hasn't been breaking on circleci, but only when i try to stand up new sites in either virtualbox or AWS (the latter being why i haven't stood up the peloton dev site yet, because my script keeps failing out on not being able to get a successful vagrant run).

blackshadowshade:
In that case, I'll link this to #1167 so it doesn't get forgotten.

I'm happy to try this out once CircleCI passes.

blackshadowshade:
Okay, merging.

cgolubi1:
Thanks.  That should unblock me from working on a bunch of other stuff, including the peloton site.

PR #2383, opened 2018-07-21
Split responderTest into multiple files for convenience of editing

### LINKED ISSUE #2193: responderTest is starting to get large enough to cause editor trouble
NetBeans is starting to complain that responderTest is now more than 1M in size. Considering that it instruments each line so it needs in the order of 10 to 100 times more memory, we probably don't want responderTest to get much larger.

Consider splitting responderTest into two files.

cgolubi1:
* Fixes #2193 (if it passes)
* I think 20 games per file is a sane compromise.  One game per file is also a possibility, but it'd take more work to deal with the tests that depend on other tests.  Potentially doable, though, so if that seems much better, let me know.


cgolubi1:
Hmm, i'm retrying the build just in case, but it seems like there's something else i'm going to need to fix here to get the data where it needs to be for QUnit.  So: this isn't ready to merge, but feel free to have an opinion on whether this approach, once debugged, would solve your problem.

cgolubi1:
Nope, i guess it was the build artifact after all.  Well, cool.  Take a look.

blackshadowshade:
That looks super! CircleCI is running all the responder tests, starting at responder00Test. The depends clauses that depend on responder00Test look like they work fine.

Regarding the intermittent CircleCI failures, I've noticed that there an option on CircleCI to "Rebuild without cache". I wonder whether that is the one we should be using by default.

Anyway, accepting this pull request. Thanks for getting onto it so quickly!

PR #2385, opened 2018-07-26
Migrate action log entries of type determine_initiative to dedicated tables

cgolubi1:
* Also migrate very old "choose_swing" entries to "choose_die_values" in the database
* Partially addresses #1991 

A couple of relevant notes, if this passes:
* The "choose_swing" script is totally unrelated to everything else in the pull, and is a very simple modification of the choose_die_values script.  It only affects very old games, and should be a no-brainer and not distracting from the rest of the pull, but if you find that it is, complain, and i'll remove it.
* "determine_initiative" is the first action log entry type that iterates over dice, so there's a model for that here, where i retain only the information about each die which is actually used to construct the action log entry.  Obviously in the future when i'm creating "attack" type action log entries, what i retain will be much more info, but you can squint and see the model that's going to be used.

### DIFF
--- a/deploy/database/schema.game.sql
+++ b/deploy/database/schema.game.sql
@@ -166,6 +166,32 @@ CREATE TABLE game_action_log_type_choose_die_values_option (
     INDEX (action_log_id)
 );
 
+CREATE TABLE game_action_log_type_determine_initiative (
+    id                 INTEGER UNSIGNED AUTO_INCREMENT PRIMARY KEY,
+    action_log_id      INTEGER UNSIGNED NOT NULL,
+    winner_id          SMALLINT UNSIGNED NOT NULL,
+    round_number       TINYINT UNSIGNED NOT NULL,
+    is_tie             BOOLEAN DEFAULT FALSE,
+    INDEX (action_log_id)
+);
+
+CREATE TABLE game_action_log_type_determine_initiative_slow_button (
+    id                 INTEGER UNSIGNED AUTO_INCREMENT PRIMARY KEY,
+    action_log_id      INTEGER UNSIGNED NOT NULL,
+    player_id          SMALLINT UNSIGNED NOT NULL,
+    INDEX (action_log_id)
+);
+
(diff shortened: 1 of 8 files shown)


blackshadowshade:
I've seen that this pull request exists, currently sitting in Changi Airport, should be back in Melbourne in under 12 hours.

Expect some further comment from me within three days. If you don't get anything by the weekend, remind me again, please.  :)

cgolubi1:
Sweet.  As long as it's on your radar, i'm happy.

blackshadowshade:
Right, I've taken a look at this, and on the surface, it looks fine.

I'd like this tested a bit before I accept it though, since it's got quite a few bits to it.

Would it be possible to deploy this to a dev site (including the database updates you have here), preferably with a relatively recent copy of the database, perhaps one from one or two months ago? We should be able to have a look at previous completed games to see how they look, as well as new games that the testers can play for us.

cgolubi1:
The easiest thing for me to do is to stand this up with a recent copy of the staging database, since that's what dev sites do by default, and i think that should meet your request.  Doing that.

blackshadowshade:
A recent copy of staging is fine. Basically, I'd like there to be games that are affected by the database update, as well as games that have already completed.

cgolubi1:
Great.  It'll be up in the next few days; i'll let you know.

cgolubi1:
irilyth: can you setup 54.225.8.189 as temporary?

cgolubi1:
shadowshade: dev site is up, and i've run the two database update scripts on it (choose_swing didn't have any effect since those games only exist in the prod database, but i assert that i tested them under virtualbox).  So any existing (started) game on that dev site should demonstrate the determine_initiative change.

irilyth:
> irilyth: can you setup 54.225.8.189 as temporary?

Done!

blackshadowshade:
Okay, looks good to me. I've checked existing games, set a swing value midgame, and set a turbo swing die.

I'm accepting this pull request.

PR #2388, opened 2018-08-11
Added missing PHPDoc comments

blackshadowshade:
Just resolving some small missing comments noted in the PHPDoc output.


cgolubi1:
Yep, looks like some phpdoc comments.

Issue #2399, opened 2018-09-14
tear down jenkins site


cgolubi1:
We are no longer using jenkins.buttonweavers.com.  Tear it down.

Acceptance criteria:
* Tear down the instance itself and any associated EBS volumes
* Remove all puppet code from the repo which is only used by the jenkins site

cgolubi1:
I'd like to get rid of the jenkins instance and its volumes at this time.  Any objections?  Shadowshade: if it's fine with you, i probably don't need anyone else's buy-in.  (I know your thumbs-upped the initial issue, but this will be hard to undo if i'm wrong, so it can't hurt to double-check.)

blackshadowshade:
Let's do this. Bye bye, Jenkins!

cgolubi1:
I finally actually did this, terminated the instance, deleted its secondary volume, and released the jenkins.buttonweavers.com EIP.

PR #2400, opened 2018-09-19
Clean up setup/config files previously used by Jenkins

cgolubi1:
* Partially addresses #2399 (i still need to tear down the actual site and release the elastic IP and stuff)
* Testing done:
    * Whether CircleCI passes will be the strongest evidence that i didn't forget anything IMO
    * I brought up a site under Virtualbox and verified the usual dance of (1) `sudo -u vagrant /usr/local/bin/run_buttonmen_tests`, (2) `vagrant provision`, (3) run QUnit tests from a browser, (4) `sudo -u vagrant /usr/local/bin/run_buttonmen_tests` --- all tests passed
    * I brought up a dev site just as a quick "make sure nothing really dumb happened affecting a public BM site" smoke test --- it seemed fine.

Shadowshade, i think your best bet for your own smoke test is to make sure nothing you rely on for your local testing (AMP these days, right?) was removed by this.


cgolubi1:
Oh, on the off chance this isn't blindingly obvious/i didn't say it before: the reason i care about doing this is that i want to get rid of the puppet code we no longer need to maintain, so that i know what pieces of puppet code i do/don't need to figure out how to get working with a newer Ubuntu/PHP.

blackshadowshade:
So, it looks reasonable. I've tested with my AMP stack, and I can:
- launch the site locally
- play a move
- start a game
- create a new player
- delete and initialise the test database
- run all PHPUnit tests
- run all QUnit tests

That looks good to me. Merging.

PR #2403, opened 2018-10-03
pull from staging to production - test

cgolubi1:
Opening this pull with no details so github will tell me why it thinks i can't merge.  I'll close it or clean it up once i figure it out.

PR #2404, opened 2018-10-03
Keep trying to fix wrapper error by printing the message rather than the object

cgolubi1:
* Addresses what i think is the root of the merge problem giving us trouble in #2403


cgolubi1:
Irilyth, please review.  Note that:
* I don't fully understand why this led to a merge conflict going into production but not going into staging, but
* As long as we agree that there was a functional regression and this change addresses it, i don't think we need to understand --- i'm satisfied to improve the code and see if that makes git happier

blackshadowshade:
Looks simple enough.

Issue #2397, opened 2018-09-05
Plan 2018-10 release


cgolubi1:
Do all testing and deployments for the 2018-10 buttonweavers.com prod release.

PR #2405, opened 2018-10-05
Push changes from master to staging on 2018-10-05

cgolubi1:
This is part of #2397.

It brings in only #2404, an attempt to fix a python script wrapper regression that was causing merge conflicts.


cgolubi1:
I'm not planning to load this on staging, because the only change is to bmutils.py

Issue #2398, opened 2018-09-08
Site upgrade


blackshadowshade:
The production instance that we are currently running needs updating, for a number of reasons:
- to bring various software products up to a currently-supported version
- to allow us to use new features
- to bring in security patches

Specific issues that may relate to this site upgrade are:
- [ ] #161: Downloading PHPUnit via puppet
- [ ] #224: Using SSL
- [ ] #439: JQuery version
- [ ] #440: JSHint version
- [ ] #575: Puppet version
- [ ] #1083: Manage third-party Javascript libraries
- [ ] #1097: buttonweavers.com should redirect to www.buttonweavers.com
- [ ] #1167: Upgrade PHP
- [ ] #1168: Upgrade PHPUnit
- [ ] #1339: Backup drive running out of space
- [ ] #1989: Connection errors
- [ ] #2143: MySQL settings to enable temporary backups
- [ ] #2197: Python 3 compatibility
- [ ] #2208: Adding the statistical environment R
- [ ] #2342: Changing CircleCI to use the new upgraded versions

blackshadowshade:
So, I see a good endpoint here as:
- we're running a current version of Ubuntu
- we have the desired (for some value of desired) versions of PHP, PHPUnit, Javascript, JQuery, JSHint, other third-party Javascript libraries, MySQL, Python, R, R libraries
- the Button Men site hosted on the new instance works at least as well as it does now
- CircleCI runs successfully using the versions that we have selected
- backups work
- new dev sites that we create use the same updated versions
- the live production database has been fully migrated across from the old instance to the new instance

blackshadowshade:
One desirable endpoint is that we develop a procedure for doing such site upgrades, so that we can do them as necessary in the future.

blackshadowshade:
Here are the current problems that we are currently experiencing (off the top of my head), due to the old versioning of our software products:
- PHP 5.3 has been out of active support for five years, and out of security support for four years. This means that we can't support strong hashing of passwords. PHP 7.1 is about to stop being actively supported, so we should probably upgrade to PHP 7.2.
- PHPUnit needs to be upgraded to support the shift to PHP 7.2. The version of PHPUnit recommended  for supporting PHP 7.2 is PHPUnit 7. This will also help with PHPUnit recognising inline comment tags, which it does not do on our version.
- R needs to be version 3.2.3 or later (ideally 3.3.0 or later) to be compatible with the R packages that I'm using to generate the stats.
[...]

PR #2411, opened 2018-11-02
Site upgrade pull request

blackshadowshade:
This addresses much of #2398.

I've based this off the original 2398_site_upgrade branch created by @cgolubi1.

My changes are:
- to fix the change of behaviour of isset() when querying non-accessible properties via the magic method __get()
- to ensure that empty() behaves itself by implementing the __isset() magic method where it didn't already exist
- to change the behaviour of __unset() to setting a property explicitly to NULL, instead of actually unsetting it altogether
- to avoid unsetting properties wherever possible

I'm pretty sure I've dealt with all the problematic bits caused by the changes of behaviour in these functions between PHP 5.3 and 5.4.


blackshadowshade:
Labelling incomplete because of the Dead Guy failures.

blackshadowshade:
Closing this one, in favour of a new pull request that will address the isset() changes only.

PR #2412, opened 2018-11-04
Upgrade buttonweavers site to Ubuntu Xenial

### LINKED ISSUE #2398: Site upgrade

cgolubi1:
* Fixes #2398 
* See discussion in #2398 for site upgrade plan, and in #2411 for shadowshade's code changes


cgolubi1:
Ugh.  With shadowshade's new changes, this is failing out on every Dead Guy game, because it's no longer rolling zero-sided dice in all cases.

blackshadowshade:
I'm happy to do the site upgrade first and postpone the changes to make the code fully PHP 5.4+ compliant. I agree that the code changes are not harmless if they're causing your test rig to fail.

However, I have to disagree that most of the changes are unnecessary. With the PHP 5.4+ changes, many of the tests are simply not testing anything---I've tested manually, and they're not picking up bad cases, even when I create them artificially. Because the unit tests are compromised, that reduces my security that the live site code is actually doing the right thing too.

After the site upgrade is done and dusted, I'd suggest that we need a test that specifically targets the Dead Guy bug. We should also put our live site users on alert that there may be odd behaviour, and that they should report their suspicions sooner rather than later.

blackshadowshade:
BTW, have I mentioned recently that I dislike PHP and it's weakly-typed variables?

blackshadowshade:
And have I mentioned recently that PHP's object orientation is still broken?

blackshadowshade:
Let's close this pull request then, in favour of #2413.

PR #2417, opened 2018-11-21
Fix isset() and empty() calls broken by PHP version upgrade

### LINKED ISSUE #1167: Upgrade to PHP 5.5 or higher

blackshadowshade:
Fixes #1167.

So, this pull request is intended to fix the calls to isset() and empty() in the code and test code that were broken by the upgrade from PHP 5.2 to PHP 7.0.

We already know from Chaos's replay testing that these changes break something with Dead Guy, so I'll mark this as incomplete until we have a responder test for that problem and the problem has been fixed.

### DIFF
--- a/deploy/amp/phpunit_bootstrap.php
+++ b/deploy/amp/phpunit_bootstrap.php
@@ -40,5 +40,6 @@ function bm_skill_rand($min = FALSE, $max = FALSE) {
     return mt_rand();
 }
 
+
 // Now include the bootstrap file from the code itself
 require_once __DIR__.'/../../src/lib/bootstrap.php';
--- a/deploy/circleci/phpunit_bootstrap.php
+++ b/deploy/circleci/phpunit_bootstrap.php
@@ -40,5 +40,6 @@ function bm_skill_rand($min = FALSE, $max = FALSE) {
     return mt_rand();
 }
 
+
 // Now include the bootstrap file from the code itself
 require_once __DIR__.'/../../src/lib/bootstrap.php';
--- a/src/engine/BMAttack.php
+++ b/src/engine/BMAttack.php
@@ -371,7 +371,7 @@ abstract class BMAttack {
      */
     protected function ensure_defenders_have_value(array $defenders) {
         foreach ($defenders as &$def) {
-            if (empty($def->value)) {
+            if (is_null($def->value)) {
                 $def->roll(FALSE);
             }
         }
--- a/src/engine/BMButton.php
+++ b/src/engine/BMButton.php
@@ -373,7 +373,7 @@ class BMButton extends BMCanHaveSkill {
      */
     public function __unset($property) {
         if (isset($this->$property)) {
-            unset($this->$property);
+            $this->$property = NULL;
             return TRUE;
         } else {
             return FALSE;
--- a/src/engine/BMDie.php
+++ b/src/engine/BMDie.php
@@ -317,7 +317,7 @@ class BMDie extends BMCanHaveSkill {
              $doHooksAllowReroll);
 
         if ($turboDieIsAboutToBeReplaced) {
-            unset($this->value);
+            $this->value = NULL;
         } elseif ($needsToReroll) {
             $this->set__value(bm_rand($this->min, $this->max));
         }
@@ -604,7 +604,7 @@ class BMDie extends BMCanHaveSkill {
      */
     public function split() {
         $oldRecipe = $this->get_recipe(TRUE);
-        unset($this->value);
+        $this->value = NULL;
         $newdie = clone $this;
 
         $remainder = $newdie->max % 2;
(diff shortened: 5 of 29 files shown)


blackshadowshade:
Just bumping this as a reminder that I'm still waiting on a Dead Guy responder test that breaks with these changes.

cgolubi1:
I stood up a replay site, and i'm kicking off some Dead Guy games.

blackshadowshade:
I agree that it's an improvement! Looking forward to seeing what else pops up.

cgolubi1:
Another thing which i think is a bug in RandomAI, not in the backend.  The backend thinks this attack is legal, where my bot couldn't find a valid skill attack:

hairy_skill_attack

cgolubi1:
To elaborate, i assume my bot got tripped up by the fact that one of the twin stinger dice needs to contribute 1 to the attack, and in fact `g(1,1)` can't contribute 1, but `g(0,2)` can.  I can probably fix that in randomAI pretty easily --- i'm just mentioning it here to make sure my read of the desired behavior (that this attack indeed should be legal) is correct.

In retrospect, i wish i'd spent more time arguing against the implementation of zero-sided dice --- i think on the order of 50% of our logic debugging time is spent on zero-sided-die-related problems, and i think that was predictable at the time, and we should have been intelligent enough to say "is the fun quotient this brings worth it relative to other features?"  I know that's hard to predict, but i'm raising it now because i think "we will be debugging issues related to this implementation every time we touch the codebase" *was* predictable, and when we go to implement other features like that, we should be asking ourselves "is this actually going to add a sufficient amount of fun?"

blackshadowshade:
I guess my take on this is that we're going to have to make calls about fractional sided dice (like the 19.5 for Dr. Lucky), negative sided dice (like Fudge), and strange-sided dice (like Wildcard).

Our experience with zero-sided dice will give us a better perspective on how much angst these potential changes will give us. The major one that we'll probably have to implement at some stage is Wildcard, but there's a good reason I haven't touched that yet.

blackshadowshade:
And just noticed that I didn't explicitly confirm that I agree with your understanding of stinger, that a g(0,2) should be able to contribute a value of 1, but a g(1,1) cannot.

cgolubi1:
Thanks.  That's not why i hadn't gotten back to this (recently, it's because of the RDS thing, so... soon?  (!)), but it's good to have confirmed.

cgolubi1:
Starting over with this a bit --- the first 3000 or so games i replayed had about 13 failures, and all of them were of the form of the example i had very early in this thread, where the defending die is zero-sided, and used to be rerolled, and now isn't any more.  So i've modified the updater to account for the simplest form of that case, and i'm hoping that'll get the number of failures down to a number i can spot-check easily:

```python
#!/usr/bin/python
##### update_replay_games
# be made in random_ai.py.

import sys
sys.path.append('lib')

import random_ai

# Custom classes
  #
  def _write_entry_type_submitTurn(self, entry):
    # Set a flag so _php_rand_str() can know if this is a submitTurn entry
    self.in_submit_turn = True
    random_ai.PHPBMClientOutputWriter._write_entry_type_submitTurn(self, entry)
    self.in_submit_turn = False

  def _php_rand_str(self, retval):
    # This change corrects a bug in which a die is incorrectly rolled for a 0-sided die which is the target of an attack
    # Handle the most common case of that problem, a single 0-sider which is the target of an attack and is the first 0-sider to be rerolled during the attack.
    if self.in_submit_turn and 'against [(0):0]' in retval.message:
      self.f.write('// 1167_isset: removing first defender 0 from retval.rand_vals: %s\n' % retval.rand_vals)
      # If a 0 is the first thing, assume it's an attacker size change, and remove the second one
      if retval.rand_vals[0] == 0:
        nextzero = retval.rand_vals.index(0, 2)
        retval.rand_vals.pop(nextzero)
      else:
        retval.rand_vals.remove(0)
    return random_ai.PHPBMClientOutputWriter._php_rand_str(self, retval)

# Initial setup
if len(sys.argv) < 2:

DIRNAME = sys.argv[1]

  d.in_submit_turn = False
  d.translate_game(gamepath)
```

cgolubi1:
Replayed about 17000 games with the above recipe.  They included 7 failures, all of which were obvious corner cases which i got to pass by hand (e.g. target is `t(0)` instead of `(0)`.  I think we should merge this --- want to rebase it down to a single commit?

blackshadowshade:
Okay, that's been rebased.

Now I can implement zero-sided option dice and get my Fanatic playable!  :)

cgolubi1:
I'm not sure that's the right takeaway, James...

Anyway, merging, and apologies that it took so long.

PR #2419, opened 2018-11-22
Fix PHP7 password hash bug

### LINKED ISSUE #2418: players can't set passwords
If you browse to http://www.buttonweavers.com/ui/prefs.html, enter current/new passwords in the box, and click "Save preferences", you get an internal error.  The apache error log looks like this:

```
[Thu Nov 22 16:17:50.045340 2018] [:error] [pid 22165] [client 10.0.2.2:52436] PHP Fatal error:  Uncaught Error: Call to undefined function password_hashed() in /var/www/engine/BMInterfacePlayer.php:184
Stack trace:
#0 /var/www/engine/BMInterfacePlayer.php(150): BMInterfacePlayer->password_hashed_universal('c11')
#1 /var/www/api/ApiResponder.php(461): BMInterfacePlayer->set_player_info(1, Array, Array)
#2 /var/www/api/ApiResponder.php(950): ApiResponder->get_interface_response_savePlayerInfo(Object(BMInterface), Array)
#3 /var/www/api/responder.php(21): ApiResponder->process_request(Array)
#4 {main}
  thrown in /var/www/engine/BMInterfacePlayer.php on line 184, referer: http://localhost:8080/ui/prefs.html
```

cgolubi1:
* Fixes #2418

### DIFF
--- a/src/engine/BMInterfacePlayer.php
+++ b/src/engine/BMInterfacePlayer.php
@@ -181,7 +181,7 @@ class BMInterfacePlayer extends BMInterface {
         if (version_compare(phpversion(), "5.5.0", "<")) {
             return crypt($password);
         } else {
-            return password_hashed($password);
+            return password_hash($password, PASSWORD_DEFAULT);
         }
     }
 
--- a/test/src/api/responder00Test.php
+++ b/test/src/api/responder00Test.php
@@ -180,6 +180,8 @@ class responder00Test extends responderTestFramework {
             'monitor_redirects_to_game' => 'false',
             'monitor_redirects_to_forum' => 'false',
             'automatically_monitor' => 'false',
+            'current_password' => 't',
+            'new_password' => 'f',
         );
         $retval = $this->verify_api_success($args);
         $this->assertEquals($retval['status'], 'ok');


blackshadowshade:
Looks good to me.

blackshadowshade:
I've noticed a problem with this pull request with the test code that's been added, which is specific to the way that I run my unit tests, namely, that I don't always run off a clean test database.

When I run the unit tests a second time, they fail with the message "Current password is incorrect."

I guess that's not surprising, because the second time, the password would have already changed.

I wonder if it's easy to fix this, perhaps by simply creating a specific user for this password change test.

blackshadowshade:
For the moment, I'd suggest backing out the unit test change. @cgolubi1, are you happy with this idea? If so, I'll submit a pull request for it.

cgolubi1:
Not necessary.  Changing the password to itself would reproduce this problem, because it's just a code coverage problem, so i'll do that.

FWIW, i think it'd make our lives easier if everyone got in the habit of running unit tests off a clean database.  I always do, and it means we don't need to worry about things like this, which are basically changes we're making to keep our test code happy.  Given that there's a workaround in this case, i don't need to insist.

I feel strongly that we end the day with a unit test for this regression --- we broke prod in a way that locks people out of the site; that's not a trivial problem.

Issue #2307, opened 2017-12-24
generating 40 t(0) dice is too silly and we should do something different


cgolubi1:
The label comes from blackshadowshade's comment about this on #2306.  My actual request is: we create and enforce a rule that, whenever radioactive is triggered (a die splits due to radioactive), all dice involved in the attack which have the radioactive skill lose it.

Note: this issue will be satisfied whether or not trip dice continue to decay after unsuccessful attacks on radioactive dice --- the request is just that IF a die decays THEN the radioactive skill is lost.

cgolubi1:
There are several types of rationale we've historically used when deciding how skills are going to be implemented.  Here are all the ones i can think of offhand:
* Authorial intent: what did the original authors of the skill have in mind?
* Prior art: how did the DHS site implement this?
* Simplicity: what makes the rule for the skill easy to state in words and understand?
* Halting: what allows us to avoid non-terminating or broken game states in the implementation?
* Fun: what seems like people would enjoy playing it?

Prior art weighed heavily in our earlier discussions of implementation, because our site didn't work yet, and we all (thought we) remembered how the DHS site worked somewhat well.  IMO it's much less important (and harder to remember) now.

For me right now, the most important considerations are simplicity and halting, and on both of those grounds, "the radioactive skill is always removed from dice when it is triggered" strikes me as a no-brainer implementation choice:
* Simplicity: Either radioactive is triggered or it isn't.  If it is triggered, the same sequence of things (decay + loss of radioactive) always happens.  This is less complicated than any implementation in which loss of the radioactive skill sometimes does and sometimes doesn't accompany radioactive decay.
* Halting: radioactive is a destabilizing skill with a lot of weird side effects.  If you can radioact at most N times, where N is the number of dice in both players' buttons - 1 (i think), that makes the number of weird side effects we will see that much more tractable.

To me, those are going to outweigh an authorial intent or fun argument unless it's really really compelling.  But if anyone likes the current behavior (which i'm going to persist in calling "inconsistent") have at it, and remind me why we think that behavior is good.

cgolubi1:
Can we please raise this topic again?  I feel strongly that i'm right about my initial ask: "whenever radioactive is triggered (a die splits due to radioactive), all dice involved in the attack which have the radioactive skill lose it".  I'd like to ask people who care to engage with the argument.

[...]

PR #2423, opened 2018-12-02
Added Jolt skill

### LINKED ISSUE #275: Jolt die skill needed
Jolt: J
(Exclusive to this website; unofficial)
Jolt Dice were created by Randy Chertkow (aka Famous) for this web site. They are represented by a "J" before a die. They are normal in all respects, except that when they attack, or are attacked. If a Jolt die attacks, or participates in an attack, the player will get to take another turn and the Jolt die losses the Jolt skill. If a Jolt die is captured, then the player who captured it takes another turn.

Inconsistencies:

```
J^ (Jolt Time & Space): Will not give more than 1 extra turn if rerolled to an odd number. 
```

blackshadowshade:
Fixes #275.

At some stage, this should probably have a responder test added (e.g., Button of Loathing vs Tirade), but from my testing, it looks like it works okay.


blackshadowshade:
One comment on this: I'll also need to update the description of Radioactive, but I don't want to have this conflict with the new wording that will be happening to address #2307.

cgolubi1:
Alright, i'm starting to try caching `update_replay_games` versions that work for various cases because i can't keep the details in my tiny brain.  In this case, it took me much longer than it should have to remember that the *only* purpose of $RANDOMBM_SKILL_ARRAY in responderTestFramework.php is to make it so that we don't have to renumber all skills in all tests recorded in the codebase (i.e. we can refer to skills by letter within the PHP tests).  It has nothing to do with the mapping from `bm_skill_rand()`'s output to skills, because that's handled by the backend, not by the test code.

So here's the `update_replay_games` override i'm now trying for this site:

```
  #
  def _php_dict_style_rand_str(self, retval):
    if retval.skill_rand_vals:
      for i in range(len(retval.skill_rand_vals)):
        if retval.skill_rand_vals[i] >= 5:
          retval.skill_rand_vals[i] += 1
    return random_ai.PHPBMClientOutputWriter._php_dict_style_rand_str(self, retval)
```

cgolubi1:
Replay testing looks good --- it's played nearly 19000 novel and replayed games.  I'm going to hack the client to play only RandomBMPentaskill novel games, to try to find any internal errors faster, but i don't think i'm going to find anything else via replay testing.  Time to stand up a dev site.

cgolubi1:
Dev site created.  irilyth, at your leisure, can you sync up: ephemeral 3.91.68.115

irilyth:
> Dev site created. irilyth, at your leisure, can you sync up: ephemeral 3.91.68.115

Done!

```
+$ host ephemeral.dev.buttonweavers.com
ephemeral.dev.buttonweavers.com has address 3.91.68.115
```

dwvanstone:
I did a bunch of testing on the dev site and found Jolt to be working well.

cgolubi1:
Okay, replay testing is still doing fine.  I say let's merge this; thanks, all.

Issue #1209, opened 2014-08-15
Refactor API to use same syntax as backend for describing random buttons


blackshadowshade:
Taken from #1187:

Proposed refactor in another ticket to make the API and the backend use the same syntax for describing random buttons

cgolubi1:
I've reminded myself of the existence of this issue, and am going to grab it for my milestone, because as discussed numerous times, i consider it blocking for any kind of pre-game "specify subset of buttons for RandomBM/opponent choice" type issues, and i also think if we're going to rewrite our API encoding, we need to do it before beta.

PR #2426, opened 2018-12-30
JSON serialize API input for submission

cgolubi1:
* Prerequisite for #1209
* This change modifies the format of ALL API posts, so it is a breaking change for clients, but the required changes should be very simple
* No API method syntax is changing, but the purpose of this commit is to unblock future work by allowing us to pass args more cleanly than urlencoding would allow

### DIFF
--- a/src/api/ApiSpec.php
+++ b/src/api/ApiSpec.php
@@ -63,7 +63,7 @@ class ApiSpec {
                 'dieValueArray' => array(
                     'arg_type' => 'array',
                     'has_keys' => FALSE,
-                    'elem_type' => 'alnum',
+                    'elem_type' => 'number',
                 ),
             ),
         ),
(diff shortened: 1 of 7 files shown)


cgolubi1:
I'm going to do a fair bit of replay testing on this --- you'll see a few changes to `ApiSpec.php`, which are all because the quality of the data types we are giving the API by default are improving, but ApiSpec had to be updated to allow the better data type, so i want to use testing to find any other edge cases where that's true, though ideally everything in the spec would be covered by responder tests already.

blackshadowshade:
So, if I understand correctly, currently I just sit on this pull request until replay testing is done?

I don't see any issues with the pull request after a brief look, although there's a function or two that I'll need to research.

cgolubi1:
Yep, you understand correctly: i put it up for PR to find out what concerns you have other than anything replay testing will turn up.

cgolubi1:
My replay site has played 4400 replayed and novel games with no differences, so i feel pretty good about that.

I'll go ahead and rebase this, and then you can review it?

blackshadowshade:
I've loaded this onto my MAMP stack, tested that a standard Soldiers game can start, swing setting works, turns can be played, and a round ends correctly. I've also tried this with a Lady Luck mirror match, and that worked too.

Looks good to me.

Issue #2408, opened 2018-10-17
Fan button set (nominally GWAR, but could also be themed to Ultimate Muscle)


blackshadowshade:
theorgg would like to test a fan button set, as mentioned here:

http://www.buttonweavers.com/ui/forum.html#!threadId=660

Discussion continued in game chat, see here:

http://www.buttonweavers.com/ui/game.html?game=43032 (and also chat from previously linked games)

Recipes and images were received by email (see next post).

blackshadowshade:
Recipes:

1. Blothar, Berzerker Warrior
(6) (20) (30) (X) `(8) `(8) `(12)

2. Balsac the Jaws O' Death
M(6) M(8) (12) (20) (P)
(A "P" swing die is from 1 to 30.)

3. Jizmac Da'Gusha
M(4,4) (8) g(12) (12) (P)
(A "P" swing die is from 1 to 30.)

4. Pustulus Maximus
g(4) g(12) (12) (20) (X)

5. Beefcake the Mighty
H(4) H(6) (12) (20) (V,V)

6.Sleazy P. Martini
Stinger(4) (20) (X) (X) (X)

7. Dickie Duncan
(4,4) (6,6) (8,8) (Y,Y)

8. Techno Destructo
(12) (8,8) p(30) p(30) (X)

9. Sawborg Destructo
(12) (12) (20) (20) p(P)
(A "P" swing die is from 1 to 30.)

10. Bozo Destructo
p(2) (2) (6) (8)

11.Estrogina
(6) (12) M(T) (T)

12.Mr. Perfect
(1) (1) (2) (4) M(6)

13.The Master
(30) (30) (30) (U)

14.Gor-Gor
(8) (12) (20) p(30) (U)

15. Slyminstra Hymen
(4) (6) g(8,8) (12) (Y,Y)

16. Sexecutioner
(6,6) (12) (X) (X) (X)

17. Flattus Maximus
(4) (12) p(12) (20) p(X)

18. Dr. Skullheadface
g(X) (X) p(Y) (Y)

19. Morality Squad
(4) (6) H(6) (V,V) `(6) `(8)

20. Cardinal Syn
g(4,4) g(8) (8) (12) (Y)

21. Oderus Urungus
(6) (8) (10) (12,12) (X)
Any Poison dice Oderus captures are scored as standard dice, not as poison dice.

22. Porcelon, Portal Potty
(6) (8) (12) (12)
[...]

PR #2427, opened 2019-01-08
Fixed incorrect interaction between warrior and stinger

blackshadowshade:
In #2408, @TheOrgg notes that Warrior Stinger dice don't seem to be able to use partial values as part of their initial multidie skill attack.

This pull request fixes that problem.

### DIFF
--- a/src/engine/BMSkillWarrior.php
+++ b/src/engine/BMSkillWarrior.php
@@ -17,7 +17,6 @@ class BMSkillWarrior extends BMSkill {
      */
     public static $hooked_methods = array('initiative_value',
                                           'attack_list',
-                                          'attack_values',
                                           'capture',
                                           'post_roll',
                                           'score_value',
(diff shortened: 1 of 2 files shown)


irilyth:
I don't think the current behavior is incorrect: Warrior dice are out of play, so they can't use any of their skills until they come into play.

This is how the old site worked, there's at least one existing button that has this behavior, and I don't think we should change it.

blackshadowshade:
So you're arguing that warrior stinger dice aren't stinger until they're in play? I think that's a difficult argument to make, based on the current skill description of warrior.

blackshadowshade:
Interesting. #1541 directly addressed this issue, so @irilyth has history on his side.

irilyth:
The current skill description of Warrior also explicitly addresses this:

> Dice with the Warrior skill are completely out of play: They aren't part of your starting dice, they don't count for initiative, they can't be attacked, none of their other skills can be used, they don't count for scoring purposes, etc. 

"none of their other skills can be used", yeah? :^ )

blackshadowshade:
We discussed this before in #1541. However, the very fact that I've looked at this in the last 24 hours and it *still* wasn't clear to me that stinger shouldn't apply means that we really need to make this interaction explicit. Currently, there's no interaction clarifications for warrior, and I think this is probably a problem.

blackshadowshade:
Also, we seem to have lost the wording "The Warrior die adds its full value to the Skill Attack" somewhere in history.

blackshadowshade:
I'll close this one, in favour of a new pull request that updates the description of Warrior.

TheOrgg:
So a stinger warrior d20 is damn near useless, if I'm reading the thread correctly.  Yes?

I guess Mr. Perfect has to become a (1)(1)(2)(3)(V).

PR #2428, opened 2019-01-08
Added skill interaction description for warrior stinger

blackshadowshade:
This follows on from the discussion on #2427.

I've added the interaction between warrior and stinger explicitly so that it's clear (especially to me) in the future that stinger doesn't apply when a die still has warrior.


cgolubi1:
Just to share with other watchers - the wording shadowshade proposes is: 'A Warrior die adds its full value to a Skill Attack'

That seems fine to me.  Anyone want to wordsmith it?  Barring objections in a couple of days, i'll merge this.

irilyth:
I don't feel strongly about it, but if what you're trying to emphasize here is that you can't use the Stinger skill, something like "A Warrior can't use the Stinger skill to add less than the full value of the die, because the die isn't in play yet" or something.

blackshadowshade:
Description updated.

PR #2434, opened 2019-02-07
Python client should sort keys when JSON-encoding them for API submission

### LINKED ISSUE #2433: replay_loop mismatches due to ordering of swing and option die settings
RandomAI's replay loop on master, which should be a complete noop, is failing regularly with diffs like:

```
13) responder99Test::test_interface_game_00096
--- Expected
+++ Actual
@@ @@
             'player' => 'responder003'
-            'message' => 'responder003 set swing values: X=13, R=5, W=10'
+            'message' => 'responder003 set swing values: R=5, W=10, X=13'
         )
         3 => Array (...)
     )
     'maxWins' => 3
     'playerDataArray' => Array (...)
     'gameChatLogCount' => 0
     'validAttackTypeArray' => Array (...)
 )
```

or

```
1) responder99Test::test_interface_game_00012
--- Expected
+++ Actual
@@ @@
             'player' => 'responder004'
-            'message' => 'responder004 set option dice: (6/12=12), (4/8=4), (6/20=20)'
+            'message' => 'responder004 set option dice: (4/8=4), (6/12=12), (6/20=20)'
         )
         2 => Array (...)
     )
     'maxWins' => 3
     'playerDataArray' => Array (...)
     'gameChatLogCount' => 0
     'validAttackTypeArray' => Array (...)
 )
```

This must be related to the API encoding change, and clearly i can't use RandomAI replay testing until it's addressed.

cgolubi1:
* Fixes #2433

The `LOGF.flush()` is unrelated --- it's something i want to get my log updated faster, and have been too lazy to submit a separate pull for it.

### DIFF
--- a/src/api/ApiResponder.php
+++ b/src/api/ApiResponder.php
@@ -944,10 +944,12 @@ class ApiResponder {
             if ($argcheck['ok']) {
                 // As far as we can easily tell, it's safe to call
                 // the function.  Go ahead and create an interface
-                // object, invoke the function, and return the result
-                $interface = $this->create_interface($args, $check);
-                apache_note('BMAPIMethod', $args['type']);
-                $data = $this->{$check['funcname']}($interface, $args);
+                // object, invoke the function on sanitized args,
+                // and return the result
+                $sanitizedArgs = $this->spec->sanitize_function_args($args);
+                $interface = $this->create_interface($sanitizedArgs, $check);
+                apache_note('BMAPIMethod', $sanitizedArgs['type']);
+                $data = $this->{$check['funcname']}($interface, $sanitizedArgs);
 
                 $output = array(
                     'data' => $data,
--- a/src/api/ApiSpec.php
+++ b/src/api/ApiSpec.php
@@ -715,6 +715,42 @@ class ApiSpec {
     }
 
     /**
+     * This function sanitizes the provided arguments into a
+     * standard format for internal use.  It is run after
+     * verify_function_args() has already passed on these arguments,
+     * so it assumes the args are sane.
+     *
+     * @param array $args
+     * @return array
+     */
+    public function sanitize_function_args($args) {
+        $sanitizedArgs = array();
+        $argsExpected = $this->functionArgs[$args['type']];
+        foreach ($args as $argname => $argvalue) {
+            if (($argname == 'type') || ($argname == 'automatedApiCall')) {
+                $sanitizedArgs[$argname] = $argvalue;
+                continue;
+            }
+
+            $expectedType = $this->determineExpectedType($argname, $argsExpected);
+            if (is_array($expectedType)) {
+                $realArgtype = $expectedType['arg_type'];
+            } else {
+                $realArgtype = $expectedType;
+            }
+
+            // Use a sanitization function if one exists, or just return the initial arg
+            $sanitizeFunc = 'sanitize_argument_of_type_' . $realArgtype;
+            if (method_exists($this, $sanitizeFunc)) {
+                $sanitizedArgs[$argname] = $this->$sanitizeFunc($argvalue);
+            } else {
+                $sanitizedArgs[$argname] = $argvalue;
+            }
+        }
+        return $sanitizedArgs;
+    }
+
+    /**
      * Determine expected type of an argument
      *
      * @param string $argName
@@ -776,6 +812,17 @@ class ApiSpec {
             }
             return FALSE;
         }
+    }
+
+    /**
+     * sanitize an array argument by sorting it by key
+     *
+     * @param array $arg
+     * @return array
+     */
+    protected function sanitize_argument_of_type_array($arg) {
+        ksort($arg);
+        return $arg;
     }
 
     /**
--- a/test/src/api/responder01Test.php
+++ b/test/src/api/responder01Test.php
@@ -3549,7 +3549,7 @@ class responder01Test extends responderTestFramework {
      * 14. responder003 performed Power attack using [(X=20):7] against [q(T=2):2]; Defender q(T=2) was captured; Attacker (X=20) rerolled 7 => 20
      * 15. End of round: responder003 won round 2 (80 vs. 18)
      * 16. responder004 added a reserve die: rz(S)
-     * 17. responder004 set swing values: T=2, W=4, X=4, Z=4, S=20
+     * 17. responder004 set swing values: S=20, T=2, W=4, X=4, Z=4
      *     responder004 won initiative for round 3. Initial die values: responder003 rolled [(4):2, (6):5, (12):1, (X=20):12, (20):8], responder004 rolled [q(T=2):2, q(W=4):4, q(X=4):1, q(Z=4):1, z(S=20):4].
      * 18. responder004 performed Skill attack using [q(T=2):2,q(W=4):4,q(X=4):1,q(Z=4):1,z(S=20):4] against [(X=20):12]; Defender (X=20) was captured; Attacker q(T=2) rerolled 2 => 1; Attacker q(W=4) rerolled 4 => 4; Attacker q(X=4) rerolled 1 => 3; Attacker q(Z=4) rerolled 1 => 1; Attacker z(S=20) rerolled 4 => 10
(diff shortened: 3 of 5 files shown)


cgolubi1:
I need to do some more replay testing to verify that this works, but i wanted to put the pull up to see if it seems acceptable to Shadowshade to address this in the python client.

The other argument is that the python client shouldn't need to sort its inputs; the server should sort what it receives so that it's always returning exactly the same response when operating on the same dice behind the scenes --- if you tell it to update dice 1, 2, and 3, it should tell you "i updated dice 1, 2, 3" and if you tell it to update dice 2, 3, and 1, it should tell you "i updated dice 1, 2, 3".  That argument says that this is a server bug.

cgolubi1:
My argument that solving this client-side is okay for now more or less amounts to "don't boil the ocean":
* We *should* provide a consistently-behaved python client for people to use regardless of whether of whether we want RandomAI to use it, so this is an improvement anyway
* We don't know of any actual player problem caused by the server parsing die values in the order given
* In order to make server-side changes, we need to be able to regression-test through the API encoding layer, which means we need tests that pass.  Once we're in good shape to replay games again, if we want to force the server to process things consistently, we can intentionally code RandomAI to scramble keys, and prove we've fixed the issue in the server.  Right now, that's much harder.

HMM.  That said, i think we had better get some replayable data and then solve this server-side, because *python's* `json.dumps()` may support a sort_keys flag, but *javascript's* `JSON.stringify()` doesn't: see https://stackoverflow.com/questions/16167581/sort-object-properties-and-json-stringify and https://www.npmjs.com/package/json-stable-stringify - so if we want players to see consistent results for die submission settings, i think we have to solve this server-side.

tl;dr: don't accept this pull.  My plan is:
* Use this code to get some stable replay results
* Back out this change and fix the server side, and then demonstrate that both:
    * Replaying against my newly-recorded results with the force-sorted client looks good (i didn't break anything server-side)
    * Playing and replaying new games using the unsorted client works (so things should look consistent using other unsorted clients, such as the JS UI)

cgolubi1:
I'm sure no one is curious why testing didn't catch this before JSON API encoding was accepted, but you're a captive audience, so i'm going to tell you anyway.  Right now, replay loop does the following things:
1. record novel games in pck format using RandomAI (to make sure they don't blow up and that RandomAI's logic about what plays are valid matches the server's logic), then translate those novel games from pck format to php format (to make sure nothing in that conversion/in randomAI's logic about what a valid PHP/respondertest game looks like, blows up on the new code)
2. Take pck games which were recorded on the previous state of master, translate *those* to PHP format, and actually run them through phpunit as responderTest games (to make sure nothing has changed in how the backend responds to particular inputs)

What replay_loop doesn't do, which is what would have caught this, is run the novel tests through phpunit as responderTest games.  The primary reason not to do that is that it's expensive for a test that's irrelevant the vast majority of the time --- the thinking is that it's almost never the case that a bug causes inconsistent results between game recording and game replay.  It takes about as much time to replay 100 games you just recorded, as it does to record 100 new games, or to replay 100 games from the previous master.  And those latter two things catch a lot of bugs, whereas replaying the 100 new games catches only a really narrow class of bugs.  (The pck -> php conversion steps are very cheap time-wise, so replay_loop just does them all the time on spec, because one time that would have caught a problem for some random reason so why not?)

However, the recording-replay loop that i run against master does replay its own games, so when a breaking bug like this is pushed, it shows up in the master tests, probably typically before the next prod release (as in this case, though in this case we were stymied because i hadn't actually looked at the replay results in awhile).

Reading the above, it seems like the right fix is for the normal replay_loop to replay its novel games... sometimes.  F'rex, every ten batches of novel games and replayed older games, it could phpunit the novel games.  That's not a big performance penalty, and it would catch stuff like this that otherwise makes a mess after being merged.

blackshadowshade:
Adding the incomplete label so that I remember not to accept this for the moment.

blackshadowshade:
You'll need to update the responder tests that are failing, because this (correctly) sorts swing keys.

cgolubi1:
Yeah, i saw, thanks.

There's also some weird problem with turbo dice i found in replay testing, that i'll need to sort out.  Oh, i bet i know what it is...  Anyway, i'll let you know when i think this is actually good to go at all, and fix the tests before then.

cgolubi1:
I'm also very tempted to go ahead and move the sort into `ApiResponder::process_request` and `ApiSpec::verify_function_args`.  I don't see any reason each function should need to sort its own args like a schlub, and later in the game if we're going to want to address BMT-number-whatever from a long time ago, and have all args internally treated as numbers if they're numbers, booleans if they're booleans, etc, sanitizing at the front gate seems like the sane way to handle things.

blackshadowshade:
I imagine the changes in the responder tests will be rather informative, in the case where other unanticipated keys also end up being sorted.

cgolubi1:
Shadowshade: feel free to have initial opinions about this solution.  It's in replay testing now.

blackshadowshade:
Looks okay to me. Once you're happy with replay testing and have rebased this, I'm happy to accept it.

cgolubi1:
I successfully replayed 6000 novel games, and now i'm working on replaying some older games in hopes that i in fact remembered to record some games using the sorted client and can replay those.

cgolubi1:
A large number of replayed games from the client-side-sorted version worked just fine, so i rebased and i'm going to say this is good to go if CircleCI passes.

cgolubi1:
Jota: i stood up a dev site with this code version at 34.224.26.217 --- this would be a good place for you to test your bot against the new API encoding.

cgolubi1:
Irilyth, when you get a moment can you sync up 34.224.26.217 == blip.dev.buttonweavers.com?

blackshadowshade:
So, I'm good to go with this as soon as we're all ready to roll.

irilyth:
> Irilyth, when you get a moment can you sync up 34.224.26.217 == blip.dev.buttonweavers.com?

Done!

```
+$ host blip.dev.buttonweavers.com
blip.dev.buttonweavers.com has address 34.224.26.217
blip.dev.buttonweavers.com is an alias for blip.dev.buttonweavers.com.
blip.dev.buttonweavers.com has address 34.224.26.217
```

cgolubi1:
Shadowshade: go ahead and merge it --- we can keep the dev site up even after it's merged, it was just the easiest way to get something for Jota to start testing the bot against.

blackshadowshade:
Roger!

cgolubi1:
Hmm, i found this in the apache error log from Jota's testing:

```
[Mon Mar 04 23:14:40.183855 2019] [:error] [pid 20974] [client X.X.X.X:Y] PHP Warning:  array_key_exists() expects parameter 2 to be array, null given in /var/www/api/ApiResponder.php on line 998
[Mon Mar 04 23:30:15.957012 2019] [:error] [pid 20974] [client X.X.X.X:Y] PHP Warning:  array_key_exists() expects parameter 2 to be array, null given in /var/www/api/ApiResponder.php on line 998
[Tue Mar 05 01:03:09.312175 2019] [:error] [pid 16405] [client X.X.X.X:Y] PHP Warning:  array_key_exists() expects parameter 2 to be array, null given in /var/www/api/ApiResponder.php on line 998
```

Taking a look to see if there's a problem there.

cgolubi1:
That code line is:

```
    protected function verify_function_access($args) {
        if (array_key_exists('type', $args)) {
```

So i think that's just what happens when someone calls the responder with no args, which is presumably something that could happen during testing.  I assume `array_key_exists()` returns false in that case, and you get dropped into the "error: no type specified" code path, and at any rate, that's not a regression from the current behavior.  Not worried about it --- Jota, if you happen to remember that the site did something particularly unfriendly in that case (where i would consider the friendly behavior to be sending you the:

```
            $result = array(
                'ok' => FALSE,
                'message' => 'No "type" argument specified',
            );
```

response), feel free to file an issue and we can clean it up.  But i think this is fine.

cgolubi1:
Jota: are you still using the "blip" dev site to test Nala, or are you all set?  When i next check, i'll take silence as permission to tear that site down.

AdmiralJota:
Should be all set -- now that you've pushed the changes to Staging, any of her future testing for this should be done there anyway.

PR #2436, opened 2019-02-26
Remove radioactive skill from uncaptured defending dice

### LINKED ISSUE #2307: generating 40 t(0) dice is too silly and we should do something different

blackshadowshade:
Fixes #2307.

I think it matches the intent of the change suggested by Chaos in #2307, but it will need to be tested by RandomAI for unexpected changes.


blackshadowshade:
CircleCI run is successful: https://circleci.com/gh/blackshadowshade/buttonmen/184

cgolubi1:
Standing up a replay site.  My initial guess is that i won't need to write code to handle expected API changes, because failures should be rare (less than one failure per hundred games), reflecting the frequency of games containing both trip and radioactive where an unsuccessful trip attack involving the radioactive die actually happens, so i can just spot-check the failures, and play a bunch of new games to look for internal errors.  Let's test that theory.

cgolubi1:
This looks good to me.  The site played 26500 each replayed and novel games, which yielded 42 total failures.  All failures involved a Trip attack with a Radioactive defender, where the attacker split and as a result of the change, the defender now loses Radioactive.  Here's an example diff:

```
1) responder99Test::test_interface_game_00047
--- Expected
+++ Actual
@@ @@
-'responder004 performed Trip attack using [t(2):2] against [%(12):5]; Attacker t(2) rerolled 2 => 2; Defender %(12) rerolled 5 => 9, was not captured; Attacker t(2) showing 2 split into: t(1) showing 1, and t(1) showing 1. responder004's idle ornery dice rerolled at end of turn: o(8) rerolled 2 => 7. '
+'responder004 performed Trip attack using [t(2):2] against [%(12):5]; Attacker t(2) rerolled 2 => 2; Defender %(12) recipe changed to (12), rerolled 5 => 9, was not captured; Attacker t(2) showing 2 split into: t(1) showing 1, and t(1) showing 1. responder004's idle ornery dice rerolled at end of turn: o(8) rerolled 2 => 7. '

/buttonmen/test/src/api/responderTestFramework.php:1318
/buttonmen/test/src/api/responder99Test.php:71650
```

So i don't get past the diff in the output to demonstrate that the diff in the die settings is correct, but the responderTest changes shadowshade committed imply that it is correct, and regardless if it were going to blow up, that many novel games wouldn't have passed.

So i think i should merge this.

cgolubi1:
Oh, for future reference, here's the modified update_replay_games i used to get the skill description change:

```
#!/usr/bin/python
##### update_replay_games
# be made in random_ai.py.

import sys
sys.path.append('lib')

import random_ai

# Custom classes
  #
    if 'gameSkillsInfo' in data and 'Radioactive' in data['gameSkillsInfo']:
      data['gameSkillsInfo']['Radioactive']['description'] = 'If a radioactive die is either the attacking die or the target die in an attack with a single attacking die and a single target die, the attacking die splits, or "decays", into two as-close-to-equal-sized-as-possible dice that add up to its original size. If any die decays during an attack, all dice involved in the attack that remain in play lose Radioactive (%). All dice that decay lose the following skills: Turbo (!), Mad Swing (&), Mood Swing (?), Time and Space (^), and Jolt (J). For example, a s(X=15)! (Shadow Turbo X Swing with 15 sides) that shadow attacked a radioactive die would decay into a s(X=7) die and a s(X=8) die, losing the turbo skill. A %p(7,13) on a power attack would decay into a p(3,7) and a p(4,6), losing the radioactive skill.'
    return data

# Initial setup
if len(sys.argv) < 2:

DIRNAME = sys.argv[1]

  d.translate_game(gamepath)
```

PR #2437, opened 2019-03-04
Updated stats script to generate player vs player stats

blackshadowshade:
This should be unproblematic because it only touches the R script that I use to generate stats.


cgolubi1:
Sure, that's fine.

PR #2443, opened 2019-04-03
RandomAI support for Jolt

cgolubi1:
This lets RandomAI continue to work after the merge of #2423


cgolubi1:
I'd argue you can merge this anyway because none of our CircleCI testing uses random_ai.py at all (and if you did i could keep doing replay testing).  But it's not super critical either way, so it's fine if you'd rather wait until we sort #2444).

blackshadowshade:
Okay, happy to merge this.

PR #2450, opened 2019-04-07
Make skill interaction text between Jolt and TimeAndSpace identical between both uses

### LINKED ISSUE #2448: stupid replay_test creatures
I see a replay test failure in my master replay loop with:

```
1) responder99Test::test_interface_game_00002
--- Expected
+++ Actual
@@ @@
             'interacts' => Array (
-                'Jolt' => 'If a die with both the Jolt and TimeAndSpace skills rerolls to an odd number, it still only gives one extra turn.' 
+                'Jolt' => 'If a die with both the Jolt and TimeAndSpace skills rerolls to an odd number, it still only gives one extra turn'
             )
         )
     )
     'gameActionLog' => Array (...)
     'maxWins' => 3
     'playerDataArray' => Array (...)
     'gameChatLogCount' => 0
     'validAttackTypeArray' => Array ()
 )
```

(One of those ends with a period, one doesn't.)  Looks like there's a mismatch in the skill interaction text in responderTestFramework.php

cgolubi1:
* Fixes #2448 (if it passes)

### DIFF
--- a/src/engine/BMSkillJolt.php
+++ b/src/engine/BMSkillJolt.php
@@ -140,7 +140,7 @@ class BMSkillJolt extends BMSkill {
     protected static function get_interaction_descriptions() {
         return array(
             'TimeAndSpace' => 'If a die with both the Jolt and TimeAndSpace skills rerolls to an odd number, ' .
-                              'it still only gives one extra turn.',
+                              'it still only gives one extra turn',
         );
     }
 }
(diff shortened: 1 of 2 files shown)


blackshadowshade:
I'm not certain this is the correct way to fix this, considering that the other interactions all don't have full stops. Can we remove the full stop from the other bit instead?

cgolubi1:
I'm not sure we're 100% consistent, but i definitely have no problem with doing this one the other way, so i don't need to be sure.  Changed.

blackshadowshade:
Looks good. I'll accept this once CircleCI passes.

cgolubi1:
Incidentally, if we *want* to be 100% consistent, we should slap a unit test somewhere that blows up if any interaction text ends with a period.  Unit tests are better at being pedantic than almost all people.

blackshadowshade:
It's probably a good idea to have some automated way of finding this sort of problem without needing to wait until your RandomAI blows up.

PR #2463, opened 2019-05-29
Specified request_once path more carefully by using __DIR__

### LINKED ISSUE #2462: Make PHP file requires more precise with respect to path
At the moment, I'm having trouble running any individual PHPUnit tests that require any of the TestDummy*.php files.

I think that I can fix this problem by using \_\_DIR\_\_ in the require path (see https://stackoverflow.com/questions/32444572/why-include-dir-in-the-require-once)

blackshadowshade:
Fixes #2462.

The unit tests all pass on my AMP stack. Let's see what CircleCI thinks of this.

Then, it's probably a good idea to test this on a dev site just in case I've accidentally broken a require_once link that affects game logic.


blackshadowshade:
And CircleCI passes too.

cgolubi1:
Does this actually solve the problem you're trying to solve, with running the various TestDummy* tests by hand?  You didn't say.

blackshadowshade:
Yes, this solves the problems, now I can run all (as opposed to only some) of the individual class tests from both NetBeans and CLI.

cgolubi1:
Sure, i'll throw a dev site together for this, honestly i think any (including you) can just click around the site for a few minutes, then we can declare victory and merge it.

cgolubi1:
I clicked around my dev site for a few minutes, took some turns, looked at the forum, changed my preferences.  It seems fine.

PR #2487, opened 2019-10-18
Exclude some common unimportant error log messages

cgolubi1:
This is to reverse some bitrot in the amount of noise e-mail i've been getting from buttonmen apache error logs, to make scanning them for real errors less time-consuming.

I tested the recipe on some live dev sites that have been spamming me, and verified that the greps work to remove the messages they're targeting.

### DIFF
--- a/deploy/vagrant/modules/apache/templates/monitor_apache_logs.erb
+++ b/deploy/vagrant/modules/apache/templates/monitor_apache_logs.erb
@@ -10,4 +10,6 @@
   | grep -v " not found or unable to stat" \
   | grep -v " Graceful restart requested, doing restart" \
   | grep -v " resuming normal operations" \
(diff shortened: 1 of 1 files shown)


cgolubi1:
For future reference, two signatures i didn't get in this round:

```
[Mon Oct 07 18:50:43 2019] [error] [client X.X.X.X] Invalid URI in request GET ../../mnt/custom/ProductDefinition HTTP
[Tue Oct 08 07:37:12 2019] [error] [client X.X.X.X] Invalid method in request \x03
```

I'd rather not exclude these because i can imagine them (especially the second one) being caused by site bugs rather than doorknob-rattling, but if i'm still annoyed after this round of changes gets onto the various live sites, i will.

blackshadowshade:
Okay, this looks fine.

PR #2516, opened 2020-01-04
Delete password reset verification key after it is used

### LINKED ISSUE #2512: Password reset codes should only be usable once
In #2509, it was noticed that the password reset link continues to work in perpetuity, even after being used.

The ability to password reset using that link should expire after the link is used.

cgolubi1:
* Fixes #2512

### DIFF
--- a/src/engine/BMInterfaceNewuser.php
+++ b/src/engine/BMInterfaceNewuser.php
@@ -282,7 +282,6 @@ class BMInterfaceNewuser {
             }
 
             // Everything checked out okay.  Update the password
-            // create user
             $query =
                 'UPDATE player ' .
                 'SET password_hashed = :password ' .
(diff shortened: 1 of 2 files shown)


cgolubi1:
I tested this by hand, and added a unit test.  I think it's pretty straightforward and you probably don't need a dev site, but let me know if you want one anyway.

src/engine/BMInterfaceNewuser.php
```diff
@@ -299,6 +298,15 @@ public function reset_password($playerId, $playerKey, $password) {
             $statement->execute(array(':id' => $playerId,
                                       ':password' => $passwordHash));
 
+            // Now delete the verification key so it can't be reused
+            $query =
+                'DELETE FROM player_reset_verification ' .
+                'WHERE player_id = :player_id ' .
```
blackshadowshade: What about clearing ALL verification keys for this player when successfully using one?

blackshadowshade:
I'm happy to accept this pull request after my comment is addressed (either in the positive or the negative).

src/engine/BMInterfaceNewuser.php
cgolubi1: Okay.

blackshadowshade:
Looks good. I'll accept this when CircleCI is done.

Issue #2469, opened 2019-08-03
Fill in FAQ page


blackshadowshade:
Spawned from #1924.

Fill in the contents of the FAQ page faq.html

blackshadowshade:
Ideas taken from #137 that haven't yet made it to the FAQ page:

- Why can't I find a particular button? (with regard to unimplemented skills and button specials)
- Why won't the game let me pass, even though I have no valid attack?
- What does it mean to adopt a button? How do I adopt a button?
- How do I play in a tournament?

blackshadowshade:
#2460 is all about adding help to how to do forum links.

blackshadowshade:
And related to the second point: How do I create and test a custom recipe?

scipiohaereticus:
Just something I noticed: The example link in the FAQ "How do I link to another forum post?" is not working. I think the reason is that the "Id" in "threadId" and "postId" is not capitalized.

blackshadowshade:
TheOrgg mentioned that "How do I use Fire dice?" would be a worthy question to add to the FAQ, and I agree.

PR #2534, opened 2020-02-09
Added FAQ text

blackshadowshade:
Addresses part of #2469.

### DIFF
--- a/src/ui/gui.css
+++ b/src/ui/gui.css
@@ -1030,3 +1030,8 @@ div.forum div.splitRight span.postTag {
 #zoom {
   background-color: green;
 }
+
+/* HELP PAGES */
+.help {
+  text-align: left;
+}
--- a/src/ui/js/Faq.js
+++ b/src/ui/js/Faq.js
@@ -43,27 +43,446 @@ Faq.bodyText = function() {
 };
 
 Faq.generalInfo = function() {
-  var text = $('<div>');
+  var text = $('<div>').addClass('help');
 
   text.append(
     $('<h1>').text('Frequently asked questions')
   );
 
-  text.append(
-    $('<p>').html(
-      'Content coming soon!'
-    )
-  );
+  text.append(Faq.tableOfContents());
 
-  text.append(
-    $('<p>').html(
-      'To encourage us to work on this page or to contribute yourself, ' +
-      'see the ' +
-      '<a href="https://github.com/buttonmen-dev/buttonmen/issues/2469">' +
-      'relevant issue</a> ' +
-      'in our Github issue tracker.'
-    )
-  );
+  text.append(Faq.content());
 
   return text;
+};
+
+Faq.tableOfContents = function() {
+  var toc = $('<ul>');
+
+  toc.append(Faq.gameplayLinks());
+  toc.append(Faq.userprefsLinks());
+  toc.append(Faq.unansweredLink());
+
+  return toc;
+};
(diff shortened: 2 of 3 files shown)


cgolubi1:
I stood up a dev site for this.  Irilyth, when you get a second, please sync up 54.89.147.252 == blip.dev.buttonweavers.com

cgolubi1:
Testers, the new FAQ page is http://54.89.147.252/ui/faq.html

cgolubi1:
I want to flag the use of the `createDocumentFragment` function.  We don't use it anywhere else.  Does it have any side effects or implications we need to care about?

I'm particularly interested because the new page is accessible without logging in, so i want to make sure the new page is very boring in terms of what it talks to.  It doesn't make any API calls at all, right?

blackshadowshade:
The createDocumentFragment function is a standard Javascript function. It creates a placeholder node that is removed when the fragment is attached to something else.

https://www.w3schools.com/jsref/met_document_createdocumentfragment.asp

On this link (point 5), it even extols the virtues of using DocumentFragments:

https://www.toptal.com/javascript/10-most-common-javascript-mistakes


And yes, you are correct that there are no API calls.

irilyth:
> I stood up a dev site for this. Irilyth, when you get a second, please sync up 54.89.147.252 == blip.dev.buttonweavers.com

Done!

```
+$ host blip.dev.buttonweavers.com
blip.dev.buttonweavers.com has address 54.89.147.252
```

cgolubi1:
Playtesters, feedback on http://blip.dev.buttonweavers.com/ui/faq.html ?  Any obvious things that should be added/removed/fixed before we call this a good start?

dwvanstone:
* I might suggest changing:

"If you select the trip attack, the (8):5 rerolls and might not even be captured."

to

"If you select the trip attack, both dice reroll and the (8):5 might not even be captured."

* "The Flying Squirrel has a [button special](http://blip.dev.buttonweavers.com/ui/faq.html#ButtonSpecial) that says that it can't do Skill attacks. Too bad. At least it has Speed dice."

Could the link on "button special" go The Flying Squirrel's button page instead so the person can read the text describing the specialness? 

* "The Japanese Beetle has a [button special](http://blip.dev.buttonweavers.com/ui/faq.html#ButtonSpecial) that says that it can't be attacked by Skill attacks. Too bad. Not much you can do."

Same with this page. Also perhaps "Not much you can do." could be, instead, "At least he only has four dice."

Great! It's nice to see this started!!!

blackshadowshade:
What about having the name "The Flying Squirrel" and "The Japanese Beetle" link to their respective button pages?

cgolubi1:
> What about having the name "The Flying Squirrel" and "The Japanese Beetle" link to their respective button pages?

Sounds fine to me.

blackshadowshade:
Changes made as requested. We need to check that I have the correct relative path to the button pages when this has been updated.

cgolubi1:
Updated the dev site.

dwvanstone:
The dev site looks good! Yay!

blackshadowshade:
The links from The Flying Squirrel and The Japanese Beetle work correctly, so yay from me!

cgolubi1:
Sounds like this is ready to merge.  Shadowshade, want to rebase it when you get a chance?

blackshadowshade:
Rebased.

PR #2570, opened 2020-06-13
Made anchors that are in hidden spoilers unclickable

### LINKED ISSUE #1153: Figure out how to handle links inside of spoiler tags
At present, if there's a link (e.g. to a game or player profile) inside of a spoiler tag, the link won't be visible as such (since the spoiler is blacked out), the mouse cursor will display as a normal arrow when you mouse over it, and it would be very easy to accidentally follow the link when you're only intending to click on the spoiler to reveal it.

We should decide how we want to handle this. 

One option is to try to override the click, so that clicking on a link inside of concealed spoiler _only_ reveals the spoiler without activating the link. But the problem with that is that if you reveal the spoiler by just highlighting it (as I expect many people normally would, since they wouldn't even know they're supposed to click), you'd never be able to actually use the link. 

Any thoughts on this?

blackshadowshade:
Fixes #1153.

### DIFF
--- a/src/ui/gui.css
+++ b/src/ui/gui.css
@@ -930,11 +930,19 @@ div.forum div.splitRight span.postTag {
 }
 
 .chatSpoiler:not(.chatExposedSpoiler),
-.chatSpoiler:not(.chatExposedSpoiler) a,
 .chatSpoiler:not(.chatExposedSpoiler) .chatQuoteBody {
     color: black;
     background-color: black;
     cursor: default;
+    pointer-events: auto;
+}
+
+/* make spoiler links not clickable until they are exposed */
+.chatSpoiler:not(.chatExposedSpoiler) a {
+    color: black;
+    background-color: black;
+    cursor: default;
+    pointer-events: none;
 }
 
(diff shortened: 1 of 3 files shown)


cgolubi1:
Dev site: http://temporary.dev.buttonweavers.com

cgolubi1:
Hi testers - this dev site http://temporary.dev.buttonweavers.com is for testing Shadowshade's change to make anchors that are in hidden spoilers unclickable.  Can someone check it out?

irilyth:
http://temporary.dev.buttonweavers.com/ui/forum.html#!threadId=9 seems to work! If I click on the part of the spoiler bar where the "Game 1" text is, it just clears the spoiler tag. If I click the "Game 1" text once it's exposed, it goes to the game.

dwvanstone:
This looks fantastic! Yay!

cgolubi1:
Awesome, thanks irilyth and dwvanstone (and thanks, Shadowshade, for fixing this - it's been annoying me too).  Merging.

Issue #1548, opened 2015-02-25
Warrior dice should be greyed out for the defender


blackshadowshade:
As requested by irilyth and devious, it should be more visually obvious that warrior dice cannot be targeted.

blackshadowshade:
The question for me is how much of this should be implemented in the back end, and how much in the front end.

The back end already supplies a full list of skills to the front end, and it is easy enough to do the logic in either place to identify warrior dice for the defender.

Does anyone have any preferences?

cgolubi1:
I don't really have a lot of preferences.  Right now, the code in Game.js which handles "Dizzy" dice is pretty specific to the dizzy/Focus combination.  So it doesn't strike me as any *less* general to have the UI identify "opponent's warrior" dice.  The backend has to tell the UI which particular focus dice are dizzy, but in the case of opponent's warrior dice, the UI already has all the information it needs.

So i think it would be simplest to do it all in the UI, and what you'd do would be to add analogous code in `Game.gamePlayerDice()` similar to what exists for dizzy/Focus.  (Look for both "Dizzy" and "dizzy" in that function.)  Should be straightforward.

irilyth:
Would it be easy to sort the Warrior dice all to one side (probably the left)? I think that'd be advantageous.

blackshadowshade:
I don't know if it's a good idea, even if it were an easy one. I would find it confusing, I think, if the order of the dice changed each round.

cgolubi1:
> Would it be easy to sort the Warrior dice all to one side (probably the left)? I think that'd be advantageous.

I'm sure it wouldn't be impossible, but it would be new logic that doesn't exist right now, whereas making them grey and unclickable is just like the existing focus case.  IMO we should try that, and if we decide we also want "off to one side", make a separate ticket.

Also, frankly, i know not everything is about us, but i think it would make bug analysis substantially more time-consuming for me if i couldn't make the assumption that the order in which the dice appear in the UI corresponded to the index numbers transmitted to the server.  Obviously, that shouldn't be an overwhelming factor, but nonetheless i'm reluctant to do something that makes debugging more painful given how much time i spend debugging.  :>)

irilyth:
Hmm, ok. I don't think it's a big deal, but it is how the old site worked, and it was pretty nice there. But we might as well try one thing at a time, and it's not worth a whole lot of disruption to change.

irilyth:
[...]

PR #2573, opened 2020-06-26
Visually distinguish warrior dice when they are active

### LINKED ISSUE #1606: Differentiate your own out-of-play Warrior dice from in-play dice
On #1548, I'd said

> What about your own Warrior dice? I think it'd be nice if they were also more clearly out-of-play than just the ` mark, but they surely shouldn't be the same gray as dizzy Focus dice. Maybe a pattern rather than a color, like a striped background or something?

I still think that'd be a good idea; especially dice with _other_ skills, some of which look like a ` mark (e.g. IIconfused T&S die), it's sometimes confusing to have them intermingled and undifferentiated.

blackshadowshade:
Fixes #1606.


blackshadowshade:
And there we go, got the QUnit tests fixed too. Now CircleCI is happy.  :)

cgolubi1:
I stood up a new dev site, http://blip.dev.buttonweavers.com/ui/ for testing this change to add a shield symbol to warrior dice when they are active, addressing an issue that @nkmcalli and @irilyth have both recently commented on.  Testers, please check it out.

dwvanstone:
The shield symbol looks great and seems to function well if the die background is a circle, but it doesn't look so good if the die background is symmetric or realistic.

Screen Shot 2020-06-29 at 11 56 17 AM
Screen Shot 2020-06-29 at 11 57 00 AM

Also, after I do an attack, the shields disappear from the warrior dice which seems confusing. I'd expect the warrior shield to stay on the die even if it's not my turn.

nkmcalli:
I added new feedback to the original issue.  I think the shield symbol is better than nothing, but I don't think it's the right solution for this problem.

blackshadowshade:
I've updated the code to generate a stripe instead of the shield symbol.

If testers are happy with this, I'll update the help page too.

blackshadowshade:
@cgolubi1, this dev site can be updated with the new code.

cgolubi1:
Updated!

cgolubi1:
What's the status here?  I see some suggestions on #1606 from a couple of weeks ago, but it looks like the conversation petered out.

blackshadowshade:
I think we should deploy the site without waiting for this feature, it looks like we're not yet at an endpoint that people agree on.

cgolubi1:
I updated this dev site with the latest CSS.  Check it out.

blackshadowshade:
@nkmcalli, @irilyth, take a look at the dev site, tell us what you think!

irilyth:
Not bad! Did we conclude that it was undesirable and/or difficult to keep all the Warrior dice always to the right of all the in-place dice, including moving a Warrior die left when it came into play?

Ah, that was here: https://github.com/buttonmen-dev/buttonmen/issues/1548#issuecomment-76104231:

> I'm sure it wouldn't be impossible, but it would be new logic that doesn't exist right now, whereas making them grey and unclickable is just like the existing focus case. IMO we should try that, and if we decide we also want "off to one side", make a separate ticket.

So, ok, let's continue to leave that aside for now.

dwvanstone:
This looks good with all of my testing.

cgolubi1:
I'll give this another day for nkmcalli to weigh in --- based on other comments, it seems like we're good to go.

cgolubi1:
Merging!  Thanks for all the iteration, everyone.

cgolubi1:
Sigh, of course that didn't get rebased before i merged it.  Oh well --- it happens.  Apologies, future us.

PR #2578, opened 2020-07-09
Added button art for Blademasters

### LINKED ISSUE #2545: Found art for the first Blademasters set
Here's the art for Blademasters 1.  (one name edited so it can be read; dark blue on black made it unreadable in the scans from the Buttonmen book)

BlademastersArashi
BlademastersDanielle
BlademastersDanny
BlademastersHorace
BlademastersInez
BlademastersJohnny
BlademastersKainar
BlademastersMacky
BlademastersMagistra
BlademastersMichie
BlademastersRenee
BlademastersScott

blackshadowshade:
Fixes #2545.


blackshadowshade:
With these new images, I'm seeing the following page for Blademasters:

blademasters_all_images

Since this pull request only adds image art, CircleCI shouldn't complain at all, so this is ready to go up for testers to take a look at.

TheOrgg:
Should I swap the sixes on Johnny to make the art and stats match?  Is there a standard formatting for two identical dice with abilities?

blackshadowshade:
There is no standard recipe order, if that's what you're wondering.

I'm not too fussed by the order of the sixes, but we'll see what others say.

dwvanstone:
My preference would be to either change the art to match the recipe. Overall looks great!

TheOrgg:
BSS, how hard is it to make the first 6 the trip instead of the second one?  That feels like it might be quicker.

blackshadowshade:
It's not difficult, just a database update. I'll get onto that now.

blackshadowshade:
And the database update is done.

cgolubi1:
shadowshade: tell me if i'm just misreading github's diff here, but it looks like the DB update file can't be right, because it seems to be a noop, and we'll certainly need an update to change the button's recipe.

blackshadowshade:
Hmmm, I believe that you're right. Let me fix that.

blackshadowshade:
Okay, that's been fixed and rebased.

cgolubi1:
Thanks, that's better.  Getting this on a dev site now.

cgolubi1:
dev site up - http://fleeting.dev.buttonweavers.com/ui/buttons.html?set=Blademasters - please check this out and give me a thumbs-up to merge this, someone.

blackshadowshade:
@TheOrgg, want to take a look at this and see if it looks right to you?

dwvanstone:
I think it all looks good.

Question: if we're changing the order of the dice (such as Johnny which goes from (6) t(6) I(8) (12) (X)   to  t(6) (6) I(8) (12) (X) ) ... will this change any existing partially complete games with Johnny? I.e. if the "first die" is captured in a game will that become the t(6) which was captured vs the (6) which was captured?

TheOrgg:
Looks right to me.

cgolubi1:
> Question: if we're changing the order of the dice (such as Johnny which goes from (6) t(6) I(8) (12) (X) to t(6) (6) I(8) (12) (X) ) ... will this change any existing partially complete games with Johnny? I.e. if the "first die" is captured in a game will that become the t(6) which was captured vs the (6) which was captured?

I think it won't be a problem, because when games start, the initial button recipe is copied into the game, and used that way in the game henceforth.  That's why i didn't flag it when i stood up the dev site.  Someone can do a smoke test to make sure --- i'm sufficiently confident that at a high level this functionality has to work (because auxiliary dice, reserve dice, RandomBM, etc), that i'd be happy with a smoke test on staging, i.e. why don't a couple of people go to staging now and start games with Johnny, and we can check on them at that point in the release?

dwvanstone:
I've accepted game 185 and have created three other test games (186-188) in various stages of capturing the first two dice. Ready for later testing!

cgolubi1:
Great, thanks all.  Merging.

PR #2509, opened 2019-12-07
Add password reset interface and API

cgolubi1:
* Fixes #141

PR #2580, opened 2020-07-10
Added skills description page

### LINKED ISSUE #1925: Add a page listing all die skills
There should be a page that lists all the supported die skills, their symbols and their descriptions.

(Created in response to #1924.)

blackshadowshade:
Fixes #1925.

This builds on the code of @AdmiralJota and the tweaks by @dwvanstone.

### DIFF
--- a/src/api/ApiResponder.php
+++ b/src/api/ApiResponder.php
@@ -22,10 +22,16 @@ class ApiResponder {
     protected $isTest;
 
     /**
-     * functions which allow access by unauthenticated users
-     * For now, all game functionality should require login: only
-     * add things to this list if they are necessary for user
-     * creation and/or login.
+     * Functions which allow access by unauthenticated users.
+     *
+     * All game functionality should require login.
+     *
+     * Only add things to this list if they are necessary for user
+     * creation and/or login, or for any content not requiring
+     * login, like help.
+     *
+     * Help functionality should be designed to make NO database
+     * calls because it is accessible without login.
      *
      * @var array
      */
@@ -36,6 +42,8 @@ class ApiResponder {
         'login',
         'forgotPassword',
         'resetPassword',
+        'loadDieSkillsData',
+        'loadDieTypesData',
     );
 
     /**
@@ -918,6 +926,26 @@ class ApiResponder {
     ////////////////////////////////////////////////////////////
 
     /**
+     * Interface redirect for loadDieSkillsData
+     *
+     * @param BMInterface $interface
+     * @return NULL|array
+     */
+    protected function get_interface_response_loadDieSkillsData($interface) {
+        return $interface->help()->load_die_skills_data();
+    }
+
+    /**
+     * Interface redirect for loadDieTypesData
+     *
+     * @param BMInterface $interface
+     * @return NULL|array
+     */
+    protected function get_interface_response_loadDieTypesData($interface) {
+        return $interface->help()->load_die_types_data();
+    }
+
+    /**
      * Interface redirect for response_logic
      *
      * @param BMInterface $interface
@@ -1095,4 +1123,5 @@ if (!function_exists('apache_note')) {
         }
         return $note_value;
     }
+
 }
--- a/src/api/DummyApiResponder.php
+++ b/src/api/DummyApiResponder.php
@@ -413,6 +413,19 @@ class DummyApiResponder {
     // End of Forum-related methods
     ////////////////////////////////////////////////////////////
 
+    protected function get_interface_response_loadDieSkillsData() {
+        return $this->load_json_data_from_file(
+            'loadDieSkillsData',
+            'noargs.json'
+        );
+    }
+
+    protected function get_interface_response_loadDieTypesData() {
+        return $this->load_json_data_from_file(
+            'loadDieTypesData',
+            'noargs.json'
+        );
+    }
 
     // Ask get_interface_response() for the dummy response to the
     // request, then construct a response.  Match the logic in
--- a/src/api/ApiSpec.php
+++ b/src/api/ApiSpec.php
@@ -311,6 +311,14 @@ class ApiSpec {
             'mandatory' => array(),
             'permitted' => array(),
         ),
+        'loadDieSkillsData' => array(
+            'mandatory' => array(),
+            'permitted' => array(),
+        ),
+        'loadDieTypesData' => array(
+            'mandatory' => array(),
+            'permitted' => array(),
+        ),
         // loadForumBoard returns:
         //   boardId: int,
         //   boardName: string,
--- a/src/engine/BMDieOption.php
+++ b/src/engine/BMDieOption.php
@@ -197,15 +197,23 @@ class BMDieOption extends BMDie {
         $typesList = array();
         $typesList['Option'] = array(
             'code' => '/',
-            'description' =>
-                'Option Dice are represented as two numbers with a slash ' .
+            'description' => self::getDescription(),
+        );
+        return $typesList;
+    }
+
+    /**
+     * Get description of option dice
+     *
+     * @return string
+     */
+    public static function getDescription() {
+        return  'Option Dice are represented as two numbers with a slash ' .
                 'between them, resembling a fraction. They function like ' .
                 'Swing Dice and can be changed at any time a Swing Die could ' .
                 'be changed. However, Option Dice are restricted to only two ' .
                 'values. For example, a 4/10 Option Die can be only a 4 or a ' .
-                '10.',
-        );
-        return $typesList;
+                '10.';
     }
 
     /**
--- a/src/engine/BMDieSwing.php
+++ b/src/engine/BMDieSwing.php
@@ -71,7 +71,7 @@ class BMDieSwing extends BMDie {
      *
      * @var array
      */
-    private static $swingRanges = array(
+    public static $swingRanges = array(
         "R" => array(2, 16),
         "S" => array(6, 20),
         "T" => array(2, 12),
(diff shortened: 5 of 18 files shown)


blackshadowshade:
Here's a (zoomed-out) screenshot of what I'm seeing:

skills_page

dwvanstone:
OOOOH! I'm very excited about this page. Thank you for fulfilling all of our dreams!

TheOrgg:
I assume this'll be updated as new abilities are functional?

blackshadowshade:
Die skills will be auto-updated, since the code reads from the list of skills in BMSkill.

Die types (like Wildcard) need to be added explicitly to the code.

However, all changes in skill/type descriptions will update automatically, since this code loads the descriptions from the base BMSkill* and BMDie* classes.

cgolubi1:
Okay, so there's definitely a test thing here: if you add new APIs, you have to add responder tests for the API, and the JS tests should use the dummy results that come from the responder tests, not hardcode their own responses.  The purpose of that is to thread together those things in an integration-test-y way, so that if we change the API we'll know, and so that if we change the API in a way that breaks the UI, we'll know that too.

If you've never done this before, say something, and i'll try to come up with a better template for it, or at least a recent-ish PR that does it.

cgolubi1:
Anyway, let's also have a dev site --- check it out folks: http://temporary.dev.buttonweavers.com/ui/skills.html

blackshadowshade:
If I understand correctly, you're referring to the hardcoded set of Api.dieSkills and Api.dieTypes. I can see now how that circumvents responder testing.

If you wouldn't mind referring me to a relevant PR, I'll see what I can do to put that right.

blackshadowshade:
@AdmiralJota, take a look at this and tell us what you think!

cgolubi1:
Sure, the last API i added was for password reset, in #2509 - the relevant set of files changed in that PR is:

```
src/api/DummyApiResponder.php
test/src/api/responder00Test.php
test/src/ui/index.html
test/src/ui/js/test_ForgotPassword.js
test/src/ui/js/test_VerifyReset.js
test/src/ui/phantom-index.html
```

(Some of those you already have covered, but definitely not the first two, and the JS tests should show using the dummy data from the responder tests.)

dwvanstone:
The dev site looks fantastic. Well done!

TheOrgg:
Looks good to me, too!

blackshadowshade:
@cgolubi1, I've added API tests for die skills and types.

The last thing remaining that I'm stuck on is how to acquire the API response in test_Skills.js to fill Api.dieSkills and Api.dieTypes. How do I do this?

irilyth:
Did we want to separate skills and types in some visual way?

blackshadowshade:
So, we've separated them out visually by using the brackets.

If we want separate tables, that would be an extra visual distinction, but I'm not sure it's strictly necessary---or helpful.

blackshadowshade:
It *might* be possible to add a line break or something equivalent between the die skills and types. @AdmiralJota, how would we go about doing that in a clean way?

src/api/DummyApiResponder.php
```diff
@@ -413,6 +413,19 @@ protected function get_interface_response_editForumPost($args) {
     // End of Forum-related methods
     ////////////////////////////////////////////////////////////
 
+    protected function get_interface_response_loadDieSkillsData($args) {
+        return $this->load_json_data_from_file(
+            'loadDieSkillsData',
+            $args['threadId'] . '.json'
+        );
+    }
+
```
cgolubi1, line 422: Too much copy-paste.  There isn't going to be a 'threadId' arg, so this is going to give you a file with a useless name.  Try copying from an API that doesn't take args, e.g. `get_interface_response_loadOpenGames` is a good one to parrot.

cgolubi1:
shadowshade: with the caveat that i haven't done this in awhile, i *think* the `getOpenGames` API method is going to continue to be a good thing to parrot.  For instance, `test_OpenGames.getOpenGames` *looks* to me like it calls the API method, and then does stuff with the results.

The way this stuff works is the tests are calling a real API, just using the endpoint dummy_responder rather than responder, and which is handing you results from JSON files, or an error if there's no such JSON file (which is why you need to use a sensible JSON filename in DummyApiResponder.php), rather than invoking PHP logic.

blackshadowshade:
Okay, I've connected up the API. Now, I'm having trouble testing because I don't know how to get my tests to run after the page load and associated async calls have completed.

Ideally, I'd like something like:

    stop();
    Skills.showLoggedInPage();
    waitUntilPageLoadIsFinished();
    runMyTests();
    start();

I've tried

    Env.callAsyncInParallel([
      Skills.showLoggedInPage,
    ], function() {
      runMyTests();
    });

but this hangs after running Skills.showLoggedInPage() and never proceeds to the callback.

How do I get this working?

cgolubi1:
Compare:

```js
# OpenGames.js
OpenGames.showLoggedInPage = function() {
  // Get all needed information, then display Open Games page
  OpenGames.getOpenGames(OpenGames.showPage);
};

OpenGames.getOpenGames = function(callback) {
  Env.callAsyncInParallel([
    Api.getOpenGamesData,
    { 'func': Api.getButtonData, 'args': [ null ] },
  ], callback);
};

# Skills.js
Skills.showLoggedInPage = function() {
  Env.callAsyncInParallel([
    Api.getDieSkillsData,
    Api.getDieTypesData,
  ], Skills.showPage);
};
```

If you implement Skills.js the way OpenGames.js is, with an intermediate getter function that takes showPage as a callback, then you'll be able to do the testing the same way.  LMK if i'm missing something; i might be, it's been awhile.

blackshadowshade:
I've got the following now:

    Skills.showLoggedInPage = function() {
      // Get all needed information, then display Skills page
      Skills.getInfo(Skills.showPage);
    };

    Skills.getInfo = function(callback) {
      Env.callAsyncInParallel([
        Api.getDieSkillsData,
        Api.getDieTypesData,
      ], callback);
    };

That doesn't solve the problem, though.

For example, here's the first test:

  test("test_Skills.showLoggedInPage", function(assert) {
    stop();
    BMTestUtils.setupFakeLogin();

    Skills.showLoggedInPage();
  //  setTimeout(function() {

      var item = document.getElementById('skills_page');
      assert.equal(item.nodeName, "DIV",
        "#skills_page is a div after showLoggedInPage() is called");

      BMTestUtils.cleanupFakeLogin();
      start();
  //  }, 1000);
});

It passes with the timeout, but fails without it, with this error message: "All test functions MUST complete jQuery activity before exiting", expected 0, result 2. That's obviously due to the two async requests sent for die skill and type info.

There's still something missing, but I don't know what.

cgolubi1:
Hmm, does the comment on the analogous `test_OpenGames.showLoggedInPage` help?

blackshadowshade:
Hmmm, so we're not actually testing the full functionality of showLoggedInPage, just the flow of what goes on inside it? I can do that.

cgolubi1:
> Hmmm, so we're not actually testing the full functionality of showLoggedInPage, just the flow of what goes on inside it?

Yep, that's right.

blackshadowshade:
In that case, the tests should now be complete.

AdmiralJota:
> It _might_ be possible to add a line break or something equivalent between the die skills and types. @AdmiralJota, how would we go about doing that in a clean way?

I'm fine with the way it is, but if you wanted to split them, the easiest way would probably be to change the Skills.directoryDiv() function so it took two arguments: headingText and displayList. Then change line 64 to `directoryDiv.append($('<h2>', { headingText }));)` and remove lines 66-70.

Then when we call Skills.directoryDiv on line 51, call it twice: once with `Skills.directoryDiv('Die Skills', Api.dieSkills.list)` and once with `Skills.directoryDiv('Die Types', Api.dieTypes.list)`.

blackshadowshade:
Makes sense, @AdmiralJota. So, should I make this change to split the codes up into two tables? Or does that just make the display more complicated?

irilyth:
I think it's ok as it is, but splitting it up would be an improvement.

blackshadowshade:
Right, we're going to leave it as-is for this pull request. If anyone expresses strong feelings about this after release, we can always adjust the look and feel then.

src/api/ApiResponder.php
```diff
@@ -22,10 +22,13 @@ class ApiResponder {
...
-     * For now, all game functionality should require login: only
-     * add things to this list if they are necessary for user
-     * creation and/or login.
+     * Functions which allow access by unauthenticated users.
+     *
+     * All game functionality should require login.
+     *
+     * Only add things to this list if they are necessary for user
+     * creation and/or login, or for any content not requiring
+     * login, like help.
```
cgolubi1, line 31: You say somewhere deeper in the code that help functionality should be designed not to make database calls.  Can you say that in this comment too, as a reminder to Future Us to limit DB-using functionality accessible to non-logged-in users?

cgolubi1:
My read of the comments is that everyone likes or is willing to tolerate this the way it is.

I want to flag that you're adding a new set of APIs which is accessible without login; i'm fine with it, but as i said in my single comment, i think we should be more explicit in ApiResponder about what the guiding principles are, for the future.  If they're "only API calls which A. don't access the DB, or B. which are necessary to get to login, are visible without logging in," i am fine with that principle, and we should say it explicitly in one place in a comment (and i think ApiResponder is the right place).

Want to modify that comment and then rebase this down to one commit?

blackshadowshade:
Comment updated, rebase done.

cgolubi1:
Thanks, merging.  We definitely have enough new stuff to do a release, so the next time i work on BM, i'll get started on that.

PR #2590, opened 2020-07-28
On AWS instances, emit cloudwatch metric entries recording API call success

cgolubi1:
This is a very simple implementation of site health metrics using cloudwatch:
* On non-AWS sites, this does nothing
* On AWS sites, it installs aws-cli (used to record metrics)
* Every five minutes, it:
    * Uses logtail to find new entries in the apache access log
    * Looks for entries which are POSTs to /api/responder
    * Parses out from each entry what the API call was, and whether it succeeded
    * Emits a metric whose value is the success (1 on success, 0 on failure), and whose dimensions are the instance ID (which effectively tells us what buttonmen site it is) and the API call in use (just in case this is interesting)

This should be enough to allow us to alarm on whether a site hasn't seen any traffic in awhile, and on whether call success rate is low, whatever proves useful to us.

### DIFF
--- a/deploy/vagrant/manifests/init.pp
+++ b/deploy/vagrant/manifests/init.pp
@@ -59,4 +59,11 @@ node default {
   include "php::base"
   include "mysql::server"  
   include "buttonmen::server"
+
+  # location-specific configuration
+  case "${ec2_services_partition}" {
+    "aws": {
+      include "cloudwatch::buttonmen-site"
+    }
+  }
 }
--- a/deploy/vagrant/modules/cloudwatch/manifests/init.pp
+++ b/deploy/vagrant/modules/cloudwatch/manifests/init.pp
@@ -1,0 +1,20 @@
+class cloudwatch::buttonmen-site {
+  package {
+    "awscli": ensure => installed;
+  }
+
+  file {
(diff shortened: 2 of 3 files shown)


cgolubi1:
I deployed this on an AWS instance, and ran it while reloading the UI a couple of times (which calls loadPlayerName), failing to login a couple of times, and successfully creating a couple of users.  Here's what that looks like in the cloudwatch console:

graph_of_metrics

blackshadowshade:
I'm happy to give this a go.

Just a question: Who has access to see these Cloudwatch metrics?

cgolubi1:
By default, only me.  I can find a way to export a dashboard if we think that's a useful thing to do, or i can give other devs access to view cloudwatch metrics in my account, but what i really had in mind was to put alarms on the metrics, so that we'd get an e-mail to one of the dev lists if a metric went pear-shaped on prod or staging.  That would give us advance notice of outages (and that's something i'll just setup offline via the AWS console, no need to wait for another release).  And the existence of AWS cli gives me the ability to either use these metrics or my own new ones to alert myself if replay sites fall over, which is the other major use case i had in mind.

PR #2592, opened 2020-08-02
Fixed PHP warning caused by incorrect bracketing

### LINKED ISSUE #2591: Missing bracket in BMUtilityXCYIterator triggers PHP warnings
While investigating #2589, I noted that the following PHP warning was being thrown in the logs:

    PHP Warning:  count(): Parameter must be an array or an object that implements Countable in /Volumes/Development/buttonmen/src/engine/BMUtilityXCYIterator.php on line 101

This is caused by incorrect bracketing in the code.

blackshadowshade:
Fixes #2591.

I *think* I understand why this hasn't caused any trouble, but I'm not entirely sure. My thinking is the following:

The bad code is

    if (count(array > 0) && ...) {...}

and the fixed code is

    if ((count(array) > 0) && ...) {...}

If you think about it,

    array > 0

produces an array of logicals, so

    count(array > 0)

is just the same as

    count(array)

and since this is in an 'if' statement, it gets converted to a logical, meaning that it's equivalent to

    count(array) > 0

However, if I could have this wrong.

Anyway, I'll let the unit tests have the last word.

### DIFF
--- a/src/engine/BMUtilityXCYIterator.php
+++ b/src/engine/BMUtilityXCYIterator.php
@@ -98,7 +98,7 @@ class BMUtilityXCYIterator implements Iterator {
         $this->tail = NULL;
 
         $this->head = array_pop($this->list);
-        if (count($this->list > 0) && $this->depth > 1) {
+        if ((count($this->list) > 0) && ($this->depth > 1)) {
             $this->tail = new BMUtilityXCYIterator($this->list, $this->depth - 1);
         }
         if ($this->tail) {


blackshadowshade:
And we have a pass on CircleCI:

https://app.circleci.com/pipelines/github/blackshadowshade/buttonmen/75/workflows/6d10b63b-db69-4fa3-88c3-72a6ce8285c4

cgolubi1:
Let's throw a replay loop at this.

cgolubi1:
3300 successful replay tests probably mean this is fine.  Merging.

PR #2596, opened 2020-08-11
use ec2_placement_availability_zone fact, because ec2_placement_region is not always available

cgolubi1:
When i pushed the changes to staging, i started getting these cron failure e-mails:

```
  File "/usr/local/bin/record_buttonmen_cloudwatch_metrics", line 8, in <module>
    REGION = sys.argv[2]
IndexError: list index out of range
```

on bmstaging:

```
$ facter ec2_placement_region

$ facter ec2_placement_availability_zone
us-east-1a
```

(vs on the dev site where i tested this code initially:

```
$ facter ec2_placement_region
us-east-1
$ facter ec2_placement_availability_zone
us-east-1c
```

So... fine.  Hand the AZ to the script and have the script chop it off.


blackshadowshade:
Wow. Yeah, let's get this into the codebase.

PR #2599, opened 2020-08-17
Remove debugging statements from cron jobs

cgolubi1:
I just made this change by hand on staging and prod because to do otherwise would be ridiculous, but here it is in the codebase.


blackshadowshade:
Fair enough.

PR #2605, opened 2020-09-05
Make bmsite cloudwatch metrics more user-friendly for production use

cgolubi1:
* Use boto instead of awscli so we're not spawning a process per metric
* Emit a sum per API call type, rather than one metric per individual call
* Emit a total metric and a failure metric, so we can look at load and at failures
* Emit a total, which will be zero if there are no calls in that interval
* Timeout the entire cron job if it's not done in 15 seconds

This PR is intended to address the problem with #2590 where it caused the prod site to bog down due to running awscli processes, and we had to disable the cron job.


cgolubi1:
This may help you visualize how these metrics look on my test site, or not.  Here is the same cloudwatch graph, showing the metric SUM, hovering over two different times --- the 16:50 UTC time period when there was activity (so we see the two totals and eight breakdown metrics for the various APIs that were invoked), and the 16:55 UTC time period when there was no activity (so we see zeroes emitted for the two total metrics, and no data emitted (shown as '-') for the eight breakdown metrics):

calls_with_activity

calls_with_no_activity

cgolubi1:
Reusing a previous branch wasn't a very smart idea --- i'm going to copy the changes here into a clean branch and submit a new identical pull with just those changes and not the old merge history.

Issue #2536, opened 2020-02-16
aws instances should emit cloudwatch metrics to help us figure out whether a site is up


cgolubi1:
The database disk filling issue was exacerbated because we found out about the problem via user reports, several hours after the fact.  It would be better if we could notify ourselves that the site wasn't working.

Proposed implementation: have each instance emit some kind of cloudwatch metric about successful and unsuccessful site responses in the log.  Then we can alarm on lack of successful responses.

cgolubi1:
Note: at a glance it seems like it might be interesting to alarm on error responses, but i'm not so interested in doing that, because a lot of error responses are likely because of player error (e.g. if you try to make an invalid attack, that'll show up as a failed API response, and nobody wants to get an alarm about that).  So i'm starting with the theory that the daily apache error log scans (which we already have) are the right way to find out about unknown bugs of the internal-error variety.

cgolubi1:
I did some experimenting with this a month or so ago, as a result of which i know how to make an IAM role that can emit cloudwatch metrics.  So i think what i need to do here is:
* Configure vagrant to assign that role to AWS instances it brings up
* Configure vagrant to install the awscli package
* Make a cron job that emits the actual metric
* Add alarms on the instances where we care about this metric (which may be only prod, since staging doesn't get regular enough use for a "did we get a success this hour" type metric to work well)

cgolubi1:
For the curious, here's a snapshot of the little dashboard i just made, with API success rates and call counts for prod and staging.  I'm pretty sure there's a way to export cloudwatch dashboards using IAM roles, so it might be that we could embed this in the site and everyone could see it, though that's definitely not a today project:

![bmapi_dashboard](https://user-images.githubusercontent.com/2898207/90403337-d4d56280-e06e-11ea-8c16-24fe5b607f73.png)

blackshadowshade:
Looking forward to seeing how this goes over the next couple of weeks.

cgolubi1:
Here's why i really wanted to get these metrics going --- this is the six replay sites i stood up yesterday (the master which records new games, and five actual test sites):

![replay_site_traffic](https://user-images.githubusercontent.com/2898207/92329044-e6d57000-f032-11ea-8696-2638bdba946c.png)

[...]

PR #2606, opened 2020-09-05
Make bmsite cloudwatch metrics more user-friendly for production use

cgolubi1:
* Use boto instead of awscli so we're not spawning a process per metric
* Emit a sum per API call type, rather than one metric per individual call
* Emit a total metric and a failure metric, so we can look at load and at failures
* Emit a total, which will be zero if there are no calls in that interval
* Timeout the entire cron job if it's not done in 15 seconds
* Actually set the IAM role for new AWS instances (so dev/replay instances can emit cloudwatch metrics at all)

This partially addresses #2536 and is the same change as #2605 (so you can see the graphs there for what the behavior looks like), just in a clean branch for git merge laziness reasons.

### DIFF
--- a/deploy/vagrant/Vagrantfile
+++ b/deploy/vagrant/Vagrantfile
@@ -63,6 +63,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
 
     # firewall settings
     aws.security_groups = ["buttonmen-dev"]
+    aws.iam_instance_profile_name = "ec2_instance_perms"
 
     aws.instance_type = "t1.micro"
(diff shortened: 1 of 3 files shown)


cgolubi1:
boto isn't smart enough on cross-region calls for this to do the right thing, so it doesn't work on replay sites.  Let me keep working on this.

cgolubi1:
Okay.  As far as i can tell, this is now working in us-east-1 and us-west-2, so it should be good to go.

blackshadowshade:
Okay, let's give this a go.

PR #2389, opened 2018-08-13
Allow custom recipes to be used

blackshadowshade:
Fixes #344. Fixes #2484. Puts in workarounds for #2614, #2625, #2636, #2637, #2675, #2676, and #2678. Fixes #2689.

This pull request allows the game creator to create a game with one or two CustomBM buttons with custom recipes. I've not done anything yet to make the custom recipes visible on the Open Games page. I also haven't changed the history page to exclude CustomBM buttons from stats.

~~I'm marking this pull request as incomplete for the moment because it doesn't seem to work correctly under Safari 11.1.2 (on Mac High Sierra 10.13.6). However, I'm not certain why this is, I'm guessing that there's some error in the Javascript that I've missed. I'd love some help tracking down what's going on here.~~

Edit: This now appears to work fine under Safari 12.0.1 (on Mac High Sierra 10.13.6).

As far as I can see, the code works fine under Firefox and Chrome.

Points to be addressed:
- [x] doubled skills:  ppp(X)
- [x] handling plasma skill syntax
- [x] large dice: (100), (10000)
- [x] negative sided dice: (-1)
- [x] decimal dice: (9.0), (9.1)
- [x] twin dice with different swing types: (X,Y)
- [x] invalid twin dice: (Z,p6), z(Y#Z,6)
- [x] twin dice with overly large components: (100,100)
- [x] compound dice with more than two components: z(Y,Y,Y)
- [x] twin combined with option: z(1/4,1/5), z(Z,1/4)
- [x] invalid option dice: (5/Z)
- [x] option dice with overly large components: (1/9999), (100/5)
- [x] invalid die types: (p6)
- [x] multiple auxiliary dice: (6) +(4) +(2)
- [x] long button recipes: (1) (1) ... (1)
- [x] invalid whitespace: z(1 , 2)
- [x] invalid twin syntax: (1,/3), (1/,3), (1/,3,5), (1,3,5/)
- [x] visibility of CustomBM recipe in hover over the name, and in other places
- [x] error text should be specific and propagate from where the error is caused
- [x] empty opponent recipe gives an internal error
- [x] games with bad recipes shouldn't start in the first place
- [x] restrict CustomBM to dev sites only
- [x] requesting button named "CustomBM" via API causes internal error
- [x] requesting custom button without custom recipe via API causes internal error
- [x] button containing only an auxiliary die shouldn't be allowed: +(2)
- [x] some maximum-length (150 character) button recipes
- [x] #2614 - Option die where one of the options is 0
[...]

PR #2612, opened 2020-09-12
Fix truthiness check on auxiliary and reserve adds to work with buttons whose first die is +/r

cgolubi1:
I think the branch name says it all.

Now that i've seen how easy it is to cast numbers at the API layer as part of addressing #712 i'm going to do it, but this PR addresses why my testing of #2389 keeps falling over.


blackshadowshade:
This only affects the Python stuff, good to go.

PR #2613, opened 2020-09-12
Cast all number arguments to int in ApiSpec, rather than one-off in ApiResponder

cgolubi1:
So this isn't all the work that's needed for #712 but i was wondering whether it would be easier to put up smaller PRs that touched smaller pieces of the codebase.  This particular piece touches only `src/api`, and does:
* Move all casts of elements in `$args` to numbers into `ApiSpec.php` (and in particular into one generic method within `ApiSpec.php` that casts all API arguments of type `number`)
* Make it so the only `(int)` casts remaining within `ApiResponder.php` are casts of elements of `$_SESSION`
* Add a regression test so that as we start stripping `(int)` casts from the codebase, we can't backslide

Future pieces will traverse `src/engine` and make that so that all `(int)` casts occur when pulling things from the database.

### DIFF
--- a/deploy/circleci/audit_php_files.php
+++ b/deploy/circleci/audit_php_files.php
@@ -28,6 +28,18 @@
 
   // Any other directories shouldn't contain PHP files
 
+  // In these directories, we limit the use of explicit (int) casts
+  $php_limit_int_cast_dirs = array(
+    "src/api",
+  );
+  $php_limit_int_cast_exceptions = array(
+    // exceptions for $_SESSION variables
+    "src/api/ApiResponder.php" => 2,
+    "src/api/api_core.php" => 1,
+    // exception for the API-spec-level validate and cast of all numeric args to int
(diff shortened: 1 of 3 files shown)


cgolubi1:
Shadowshade, obviously no rush on this given how long it takes me to iterate on logic refactoring types PRs, but i just wanted to make sure you know that afaic this is on your plate to say what you want to see before accepting it.  If i do related refactoring deeper in the stack, it will be as a separate PR --- this one is complete.

blackshadowshade:
Yeah, I've been busy this week. I'll get to it when I can.

blackshadowshade:
@cgolubi1, do I understand correctly that we're currently not calling sanitize_argument_of_type_number() anywhere? And if so, why not?

blackshadowshade:
Oh wait, I see it now. The call comes from the general sanitize function

blackshadowshade:
Right, this looks good to go.

Issue #2486, opened 2019-10-09
Option dice break when one of the options is a zero-sider


blackshadowshade:
The site throws an internal error if there I try to choose zero sides in a (0/4) die.

AdmiralJota:
If you make it (4/0) instead, does it through a Divide By Zero error?

jl8e:
I don't have the actual code to fix, but option dice set a min of one in init, and don't override it in set_optionvalue if it's a 0-sider. (Presumably need to set min to 1 explicitly if the chosen value is not 0, else it'll be rolling 0-max in at least turbo option situations.)

blackshadowshade:
Yeah, that makes a lot of sense. I figure that it shouldn't be *too* difficult to fix. However, I'm a little hesitant to be 100% certain about that because of #2417.

jl8e:
Both is always an option

PR #2628, opened 2020-10-03
Proof-of-concept: how would we remove casts from BMInterface.php?

cgolubi1:
Shadowshade, obviously don't merge this as written (among other things it definitely fails tests, though i think not any logic tests), i'm just putting it up to get your opinion about the approach.

This is an initial proposal for how we'd remove (int) and (bool) casts from BMInterface*.php.  The gist of it is:
* A new class, `BMDB`, that wraps database access and contains a small set of generic functions to do selects (and eventually also updates):
* A new exception class, so that if we want to be picky about the database select, e.g. "this query should return exactly one row", we have a way to communicate up "the database state wasn't what i expected"
* Within BMDB, the arguments to each function are the query to run, the parameters to pass, and the set of column names and *types* to expect in each row.  The types tell us how to cast the value we select, and they let us do the set of slightly-complicated-but-not-that-complicated things we do with DB types, of which the most interesting common example afaict is "this column contains either NULL or an int, and if it's an int i want to cast it to an int, but if it's NULL it should stay NULL"

* Within BMInterface and other classes that make database calls, we replace direct database connection executions with calls to these BMDB functions.

These are the advantages i see for BMInterface and friends:
* Because we're passing a parameter that says what type we expect each column to be, this provides inline documentation of what we expect the return values to look like, so when we return to code for later development, we don't have to guess.
* Because we know every parameter has always been cast to whatever type it's supposed to be cast to, we can vastly simplify the logic that brings these parameters into PHP game variables:
    * If something is a bool, we can just assign it to a parameter that's supposed to be a bool rather than futzing with select statements.
    * If something is int_or_null, the code does still have to care about those two cases to differentiate NULL and 0, because that's the way our codebase works so there's no easy out.  But what we don't have to do is stamp the if-not-NULL-cast-to-int pattern all over the codebase and worry about whether we got one wrong, because that's handled on select.  And we can know that if we got a value from a database and it fails isset(), it's precisely NULL and not some other non-set value.

To me, this seems like it would be more elegant and easier to deal with, in addition to giving us a path to resolving #712.  But you're the one who would mostly have to live with it, so i want to know what you think.

(And i'm also totally aware that you don't want me ripping out and rewriting a bunch of BMInterface before tournaments are merged.  I'm not going to do that.  I just wanted to put up a PR with the approach so you can tell me whether you like it.)

### DIFF
--- a/src/engine/BMDB.php
+++ b/src/engine/BMDB.php
@@ -1,0 +1,112 @@
+<?php
+/**
+ * BMDB: database interface to standardize types 
+ *
+ * @author: Chaos
+ */
+
+/**
+ * This class contains all the logic to do with selecting from and updating
+ * the buttonmen database tables.
+ *
+ * It is not intended to be instantiated --- it's a set of public functions
+ * that callers will invoke using their own database connection.
+ *
+ *  */
+class BMDB {
+    // properties
+
+    /**
+     * Connection to database
+     *
+     * @var PDO
+     */
+    protected static $conn = NULL;
+
+    /**
+     * Constructor
+     *
+     * @param PDO $conn
+     */
+    public function __construct($conn) {
+        self::$conn = $conn;
+    }
+
+    /**
+     * Execute a query which fetches a single value from the DB
+     *
+     * @param $conn
+     * @param string $query
+     * @param array $parameters
+     * @param array $returnType
+     * @return array
+     */
+    public function select_single_value($query, $parameters, $returnType) {
+        $statement = self::$conn->prepare($query);
+        $statement->execute($parameters);
+        $result = $statement->fetch(PDO::FETCH_NUM);
+        if (!$result) {
+            throw new BMDatabaseException("DB select_single_value found no result");
+        }
+        if (count($result) != 1) {
+            throw new BMDatabaseException("Expected 1 result from DB query, found " . count($result));
+        }
+        return ($this->cast_db_column($result[0], $returnType));
+    }
+
+    /**
+     * Execute a query which fetches an arbitrary number of rows from the database
+     *
+     * @param $conn
+     * @param string $query
+     * @param array $parameters
+     * @param array $columnReturnTypes
+     * @return array
+     */
+    public function select_rows($query, $parameters, $columnReturnTypes) {
+        $statement = self::$conn->prepare($query);
+        $statement->execute($parameters);
+        $rows = array();
+        while ($result = $statement->fetch(PDO::FETCH_ASSOC)) {
+            $row = array();
+            foreach ($columnReturnTypes as $column => $returnType) {
+                $row[$column] = $this->cast_db_column($result[$column], $returnType);
+            }
+            $rows[] = $row;
+        }
+        return $rows;
+    }
+
+    /**
+     * Cast a value fetched from a DB to the specified return type
+     *
+     * @param $conn
+     * @param string $query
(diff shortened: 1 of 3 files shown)


blackshadowshade:
I think it certainly makes sense to do all the type checking and casting in the one place, as opposed to scattered across multiple calls.

Code looks good on a first glance, but I'm still a bit sleep-deprived, so I'll take a closer look at this (and indeed the whole BMInterface code base) when we're closer to wanting to merge this.

cgolubi1:
Okay, cool; thank you.  I'll take that as a tentative green light for the approach, and think about whether there's a way to add the functionality and start adopting it that will stay out of the way of tournament code.  (The simple answer is, instead of starting by migrating BMInterface.php, start by migrating one of the other interface files that isn't also involved in tournaments.)

src/engine/BMDB.php
```diff
@@ -0,0 +1,112 @@
...
+ *
+ * @author: Chaos
+ */
+
+/**
+ * This class contains all the logic to do with selecting from and updating
+ * the buttonmen database tables.
+ *
+ * It is not intended to be instantiated --- it's a set of public functions
+ * that callers will invoke using their own database connection.
```
cgolubi1, line 13: This comment isn't true any more --- i changed my mind while writing the code, and now the class is instantiated with an open database connector.  IMO even better would be if we could eventually move the connector into this class, and just invoke it with `$isTest` or whatever.  That's because i don't like the thing where if a database connection temporarily fails (e.g. because the database is being restarted as part of a routine nightly restart), we throw exceptions to users.  The only reason we can't catch failures where `$conn` is undefined and retry connecting, is that the connection logic is so scattered.  So we might as well plan to have `BMDB` eventually own all of the connection management, because then it will be able to do friendly things like that.

blackshadowshade:
I'm currently looking at the 0-sided option die problem, and it's an easy fix. However, I'm hitting the string/int problem again.

The problem starts from BMInterfaceGame->submit_die_values($playerId, $gameId, $roundNumber, $swingValueArray, $optionValueArray), where neither $swingValueArray nor $optionValueArray are currently guaranteed to be arrays of int.

The obvious corollary is that BMDieOption->set_optionValue($optionValue) cannot assume that $optionValue is an int.

For most values of $optionValue, we don't really care. But, of course, for zero, we do care because we're trying to avoid casting NULL or '' to zero.

@cgolubi1, feel like weighing in on where we do this fix? Do we do it right at BMInterfaceGame->submit_die_values(), considering that these values don't hit the database (but rather alter a BMGame instance directly) before we try to use them?

blackshadowshade:
Linking to #2486 because of the previous comment.

cgolubi1:
I don't have time to look at the code right now, but this has to be a weakness in `ApiSpec.php`'s checking.  Presumably when a top-level item is a number, we cast that number to an int, but when a top-level item is an array of ints, we fail to cast each element to an int.  So the right fix is to improve `ApiSpec` and `ApiResponder` so that those casts are done in the array case too.  We shouldn't have to touch the guts of the system.

blackshadowshade:
Okay, that makes sense. I'll take a closer look at ApiSpec->sanitize_function_args(), that seems to be where we explicitly change types.

blackshadowshade:
I believe that this pull request can now be closed, since it's been superseded by #2674.

@cgolubi1, if you agree, go ahead and close this one.

PR #2583, opened 2020-07-12
Added functionality to cancel third-party games if you are the creator

blackshadowshade:
Fixes #2332.

I know that @irilyth will be especially pleased about this functionality.

PR #2630, opened 2020-10-03
Force $_SESSION['user_id'] to be int or NULL before it's passed to any game engine functions

cgolubi1:
* Partially addresses #712
* Should make the specific inconsistent behavior discussed in PR #2583 impossible to build

### DIFF
--- a/deploy/circleci/audit_php_files.php
+++ b/deploy/circleci/audit_php_files.php
@@ -40,6 +40,23 @@
     "src/api/ApiSpec.php" => 2,
   );
 
+  // In these directories, we limit explicit references to $_SESSION variables
+  $php_limit_sessionvar_dirs = array(
+    "src/api",
+    "src/database",
+    "src/engine",
+    "src/lib",
+  );
+  // Files in the above dirs which are allowed to contain $_SESSION variables at all
+  $php_limit_sessionvar_exceptions = array(
+    "src/api/api_core.php",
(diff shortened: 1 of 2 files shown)


cgolubi1:
This PR just adds a utility function to cleanly provide `$_SESSION['user_id']` as either an int or as NULL, and forces (via regression test) all `ApiResponder.php` functionality to use that function rather than directly access `$_SESSION['user_id']`.  Once this is merged, all functions in `BMInterface*` will be able to safely assume `$_SESSION['user_id']` (under whatever variable name it's been passed to them, typically `$playerId` i think) is int or NULL.

ApiResponder uses a couple of other session variables (`user_name` and `skipped_games`), the latter somewhat untidily, but i think those are not a major source of logic noise right now, so i'm not going to worry about them.

While i was at it, i made the regression test complain if we directly reference `$_SESSION` variables anywhere outside of the two `src/api` files that currently do so, just in case Future Us get any ideas about that.

cgolubi1:
Let me know if you'd like any testing done on this which i should setup (i.e. replay testing or a dev site), or if you're happy with circleci.

blackshadowshade:
I think this change makes a lot of sense. If it were to cause trouble, it'd do so in many many places, and that would certainly show up in automated testing.

We'll have a chance to take a look during testing on staging anyway.

PR #2641, opened 2020-10-18
Fixed zero-sided option die min value

### LINKED ISSUE #2486: Option dice break when one of the options is a zero-sider

### LINKED ISSUE #2614: Internal error when choosing an option die side count of 0
In a custom button game with `(22)` vs `(0/64)`, if the second player chooses an option value of `0` it yields an internal error.

Log message:

```
Caught exception in BMInterface::submit_die_values: Invalid die value:  is not between 1 and 0 for die (0/64)
```

blackshadowshade:
Fixes #2486. Fixes #2614.

~~The proposed fix in this pull request uses a non-typed comparison for the option die value. This will need to be strengthened once we are a bit further down the road with ensuring correct typing in the BMInterface* classes (cf. #2628).~~

This pull request was designed to fix the underlying bug with zero-sided option dice without conflicting with the custom recipe pull request (#2389), which disables 0-sided option dice explicitly.

Once both this pull request and #2389 have been accepted, the plan would be to submit another pull request that removes the explicit rule banning 0-sided option dice, and simultaneously removes the special skill flag for my fanatic, which has a (0/4) die.

### DIFF
--- a/src/engine/BMDieOption.php
+++ b/src/engine/BMDieOption.php
@@ -197,6 +197,11 @@ class BMDieOption extends BMDie {
         $this->needsOptionValue = FALSE;
         $this->valueRequested = FALSE;
         $this->max = $optionValue;
+        if (0 === $optionValue) {
+            $this->min = 0;
+        } else {
+            $this->min = 1;
+        }
         $this->scoreValue = $optionValue;
 
         return TRUE;
--- a/test/src/engine/BMDieOptionTest.php
+++ b/test/src/engine/BMDieOptionTest.php
@@ -223,6 +223,25 @@ class BMDieOptionTest extends PHPUnit_Framework_TestCase {
 
     /**
      * @depends testInit
+     * @covers BMDieOption::set_optionValue
+     */
+    public function testSet_optionValue_including_zero() {
+        $this->object->init(array(0,7));
+
+        $this->assertFalse($this->object->set_optionValue(5), 'Invalid option value set should fail.');
+
+        $this->assertTrue($this->object->set_optionValue(7), 'Valid nonzero option value set should pass.');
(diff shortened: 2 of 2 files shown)


cgolubi1:
I think everything you suggest is fine, and i think i should also track down the issue i mentioned in #2628 in response to your comment, and submit a PR for that (which will be in the API layer, so it won't conflict with anything).

blackshadowshade:
Now that the code has been merged to correctly sanitise the type of option values, I've strengthened the type checking of option values and rebased so that this pull request includes the sanitisation.

blackshadowshade:
This has been rebased to be after the incorporation of the CustomBM code, so this should now be ready for consideration.

blackshadowshade:
This has been rebased to be after the switch to testing on a production site.

blackshadowshade:
And rebased to be after the BMInterface* refactors.

cgolubi1:
Hmm, okay, so i'm reminding myself what's happening with this PR.  The big question is: do i need to test that this change *works*, or just that it doesn't break games without zero-sided option dice?  I think the latter, so i think i just throw this at a replay loop, and if it passes, we merge.  Yeah?

blackshadowshade:
Right. We can do some targeted testing once I re-enable my Fanatic in the next pull request.

cgolubi1:
Sounds good.  The replay loop is rolling; i'll keep you posted.

cgolubi1:
5500 replayed games so far.  Once it gets to 10000 (presumably another day or so), i'll declare victory and merge this.

cgolubi1:
20000 replayed games with no failures.  Merging.

PR #2644, opened 2020-11-03
Updated password reset text re unvalidated email address

### LINKED ISSUE #2561: Change password reset info regarding need for verification
We've now had two players (one on a dev site and one on the live site) who have attempted to do password reset with an unverified account. This fails, and I think that the behaviour here is reasonable.

However, we should add some wording to the password reset page to say that this will fail on an unverified account, and if they need help verifying their email address, they should contact us at help@...

blackshadowshade:
Fixes #2561

### DIFF
--- a/src/ui/gui.css
+++ b/src/ui/gui.css
@@ -1131,3 +1131,7 @@ div.forum div.splitRight span.postTag {
 #skills_page table.skills p.interaction {
     font-style: italic;
 }
+
+#forgot_password_action_form {
+    padding-bottom: 3em;
+}
--- a/src/ui/js/ForgotPassword.js
+++ b/src/ui/js/ForgotPassword.js
@@ -58,9 +58,12 @@ ForgotPassword.actionRequestReset = function() {
     'text': 'Request password reset',
   }));
   requestdiv.append($('<p>', {
-    'text': 'If you have forgotten your Button Men password, you can reset ' +
-            'it.  Enter your username, and we will send a password reset ' +
-            'link to the e-mail address we have on file for your account.'
+    'text': 'If your email address has been verified and you have forgotten ' +
+            'your Button Men password, you can reset it.'
+  }));
+  requestdiv.append($('<p>', {
+    'text': 'Enter your username, and we will send a password reset link to ' +
+            'the e-mail address we have on file for your account.'
   }));
   var requestform = $('<form>', {
     'id': 'forgot_password_action_form',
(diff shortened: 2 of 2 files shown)


blackshadowshade:
CircleCI success here: https://app.circleci.com/pipelines/github/blackshadowshade/buttonmen/134/workflows/18e2adfb-246f-434b-b86c-5666a6e9d4b6/jobs/498

cgolubi1:
This will want a dev site, but i ran out of time to stand one up today.  Soon!

cgolubi1:
Finally actually stood up the dev site for this --- check it out at http://fleeting.dev.buttonweavers.com/ui/

dwvanstone:
I want to test this feature, but I want to see the behavior it's trying to fix. Is there a non-production site that I can use that has the bug?

blackshadowshade:
Any of the other dev sites should have the current behaviour.

dwvanstone:
I tried to see the bug in action on ephemeral.dev.buttonweavers.com and its correction on fleeting.dev.buttonweavers.com

I created a user (devious11 on ephemeral, devious12 on fleeting) but did not verify the account. I then selected Forgot Password and typed in my username. 

In both cases I received an email. They look the same. I clicked on the link in the email, attempted to change the password, and got the same error in red:

"User with ID 34 is not an active player"  


I don't see a change in behavior between the two websites.

blackshadowshade:
Take a look at the difference between

http://fleeting.dev.buttonweavers.com/ui/forgot_password.html

and 

http://ephemeral.dev.buttonweavers.com/ui/forgot_password.html

That's all I've implemented: change in the text on that page, not active response text.

dwvanstone:
I'm not a fan of fixing a behavior by adding more explanatory text. Usually if I'm on a website and want to reset my password I expect a box to put in my email and a button to send me a password reset. The rest I mostly ignore.

The use case here (I believe) is that the player thought they verified their account or didn't remember if they did. 

I think it would be better to change the behavior after they click "Request password reset link".

Idea 1: change the error text

Right now it gives an error (in red) if the account doesn't exist.

What if it gives another error if the account exists but hasn't been verified. "NAME has been created but not verified. Contact help@buttonweavers.com for help."


Idea 2: Go to a new page

After clicking the reset link, the website goes to a new page that says  "The account NAME has not been verified. Click here to send a new validation email."


Idea 3: Send a different email

After clicking the reset link, the same text appears but the email asks them to validate their email and after that they can change their password.

cgolubi1:
I like dwvanstone's first idea: have the password reset request fail with an error message explaining that since the account isn't verified, nothing good is going to happen.  That doesn't seem like it should be too hard --- it puts the logic in the backend where it belongs, and that way if for some reason we changed the behavior later, we wouldn't have stale UI text floating around.

blackshadowshade:
Alright, I'll have to dig through the code to work out how we know whether a player is verified.

cgolubi1:
I'm pretty sure it's just `status == 'ACTIVE'` in the player DB.

cgolubi1:
Any update on this?  I still have a dev site kicking around, and i don't think it has a lot of valuable state, so if this isn't on your front burner, i'll take down the dev site and we can stand up a new one once there's a new version.

blackshadowshade:
Yeah, ditch this for the moment. This needs to be entirely reworked from scratch.

PR #2645, opened 2020-11-03
Made opponent name persistent when changing first player

### LINKED ISSUE #2572: Create Game: opponent's name is erased when changing first player
Create a new game.

Select an opponent. Then hit Change First Player. The opponent is set back to "Anybody."

This kept happening to me as I was setting up games for a tournament - I wouldn't notice and the games were being placed in Open Games.

blackshadowshade:
Fixes #2572.

### DIFF
--- a/src/ui/js/Newgame.js
+++ b/src/ui/js/Newgame.js
@@ -208,6 +208,13 @@ Newgame.createPlayer1Toggle = function() {
   });
 
   player1Toggle.click(function() {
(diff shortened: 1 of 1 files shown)


cgolubi1:
Check this out at http://blip.dev.buttonweavers.com/ui/

dwvanstone:
This change looks good. It fixes the bug I reported. Thanks!!

cgolubi1:
Thanks for reviewing.  Merging now.

PR #2646, opened 2020-11-03
Visually indicate that die is irrelevant for initiative

### LINKED ISSUE #835: During take initiative, indicate which dice are not relevant
I've often thought that it shouldn't be hard to show the player which dice are not part of the initiative process, to avoid confusion when using focus and chance dice.

Mostly, for me, I puzzle over situations where I or my opponent have Trip dice, but perhaps there are other dice which are ignored in the initiative process that I'm forgetting.

Anyway, I'd think it should be fairly easy to dim such dice (and include a note in the rollover to explain?) to clarify things a bit.

blackshadowshade:
Fixes #835.

On my system, the updated display looks like this:

irrelevant_for_initiative

Let me know what you think about how this looks.

### DIFF
--- a/src/engine/BMDie.php
+++ b/src/engine/BMDie.php
@@ -353,7 +353,6 @@ class BMDie extends BMCanHaveSkill {
         return $val;
     }
 
-//
     /**
      * Returns ten times the "real" scoring value
      *
@@ -391,8 +390,6 @@ class BMDie extends BMCanHaveSkill {
         }
     }
 
-    //
-
     /**
      * Return die's initiative value.
      * Negative means it doesn't count for initiative.
@@ -405,6 +402,17 @@ class BMDie extends BMCanHaveSkill {
         $this->run_hooks(__FUNCTION__, array('initiativeValue' => &$val));
 
         return $val;
+    }
+
+    /**
+     * Add flag to signal whether the die is considered by initiative
+     */
+    public function add_initiative_flag() {
+        $initiativeValue = $this->initiative_value();
+
+        if ($initiativeValue && ($initiativeValue < 0)) {
+            $this->add_flag('IrrelevantForInitiative');
+        }
     }
 
     /**
--- a/src/engine/BMFlagIrrelevantForInitiative.php
+++ b/src/engine/BMFlagIrrelevantForInitiative.php
@@ -1,0 +1,13 @@
+<?php
+/**
+ * BMFlagIrrelevantForInitiative: Used to signal that a die is irrelevant for initiative
+ *
+ * @author: james
+ */
+
+/**
+ * This class is a flag that signals that a die is irrelevant for initiative
+ */
+class BMFlagIrrelevantForInitiative extends BMFlag {
+
+}
--- a/src/engine/BMGame.php
+++ b/src/engine/BMGame.php
@@ -4075,6 +4075,10 @@ class BMGame {
             }
 
             foreach ($player->activeDieArray as $dieIdx => $die) {
+                if (BMGameState::REACT_TO_INITIATIVE == $this->gameState) {
+                    $die->add_initiative_flag();
+                }
+
                 if (!empty($die->flagList)) {
                     foreach (array_keys($die->flagList) as $flag) {
                         // actively lie about auxiliary choices to avoid leaking info
--- a/src/ui/gui.css
+++ b/src/ui/gui.css
@@ -375,6 +375,10 @@ select.center {
     background-color: #ffa500;
 }
 
+.recipe_irrelevant_for_initiative {
+    color: #cccccc;
+}
+
 .die_overlay {
     position:relative;
     font-size: 21px;
--- a/src/ui/js/Game.js
+++ b/src/ui/js/Game.js
@@ -2432,16 +2432,10 @@ Game.dieRecipeTable = function(table_action, active) {
             Game.dieValueSelectTd('init_react_' + i, initopts,
               Api.game.player.activeDieArray[i].value, defaultval));
         } else {
-          dieLRow.append($('<td>', {
-            'text': Game.activeDieFieldString(
-              i, 'value', Api.game.player.activeDieArray),
-          }));
+          dieLRow.append(Game.valueCell(i, true));
         }
         dieRRow.append(opponentEnt);
-        dieRRow.append($('<td>', {
-          'text': Game.activeDieFieldString(
-            i, 'value', Api.game.opponent.activeDieArray),
-        }));
+        dieRRow.append(Game.valueCell(i, false));
       } else if (table_action == 'adjust_fire_dice') {
         dieLRow.append(playerEnt);
         var fireopts = [];
@@ -2520,6 +2514,30 @@ Game.dieRecipeTable = function(table_action, active) {
   }
 
   return dietable;
+};
+
+Game.valueCell = function(dieIdx, isPlayer) {
+  var activeDieArray;
+  if (isPlayer) {
+    activeDieArray = Api.game.player.activeDieArray;
+  } else {
+    activeDieArray = Api.game.opponent.activeDieArray;
+  }
+
+  var cellText = Game.activeDieFieldString(dieIdx, 'value', activeDieArray);
+  var valueCell = $('<td>', {
+    'text': cellText,
+  });
+
+  if ('' !== cellText) {
+    var die = activeDieArray[dieIdx];
+    if (('properties' in die) &&
+        (die.properties.indexOf('IrrelevantForInitiative') >= 0)) {
+      valueCell.addClass('recipe_irrelevant_for_initiative');
+    }
+  }
+
+  return valueCell;
 };
 
 Game.playerWLTText = function(player) {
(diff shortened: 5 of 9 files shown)


TheOrgg:
This'll be nice.

cgolubi1:
This is going to want dev and replay testing; dev so people can report what they think of it, replay because it adds a flag.

cgolubi1:
Dev site is at http://momentary.dev.buttonweavers.com/ui/ --- check it out

TheOrgg:
Created a few games on the test server, but I'm not sure what bugs to look out for.

blackshadowshade:
I guess you want to check for matchups between buttons with varied skills, some which affect initiative and some which don't. However, don't let me guide you on testing too much, part of the skill in testing is finding things that the developers never thought of.

jl8e:
A thing to probably check is how it handles focus/chance dice that don't count  (I know there's at least one fanatic that does this)

blackshadowshade:
Also, as you test, it would be useful for us (and this is generally true of non-trivial pull requests) if you list in some detail what you have tested.

TheOrgg:
Had a little time, set up a few matches against a friend's account he's letting me use.

Game 194 is frozen for some reason; every time either accountt ries to go to it, the screen is blank and doesn't load.

So far the only thing I've noticed is during the swing die selection, the non-initiative dice are not indicated.  During Chance or Focus, they are.  Is that the correct behavior?

irilyth:
BTW, on the staging site and dev sites, it's totally fine to create multiple accounts for testing.

blackshadowshade:
I've only indicated non-initiative dice in the phase where you're trying to gain initiative.

Did people want me to indicate them also during swing/option setting?

TheOrgg:
I could see it going either way.  I'd say post a thread asking about it on the Forum.  I know most people use the formula to determine their swing die, but I usually just eyeball it.

blackshadowshade:
@nkmcalli, @Caligari, your thoughts?

cgolubi1:
Trying replay testing again with this updater, which expects to see the new property during REACT_TO_INITIATIVE only and on dice with rage, stinger, or trip only:

```python
#!/usr/bin/python
##### update_replay_games
# be made in random_ai.py.

import sys
sys.path.append('lib')

import random_ai

SKIP_INITIATIVE_SKILLS = set(['Rage', 'Stinger', 'Trip'])
def add_initiative_flag(diearray):
  retval = []
  for die in diearray:
    if SKIP_INITIATIVE_SKILLS.intersection(die['skills']):
      die['properties'].append('IrrelevantForInitiative')
    retval.append(die)
  return retval

# Custom classes
  #
    if data['gameState'] == 'REACT_TO_INITIATIVE':
      for i in range(len(data['playerDataArray'])):
        data['playerDataArray'][i]['activeDieArray'] = add_initiative_flag(data['playerDataArray'][i]['activeDieArray'])
    return data

# Initial setup
if len(sys.argv) < 2:

DIRNAME = sys.argv[1]

  d.translate_game(gamepath)
```

TheOrgg:
What about slow?  (Also Evil (E) and Sleep (Z) also don't count, though they're not yet active.  Noting it here for reference)

blackshadowshade:
The current code does not gray out the dice for a slow button (like Giant).

cgolubi1:
Yeah, but Slow *dice* should count, and i'll make sure they do in testing.  Incidentally, i didn't notice slow initially because (see http://www.buttonweavers.com/ui/skills.html ) the slow skill uses the wording "These dice are not counted for the purposes of initiative." whereas all other initiative-excluded skills seem to use "...do not count for determining who goes first."

blackshadowshade:
This code should work perfectly fine for slow dice.

TheOrgg:
There's at least two sets that have Slow dice other than Giant (and he has a button special to make sure he NEVER goes first, not slow dice).

blackshadowshade:
@TheOrgg, should be easy enough to test these out on the dev site!

TheOrgg:
I will do so if they've been added.

Also, did you find what is hanging up Game 194 on the test server?

blackshadowshade:
Regarding Game 194, here's the information I have:

The two players are theorgg and SextusUrungus.

The two buttons are

    Stumbling Clowns: (8):5 t(8):6 (10):7 t(10):2 (X=4):1

and

    IIconfused: fn(X=6):5 fn(X=6):4 gn(X=6):1 gn(X=6):2 `H(Y,Y=3):6 `d(Y=3)!:3 `mp(Y=3):3 `^(1,Y=3):4

The game is currently in BMGameState:REACT_TO_INITIATIVE, waiting for SextusUrungus to decide whether or not to react with focus.

blackshadowshade:
Right, I think I've dealt with that bug. @cgolubi1, if you wouldn't mind updating the code base for us, please.

@TheOrgg, once the code has been updated, that game should work perfectly well.

TheOrgg:
What exactly happened?

blackshadowshade:
Basically, Javascript didn't want to evaluate die.properties... when 'properties' was not a property of 'die'. As usual, it simply stopped rendering the page beyond that line.

TheOrgg:
The dev site is still getting hung up on that  Game 194.

blackshadowshade:
It hasn't been updated yet.

cgolubi1:
I tried to update this dev site, and the instance wound up in (possibly was already in) a weird state --- i'll look into it more later.  (Nobody panic; no dev action is needed.  Just wanted to let people know this dev site appears to be down right now.)

cgolubi1:
I think i've updated the code and the new Game.js is in place and http://momentary.dev.buttonweavers.com/ui/game.html?game=194 still isn't loading.  Do other folks see the same thing?  If not, then i just have bad JS cached.

cgolubi1:
For what it's worth, though, i haven't done JS debugging in forever, but firefox dev tools is telling me jquery is throwing:

```
Uncaught TypeError: right-hand side of 'in' should be an object, got undefined
```

So i wonder if `in` doesn't work if `die` isn't defined in the first place, and if that's the situation you're in.

TheOrgg:
Nothing's coming up at all for me.

blackshadowshade:
> For what it's worth, though, i haven't done JS debugging in forever, but firefox dev tools is telling me jquery is throwing:
> 
> ```
> Uncaught TypeError: right-hand side of 'in' should be an object, got undefined
> ```
> 
> So i wonder if `in` doesn't work if `die` isn't defined in the first place, and if that's the situation you're in.

I'm seeing the same error, but I can't work out why 'die' is returning undefined.

blackshadowshade:
Interesting. By hacking the dev site (sorry Chaos if this complicates things!), apparently the issue is that the die index is going up to 5, while there are only four dice in the activeDieArray (thus the index should only go up to 4).

I'll take a look at why that's happening.

blackshadowshade:
Oh! Obvious!

The for loop in Game.js at line 2402 is running from i = 0 to maxDice, where maxDice is the maximum number of dice for both the player and opponent.

Argh.

Easy fix coming up.

blackshadowshade:
Okay, bug should now be fixed better. Try that.

TheOrgg:
Game 194 is still held up and not loading.

blackshadowshade:
Not surprising, the bug fix needs to be deployed to the test site before you'll see any change in behaviour.

cgolubi1:
Dev site is updated.

cgolubi1:
The game is loading for me now; check it out, other testers.

blackshadowshade:
Yep, loading for me now.

TheOrgg:
I played out that dozen or so matches out, and everything seemed just fine.  

Was the 194 bug related to this enhancement, or was that a separate issue that hadn't been discovered before?

blackshadowshade:
The bug you found was related to the changes I made, so specific to this enhancement.

cgolubi1:
Where are we?
* I never reported back on the success of my replay-testing, so i should check on that.
* TheOrgg says this works now UI-wise
* Anyone else have anything to say or test before we merge this, assuming replay testing looks fine?

dwvanstone:
I've tried a bunch of games on this -- it looks really good. Wranklepig adds its own wrinkle - you can reroll a chance stinger die at the beginning and the die is grayed out but it still works. I think it's nice.

I don't know what would happen with a focus-stinger (or focus-trip or focus-rage) but I couldn't find a button to test it on.

cgolubi1:
Replay testing tells me that Warrior dice are also going to show up as greyed out (try a game with IIconfused).  dwvanstone, do you approve of how that looks too?  If so, i'll call this good and merge it.

dwvanstone:
> Replay testing tells me that Warrior dice are also going to show up as greyed out (try a game with IIconfused). dwvanstone, do you approve of how that looks too? If so, i'll call this good and merge it.

Yes. I like that the warrior dice are greyed out. You can also see other players' games greyed out even if you're not part of the game. Go as a user who is not "dvstest2" nor "dvstest3":

http://momentary.dev.buttonweavers.com/ui/game.html?game=221

I think that's okay too.

cgolubi1:
Great!  Thanks for all the iteration on this, everyone.  Shadowshade, why don't you rebase this down to one commit, and i'll merge it?

blackshadowshade:
Rebase done, CircleCI is happy.  :)

PR #2647, opened 2020-11-06
Updated image of Mississippi

### LINKED ISSUE #2574: Update button art for Mississippi
The US state of Mississippi is changing its state flag; a new design hasn't been decided yet, but when one is, we should update the button art as well.

blackshadowshade:
Fixes #2574.

This pull request is still open for comments regarding the new button image.


blackshadowshade:
When viewing the two buttons next to each other (i.e., the Rich Diff button on the Files changed page), it's quite obvious that the new image is quite blurry. I'm just wondering why that is.

blackshadowshade:
Pinging @Hurtay for your insight into the blurriness.

TheOrgg:
I edited the original Mississippi button to keep the text and dice frames the same.  The blurryness of the new one is because I had to shrink it down so much.

Do we have the larger size images for the States set?

blackshadowshade:
@Hurtay (Kaori) has all of those. You should be able to ask her for those.

cgolubi1:
Thanks very much for getting the new art!

I think we should accept the blurry art if need be to get it into the next release.  I don't know when the next release is, but i'll get an answer to that question in the next day.  And obviously (IMO) nothing very bad happens if we merge blurry art now and improve it later.

TheOrgg:
It just needs to be done by the time Chaotic Dice are done.

And the text at the bottom already has a court case against it; it was designed to be fought against, and likely will change in the next five or six years.

But for now, this one is official.

@Hurtay , any larger button artwork you have, I would very much appreciate it for the wiki (and times like this where I edit things a bit).

cgolubi1:
I'm going to go ahead and merge this, and thank you!  (I know the button isn't the prod site, but it shows up on dev sites.)  We'll keep iterating as we go.

PR #2650, opened 2020-11-14
Added try/catch blocks to BMInterface functions

### LINKED ISSUE #1294: add try/except blocks to public BMInterface functions which don't have them
The BMInterface public function `get_profile_info()` executes a database query without a try/except block.  Add one.

blackshadowshade:
Fixes #1294.

I've been through all the functions in BMInterface*.php and added try/catch blocks to those functions that are not already covered by a try/catch block.


cgolubi1:
I'm just going to accept this on the strength of the normal responder tests --- they should hit the happy path of all the interface functions, and the implication is that we already weren't hitting the unhappy path.

PR #2654, opened 2020-11-20
Changed commas to CHAR(59) for flavour text

### LINKED ISSUE #2653: Problem with semicolons in MySQL UPDATE
When I addressed #1699, I decided to use HTML encoding for special characters in flavour text. This appears to work fine for our deployments.

However, I'm having trouble doing automatic database rebuilds on my local MAMP stack because of the semicolons in the strings. It appears that I'm not the only one who has this issue: https://stackoverflow.com/questions/31816399/cant-escape-semicolons-in-a-mysql-update-statement

I need to change the deployment file data.button.sql so that it works both for my MAMP deployment as well as the live and dev sites.

blackshadowshade:
Fixes #2653.

I've not included any update file here, since I shouldn't actually be changing the database with these changes.

This change fixes my ability to deploy automatically on my MAMP stack.


cgolubi1:
We need an update file --- otherwise, new/testing sites will have a different database than prod and staging.  We don't want that kind of slippage.  If these changes aren't compatible with Linux sites, then they'll break other testing and we can't use them.  If the changes are harmless on Linux sites, then we should go ahead and apply them to prod and staging.  So go ahead and add an update file, and we can check it out and make sure it's harmless.

blackshadowshade:
Update file added and rebased.

blackshadowshade:
Just for clarity, this update file should be a no-op, so a pre-update and post-update database should be essentially identical.

cgolubi1:
Okay.  Quick note that you said "commas", but this is actually impacting semicolons, and in particular the semicolons which are at the end of HTML characters.  So i definitely want to look at this in my stack and make sure the impacted flavor texts look correct.

cgolubi1:
You're right --- this is a noop.  Thanks for including the updates file anyway; if we don't have to guess, we don't have to guess wrong.

PR #2655, opened 2020-11-21
Made selected button persistent on filter change

### LINKED ISSUE #1010: Create Game shouldn't lose your selection if you change limits
If you select a button from the Create Game page, then change the limits (e.g., setting "Tournament legal" from "ANY" to "yes"), then as long as the button you selected still matches the filters, it should still remain selected. At the moment, when you make the change, it reverts to "Choose your button" regardless.

blackshadowshade:
Fixes #1010.

I've tested this with Limax: (W) om(34) (T). The selection remains there if I filter according to die skill = ornery or morphing or ANY, but not if I select some other skill.

### DIFF
--- a/src/ui/js/Newgame.js
+++ b/src/ui/js/Newgame.js
@@ -608,8 +608,9 @@ Newgame.getSelectOptionList = function(
       'text': blankOption,
     }));
   } else {
-    // If there's no default, put an invalid default value first
-    if (selectedval === null) {
+    // If there's no default or the selected value doesn't exist
+    // in the dropdown, put an invalid default value first
+    if ((selectedval === null) || !(selectedval in valuedict)) {
       optionlist.push($('<option>', {
         'value': '',
         'class': 'yellowed',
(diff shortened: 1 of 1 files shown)


cgolubi1:
This presumably wants a dev site.

cgolubi1:
Check this out on ephemeral.dev.buttonweavers.com

dwvanstone:
It's not quite acting as I'd expect.

When I'm choosing my own button, it starts with "Choose your button"

* Selecting "Auxiliary" keeps it on "Choose your button". This seems right.
* Let's say I pick Mordred. That seems right so far.
* If I now select "Shadow" it keeps Mordred. That seems right.
* If I now select "Rush" it changes to Amelia. I'd expect it to go back to "Choose your button".

Similarly for the opponent's button, I'd rather it go back to "Choose opponent's button" if the selection doesn't match any criteria rather than go to the first button in the skill list.

I worry that someone may not notice the button name changing and may end up with a button they did not choose.

cgolubi1:
I can reproduce the behavior dwvanstone describes, and agree that it's not right.

blackshadowshade:
Okay, I've added some extra logic to handle the case when your selected button doesn't exist in the dropdown. This site can be updated.

cgolubi1:
I've updated http://ephemeral.dev.buttonweavers.com/ui/ --- dwvanstone and others, check it out.

dwvanstone:
It looks fantastic. Thanks!

cgolubi1:
Thanks for testing!  Shadowshade, want to rebase this down?

blackshadowshade:
Rebase done.

PR #2658, opened 2020-11-22
Made BMAttackShadow->validate_attack() function declaration consistent

### LINKED ISSUE #2657: BMAttackShadow->validate_attack() has inconsistent function declaration
In #2253, I noted that the function declaration for BMAttackShadow->validate_attack() was not consistent with the function declaration in its parent class, BMAttackPower, which causes a fatal error in PHP 7.2 and up.

This behaviour has been noted elsewhere, e.g. https://www.drupal.org/project/drupal/issues/2927173. In the resulting bug post (https://bugs.php.net/bug.php?id=75590), one responder states:

    The fact that PHP < 7.2 accepted this was an unintentional side-effect.  The code you present doesn't follow valid inheritance contract.

blackshadowshade:
Fixes #2657

### DIFF
--- a/src/engine/BMAttackShadow.php
+++ b/src/engine/BMAttackShadow.php
@@ -22,9 +22,10 @@ class BMAttackShadow extends BMAttackPower {
      * @param BMGame $game
      * @param array $attackers
      * @param array $defenders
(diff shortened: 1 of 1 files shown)


cgolubi1:
I'm assuming this is fine because it passes tests.  (If it subsequently turns out not to work, we'll need some more tests.)

PR #2674, opened 2020-12-23
Add simple BMDB layer which casts selected values to consistent types

cgolubi1:
Don't merge this yet, because it'll need a rebase.

This is intended to be a version of #2628 which can be merged without interfering with tournaments, and will give me a framework to build on driving down (int) casts across the codebase.

First commit, just implement BMDB:

* BMDB implements a couple of types of select
* No callers invoke BMDB yet --- that will happen in future PRs
* Functions are fully tested using dummy DB connector objects

Second commit, use BMDB to model all database accesses in BMInterfaceHistory.php

* All `(int)` casts are removed, because the database row selects now take care of them
* All direct accesses to `self::$conn` are now replaced with `self::$db` calls (note that this was possible to do with the existing infrastructure because BMInterfaceHistory.php doesn't do any DB updates --- i may have to add a new BMDB function when i go to migrate a class which does DB updates, but one thing at a time)
* I had to add a third int type --- at this point, i think this handles all the ways NULL might be treated when we look for an int in the database:
    * `int`: raise an exception if NULL is received --- this is what we *want* the default to be, because if the caller hasn't thought about the possibility that their select might get a NULL, that probably means there's a bug if their select *does* get a NULL, and we want to draw attention to it (by failing), not paper over it
    * `int_or_null`: return an int type if the DB entry is set; if it's unset, return NULL.  In this case, the calling function has to be aware that it might wind up with an int or a NULL, and has to be able to do the right thing.  We have a lot of code that works that way, but we explicitly flag it because the caller has to know that they might have a NULL
    * `int_null_becomes_zero`: it's not clear to me that this behavior is desirable, but it's the way some of the BMInterfaceHistory.php code works, and it's not very harmful in that case (the situation is that the SQL is doing some sums to get some of the fields that are returned, and if there's nothing to sum, apparently `SUM()` returns NULL, but we want zero), so let's support it with a funny name that makes it clear it's funny behavior

### DIFF
--- a/src/engine/BMDB.php
+++ b/src/engine/BMDB.php
@@ -1,0 +1,121 @@
+<?php
+/**
+ * BMDB: database interface to standardize types
+ *
+ * @author: Chaos
+ */
+
+/**
+ * This class contains all the logic to do with selecting from and updating
+ * the buttonmen database tables.
+ *
+ * It should be instantiated with a PDO database connection created by the caller
+ * (which will be a TestDummyPDOConn in the case that the caller is unit testing).
+ */
+class BMDB {
+    // properties
+
+    /**
+     * Connection to database
+     *
+     * @var PDO
+     */
+    protected static $conn = NULL;
+
+    /**
+     * Constructor
+     *
+     * @param PDO $conn
+     */
+    public function __construct($conn) {
+        self::$conn = $conn;
+    }
+
+    /**
+     * Execute a query which fetches a single value from the DB
+     *
+     * @param $conn
+     * @param string $query
+     * @param array $parameters
+     * @param array $returnType
+     * @return array
+     */
+    public function select_single_value($query, $parameters, $returnType) {
+        $statement = self::$conn->prepare($query);
+        $statement->execute($parameters);
+        $result = $statement->fetch(PDO::FETCH_NUM);
+        if (!$result) {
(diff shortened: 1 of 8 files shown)


cgolubi1:
Shadowshade, since i don't think you looked at this yet, i went ahead and added a second commit that converts BMInterfaceHistory.php to use BMDB.  AFAIK these are both ready to go, but obviously if we like it, it'll have to be rebased down.  LMK what you think and what you want for testing once you have a chance to take a look.

blackshadowshade:
I'm okay with the changes here, subject to some testing on a dev site of the history function.

For consistency with the (re-)naming of exceptions in the custom recipe pull request, I'd go with "BMExceptionDatabase".

cgolubi1:
Renamed the exception class.  If it passes tests, i'll stand up a dev site.

cgolubi1:
Shadowshade: check it out at http://stopgap.dev.buttonweavers.com

blackshadowshade:
The dev site at stopgap looks fine, I've tested the history page with restrictions on game ID, start date, and last played date, no errors popping up as far as I can see.

This should be fine to rebase and merge.

cgolubi1:
Thanks!  Rebased.  If it passes testing, you can merge it, or i can.

PR #2680, opened 2020-12-29
add some New Game functionality to the python api-client

danlangford:
i needed this functionality and thought i could give it back to the community


cgolubi1:
This looks good to me in by-hand testing; we should fix the backend/API-layer uncleanness i found while testing so that clients can't get stacktraces, but that's no reason not to have a client.

Issue #2683, opened 2020-12-29
Get development environment working on M1 Macs


jl8e:
VirtualBox doesn't work on the new Macs, and that's not likely to change soon, if ever.

Docker might be an alternate, but it's not supported yet, either. 

There's a free license for VMWare Fusion, which can be used with Vagrant, supposedly as a drop-in replacement for VirtualBox, so it might yet be as easy as getting an ARM Linux image of Linux instead of the x86 one, but it's not going to be available until next year sometime. (And getting that answer was way harder than it should've been.)

So, I'm going to have to operate with an unvirtualized environment. I can do that. 

So, what am I going to need? I have PHP 7.3.24 already, as well as Python 2 and 3. I'm going to need:
MySQL
Apache
various PHP modules

Anything else? Do I need Puppet? (I think not; I'm not rebuilding the environment from scratch all the time, so I'll just need a script to reset the db.)

blackshadowshade:
I'm using MAMP, and that seems to work fine.

blackshadowshade:
Regarding a script to reset the DB, there's already one in the deploy/amp folder.

blackshadowshade:
If you're not going down the MAMP route, you may still want to peruse the WAMP primer that Fog wrote, which may give you some hints about what needs to be configured:

https://github.com/buttonmen-dev/buttonmen/wiki/Developer-guide%3A-setting-up-a-WAMP-stack

jl8e:
ok, this is going to be easier than I feared. :)

Is that the bundle from mamp.info?

Not sure which route I'm going to take; I usually end up wanting MySQL for other purposes anyway, but an easy-to-install bundle has a lot of appeal.

blackshadowshade:
Yep, mamp.info would work, as would the sourceforge site.

If you run into trouble, feel free to pop onto our IRC channel (\#buttonmen on irc.freenode.net).

blackshadowshade:
I've kept the primer pretty much up to date, since I needed to use it to set up my work computer about a year ago.

jl8e:
Thanks

(Adds "IRC client" to list of things to install on the new box.)

blackshadowshade:
I use Adium because it quacks.  :)

blackshadowshade:
I've just taken a look at some of the functionality using the MAMP, and there's the problem that the requisite MySQL version is 5.7, which enforces strict mode. This doesn't allow times to default to 0.

[...]

PR #2684, opened 2020-12-30
Replaced one more semicolon in the SQL for button flavour text

blackshadowshade:
Addresses a semicolon I forgot to address for #2653, as discovered in #2683.

Again, this database update should be a no-op.


cgolubi1:
Seems fine, testing corroborates that it's a noop.

PR #2688, opened 2021-01-05
Migrate BMInterfacePlayer and BMInterfaceGameChat to use BMDB

cgolubi1:
* Partially addresses #712
* This change is effectively a noop --- it does select `player_fanatic_id` as NULL rather than 0 (which required a unit test change), but that field is not used by anything and is never set (it is NULL for 100% of players on the prod and staging sites) and IMO it's fine for us to select it as NULL if it's NULL
* It also fixes the setChatVisibility handling in ApiResponder to expect a bool rather than a string.  I'm surprised the string handling was even working, given that the ApiSpec configuration definitely means it's getting a bool.  PHP is very weird.  Anyway, by my testing it works just fine with a bool.
* I added update functionality to BMDB for this change --- nothing in BMInterfacePlayer.php required that functionality to do anything special, so i just made it minimal.  I can expand it later if needed for other classes.
* I tested this a bunch in virtualbox, loading and modifying preferences, and i believe it's just fine --- i know i removed some code that claimed we needed to explicitly cast booleans to 1-character ints so the right thing would happen in the DB, but that doesn't seem to be true.  It definitely looks to me like if we update a mysql int field by handing it a bool value, the right thing will happen.  I suspect that's vestigial code from some past time when the values might have internally been a string representation of an int, which indeed might not have done the right thing.  But that's the world we're moving away from --- with this codebase we can be sure that whatever got past the API is a bool, so we shouldn't need to cast it.

### DIFF
--- a/deploy/circleci/audit_php_files.php
+++ b/deploy/circleci/audit_php_files.php
@@ -31,6 +31,7 @@
   // In these directories, we limit the use of explicit (int) casts
   $php_limit_int_cast_dirs = array(
     "src/api",
+    "src/engine",
   );
   $php_limit_int_cast_exceptions = array(
     // exceptions for $_SESSION variables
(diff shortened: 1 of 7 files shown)


cgolubi1:
Shadowshade: let me know if you want a dev site for live testing, or if you want to test it on MAMP, or if you're happy to merge it with unit tests, or whatever your preference.

blackshadowshade:
I've tested these changes on my MAMP stack. Preference changing works as far as I can see, and the corresponding database state looks good. Game chat works, and the corresponding database state looks good.

Merging this now.

PR #2690, opened 2021-01-08
Migrate BMInterfaceForum to BMDB

cgolubi1:
* Partially addresses #712

### DIFF
--- a/deploy/circleci/audit_php_files.php
+++ b/deploy/circleci/audit_php_files.php
@@ -48,7 +48,6 @@
     "src/engine/BMDieSwing.php" => 1000,
     "src/engine/BMGame.php" => 1000,
     "src/engine/BMInterface.php" => 1000,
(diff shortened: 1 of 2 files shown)


cgolubi1:
This one is a bunch of lines because BMInterfaceForum does a lot of DB things, but it's pretty by-the-book.  I tested what i think are all the forum actions manually.

I didn't test deleted forum posts because i don't think there's any API way to get one (and there are no deleted posts on the prod site), so i'm not going to worry about it.

cgolubi1:
Rerunning circleci to see if it passes this time, because i don't think the failure is me.  FYI, this isn't the only recent build of mine i've seen fail in qunit with this `UserPrefs not found` type thing (the previous one was my master branch after merging from upstream).  I think we've introduced a new race condition into the QUnit tests somehow.

blackshadowshade:
Looks okay to me on my MAMP stack. Accepting this one.

PR #2696, opened 2021-01-18
Add --follow-symlinks flag to sed commands for staging site (and deprecated dev site)

cgolubi1:
This is the change that should mean i won't have to manually update the site type in Config.js for staging on the next BM release.

### DIFF
--- a/deploy/vagrant/modules/buttonmen/manifests/init.pp
+++ b/deploy/vagrant/modules/buttonmen/manifests/init.pp
@@ -98,7 +98,7 @@ class buttonmen::server {
       exec {
         "buttonmen_update_config_sitetype":
           command =>
-            "/bin/sed -i -e '/^Config.siteType =/s/production/development/' /var/www/ui/js/Config.js",
+            "/bin/sed --follow-symlinks -i -e '/^Config.siteType =/s/production/development/' /var/www/ui/js/Config.js",
           require => Exec["buttonmen_src_rsync"];
       }
     }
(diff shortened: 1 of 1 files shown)


cgolubi1:
I didn't test this, but i got the syntax from my setup_dev_site script, which contains the line:

```
ssh ${sshargs} $ipaddr "sudo /bin/sed --follow-symlinks -i -e '/^Config.siteType =/s/production/development/' /var/www/ui/js/Config.js"
```

blackshadowshade:
The easiest way to test this is when we next deploying a test site, check that the Config.js is updated.

Do you want this merged before deployment, or shall we wait until afterwards? I don't think there's any hurry?

cgolubi1:
> The easiest way to test this is when we next deploy a test site, check that the Config.js is updated.
> 
> Do you want this merged before deployment, or shall we wait until afterwards? I don't think there's any hurry?

No, that won't help.  These two lines are for:
1. The deprecated site named `dev.buttonweavers.com`, which is no longer a site we run
2. The site named `staging.buttonweavers.com`

There isn't puppet code that manages these flags for individual dev sites --- i do that with a wrapper script which contains the line i quoted.  So the only way to test this code is to stand up a new site named `dev.buttonweavers.com`, or test it on staging itself (by merging it without testing it).  I think we should do the second one.  The costs of failure are not high.

blackshadowshade:
Ah, okay. I was under the impression that the settings for dev.buttonweavers.com would carry across to dev sites, but that was obviously incorrect.

Let's merge this then. We'll see any potential problems the next time staging is deployed.

PR #2701, opened 2021-01-20
Added interaction text for Warrior Turbo and Warrior Konstant

### LINKED ISSUE #2677: add interaction text for Warrior + Konstant
As implemented, a Warrior Konstant die can't add a negative value before it's brought into play (see screenshot below).  This behavior is correct, since it parallels the behavior of Warrior Stinger.  But unlike Warrior Stinger, it's not called out in the skill interaction text.

We should add skill interaction text clarifying that a Warrior Konstant die can only add a positive value to a skill attack.

warrior_konstant

### LINKED ISSUE #2679: Document interaction between warrior and turbo
Per http://www.buttonweavers.com/ui/forum.html#!threadId=914 it's not blindingly obvious that if you have a warrior turbo die, it's correct that you should get to change its size.  It *is* correct (because the reroll happens after the die is in play, after which normal die skills take effect), but it should be documented as a skill interaction.

blackshadowshade:
Fixes #2679. Fixes #2677.


cgolubi1:
This seems fine to me.  Thanks.

PR #2666, opened 2020-12-07
Added Cybersuit Corp set (was Corpo World)

blackshadowshade:
I've updated this pull request to be mergeable once we are happy with it. Fixes #2664.

~This is the initial testing pull request addressing #2664. @cgolubi1, please deploy a test site for this when you can, so that testing can begin.~

~This is NOT TO BE MERGED, and I'm deliberately allowing the unit tests to fail until testing is completed.~

Issue #2346, opened 2018-04-16
Add new Fanatics buttons


blackshadowshade:
We've had some interest in having new Fanatics buttons for players on the new site, and I think it would be a great idea to make this work.

We need to discuss what guidelines are reasonable when we are vetting potential recipes.

Naturally, any button that triggers a bug will be deactivated (but potentially playable on a testing site) until the bug is fixed.

irilyth:
This might have some overlap with #344 and #231.

AdmiralJota:
A few thoughts, I guess:

* I agree that it would be good to implement #344 before we roll into Fanatics full time. I'll post over on that one too, so we can start that discussion up again.

* Do we want to have specific recipe guidelines guidelines on what is or isn't allowed, or do we want to base it off of the results of testing the buttons, or do we just want to go with the "I know it when I see it" plan for rejecting extremely unbalanced buttons, or do we just want to allow everything, or what? 

    Out of those options, I think I'd lean towards number two: base it off of the testing. Here are some possible guidelines we might consider:

    * You need to pit your recipe against at least N (20?) other buttons that are deemed fair (TL? Win ratio between 40%-60%?). 
    * At least N/2 of those games have to have your *opponent* playing as your new recipe. 
    * Your new recipe's win rate in those games has to be between X%-Y% (35%-65%?).
    * If we think you're intentionally trying to game the results we'll auto-reject it.

* Also, given that we don't have player mail yet, what process do we want for people actually submitting their new Fanatic recipes that they're ready to have approved? Post to a special forum thread? It seems like they shouldn't be obligated to sign up for a GitHub account just to request their Fanatic.

blackshadowshade:
Regarding win rate, is there a case for retiring old Fanatics at the same time as enabling new Fanatics?

Also, people should certainly be able to submit their new Fanatics recipes on a special forum thread. However, the specific forum thread might end up being temporary, since we might set this up on a dev site devoted to testing Fanatics.

irilyth:
Why would we want to retire existing Fanatics?

blackshadowshade:
After thinking a little more about this, once we start flagging 'fair' and 'unfair' buttons (#1211), it isn't actually necessary to retire broken buttons.

irilyth:
I don't think it's particularly necessary to retire "broken" buttons at all, but if we want to, I don't think we particularly need to create new buttons to replace them in any case. :^)

dwvanstone:
[...]

Issue #2692, opened 2021-01-12
Hit tables, and their problems


jl8e:
Pulling this out of #2389

>@jl8e, I have no problems with hit tables, other than they exhaustively calculate all possible attacks. All we need to do is cap the size of the hit table to not look past the max value of the target dice (plus negative konstant, which is what I forgot before), and the hit tables will end up being far more manageable.

With multiple konstants, it's going to be tricky to actually shave much off the table. In particular, the example button with 12 k16 won't be any smaller.

Partly because there wasn't an engine yet when I wrote hit tables, I suspect they've never been profiled in comparison to the straight-up implementation. (Also, I never wrote the straight-up implementation. Premature optimization is still the root of all evil.)

At small numbers of dice, the memory cost isn't a big deal, but n! is clearly catching up to us. (I suspect the situations where it blows up are where the dice have lots of unique values.) 

We may want to pivot to straight one-v-many search at some threshold of dice, which would also allow us to profile the two against each other, particularly in the degenerate cases. (lots of dice, Golo vs nelde.)

We may also be able to optimize the data structure, but that just pushes off the issue. (Profiling again would help there.)

jl8e:
Developing some shower thoughts:

A new approach, with no real memory overhead. When originally writing this stuff, I was trying to avoid including knowledge about how specific skills worked as much as possible, and I'm willing to slip some on this now, because not slipping has proved to have too many problems. (Having written this, it turns out I haven't slipped that much.)

Use the many-vs-one iterator, which goes over all the possible subsets of attacking dice. (And may also need testing, because AFAIK it's never been used.)

With each subset:

A) min_total=0, max_total=0

B) get the list of possible values each die can contribute.
If it's length 1, that's fine, add it to both min_ and max_
If it's length 2, it's an either-or like a constant. Add the lower to min_ and the greater to max_ (Is the list guaranteed to be in order? If not, that's a needed fix)
If it's length >2, pretend it's a continuous range. (It's not always. In particular, konstant stinger dice have a hole in the middle at zero.) Add the low end to min_. and the high to max_

C) Figure out how much fire can contribute to the upper and lower ends, and adjust min_ and max_ as needed. (lower is a konstant-only situation) It's likely we should do some of this as we're adding up min_ and max_

[...]

PR #2706, opened 2021-02-18
Added 2021 Fanatics

blackshadowshade:
Addresses the 2021 part of #2346.

I have no idea whether irilyth's recipe will work correctly without the changes introduced in the custom recipe pull request, so I would suggest only deploying this to a test instance once #2389 has been accepted.

This pull request will conflict with #2666 and any other pull request that adds new buttons, ~~so I'm going to leave it as incomplete (and not fix the unit tests) until we're ready to merge it.~~


blackshadowshade:
I'd also like comment on whether bosco312's image is complete, considering that it doesn't have the button name on it.

blackshadowshade:
Bosco's image has been updated.

dwvanstone:
Is this something that can fixed for the April 2021 release?

blackshadowshade:
I've rebased this pull request and altered the responder test to account for the new number of buttons and sets.

This is ready for testing and comment.

cgolubi1:
Okay.  We're still in the merge freeze for the September release, but let's start thinking about this PR.  I'm going to stand up a replay site and play novel games using these new buttons.  Is it safe to assume there should be a dev site too?

blackshadowshade:
Yes, I think we should do some manual testing, especially with irilyth's Fanatic.

blackshadowshade:
This has been rebased to be after the switch to testing on a production site.

blackshadowshade:
And rebased to be after the BMInterface refactor.

cgolubi1:
I stood up https://fanatics.dev.buttonweavers.com/ui/ - testers, check out https://fanatics.dev.buttonweavers.com/ui/buttons.html?set=2021%20Fanatics and try some games?

dwvanstone:
I've tested the buttons - everything looks fine. Scipio & GamerXYZ gave a definite thumbs up. I've asked for the other three to review to make sure the images/descriptions/formulas are right.

blackshadowshade:
Glad to hear that. My main concern was irilyth's button. Have you played a 5 round game right to the end with his button?

dwvanstone:
> Have you played a 5 round game right to the end with his button?

I have not. I will test that.

Bosco312 noted that his background is white color rather than transparent. I don't remember who made it - I'll see if they can fix that.

cgolubi1:
RandomAI has played a bunch of novel games with the new buttons.  Since you're interested in 5-round games, i've changed its configuration so it should now play random 5-round games.  I'll let you know if i see any crashes.

dwvanstone:
Scipio requests a flavor text change for his button to include a comma after "watches": "cheese, chocolate, watches, and the rule of three.""

dwvanstone:
The site became non-responsive between 7:28AM through 7:32AM (east coast time) this morning.

dwvanstone:
I created a game (243) where the irilyth button had all 11 buttons active on its side. It seems to be working fine. I'll try adding an auxiliary button to get 12 buttons active.

dwvanstone:
Game 244 has all 12 buttons available on irilyth's side and it seems to work fine. I think and the randomized 5-game tests should be sufficient that irilyth's button doesn't break the site.

cgolubi1:
> The site became non-responsive between 7:28AM through 7:32AM (east coast time) this morning.

Thanks for the timestamp --- i'll take a look when i get a chance and see if it's anything relevant to the buttons.  Barring positive evidence, we'll assume it's a coincidence; i think the logs should say if the DB was frozen.

irilyth:
My button's recipe, art, and flavor text, all look good to me!

cgolubi1:
On the dev site, i see database hangs with timestamp:

```
[Fri Oct 01 11:51:45.290827 2021] [:error] [pid 20494] [client X.X.X.X:Y] PHP Warning:  PDO::prepare(): MySQL server has gone away in /var/www/engine /BMDB.php on line 66, referer: http://fanatics.dev.buttonweavers.com/ui/index.html?mode=monitor
```

That's not quite the same time dwvanstone reported (about 30m off), but i'm assuming it's the same issue.

At the time of the hang, a couple of irilyth vs CustomBM-with-longish recipe games were in progress:
* https://fanatics.dev.buttonweavers.com/ui/game.html?game=243 (irilyth vs `t(8) t(8) t(2) t(8) p(30) p(30) p(30) p(30) p(30)`, ran from 11:20 to 11:36 UTC)
* https://fanatics.dev.buttonweavers.com/ui/game.html?game=244 (irilyth vs `t(8) t(8) t(2) t(8) p(20) p(30) p(30) p(30) p(30) (30)`, ran from 11:37 to 12:00 UTC)

I'll run RandomAI with these exact matches --- if i can reproduce the issue reliably, that won't hurt anything.

But even before i have results from actually doing that, it seems pretty obvious that i'm going to find the problem is generation of the hit table to figure out what legal attacks one or the other of these buttons (probably irilyth, because of the stingers) actually has.

I note that in game 244, which is the game whose timestamp matches the database hang visible from the error log (see the message at the top of this comment), the attack immediately before the long timestamp gap is: `dvstest2 performed Trip attack using [t(2):1] against [Gg(32):16]; Attacker t(2) rerolled 1 => 1; Defender Gg(32) rerolled 16 => 30, was not captured`.  And this was the first attack of the round, so that means the next thing that happened was dvstest3 (playing irilyth) calculating hit table results with its full set of dice, vs the CustomBM's full set of dice, with a stinger showing 30 (and two other much smaller stingers) in the mix.  I strongly suspect i'm going to find that "stinger showing 30 with a very large number of other dice on both sides of the table" is relevant to the problem.

Definitely y'all can keep pondering how to make hit table faster --- i want to throw the question out there of how we make hit table *less intrusive*.  It's not the best if gameplay is slow for players who choose to play pathological buttons like this new irilyth recipe.  It's worse if the site hangs for those games.  And it's even worse than that if the site hangs for *all* players as a result of those pathological games.  I can think of a couple of types of solutions here:
1. Figure out how to make it so that multiple clients really can be doing database things at the same time and so that no particular database call (including, say, backups) locks out all other users.  This won't improve anything for the player of the game with the slow hit table, but it'll mean we can allow such buttons without breaking players who don't use them.
2. The hit table calculation in this case isn't actually necessary for gameplay --- it's providing a player convenience by populating the "valid attack types."  We could have voluntary hit table calculations time out after a non-intrusive amount of time, and just have the API/UI tell the player "sorry, i couldn't figure out what your valid attacks are.  Here's a list of all possible attack types in this game.  You sort it out, and i'll validate your single actual proposed attack when you send it to me."  (We'd have to sort out some failure modes the backend does need to know about, such as "what if the player has no valid attack?"  But i bet we could do that.)

cgolubi1:
I guess i should note before i go too far down this path, that we *don't* have evidence of an internal error --- as in, in the past we've seen: `PHP Fatal error:  Allowed memory size of 134217728 bytes exhausted (tried to allocate 1179648 bytes) in /var/www/engine/BMUtilityHitTable.php on line 82`.  That didn't happen here, it just hung.  I still think it's probably a hit table issue, just noting that there's one specific signature that it doesn't match.

cgolubi1:
Hmm, so i did some testing, and it occurs to me that i don't really understand the mechanism of this failure.

In particular, i did manual `BMAttackSkill::find_attack()` testing under unit test, using the attackers and defenders from the attack i called out above.  It completes in 800-1000ms with no error.  Just to see what would happen, i added another `Gg(32):30` (13th overall die), and that made the attack take around 2.9-3.0s, but still complete.  I added another `Gg(32:30)` (14th overall die), and that made the attack take around 10-11s, but still complete.  I added another `Gg(32:30)` (15th overall die), and *that* made the attack fail with a memory error, after around 25-30s.  I ran each of those tests a few times, and the results (success or failure, approximate running time) seemed consistent.

So that's all fine, but given that behavior (either `find_attack()` returns in 10 seconds or less, or fails with a memory error), why would the 12-die case succeed but hang the database for multiple minutes?  Obviously the hardware is different and that's going to matter, but why is this nonsense interfering with database access at all?

cgolubi1:
Ah, of course --- i should have checked `dmesg` on the dev site in the first place: the *apache* process was terminated by the OOM killer:

```
[Fri Oct  1 11:51:43 2021] Out of memory: Kill process 20418 (apache2) score 91 or sacrifice child
[Fri Oct  1 11:51:43 2021] Killed process 20418 (apache2) total-vm:339500kB, anon-rss:50844kB, file-rss:3800kB
```

So presumably there's some kind of cumulative effect across the various things apache is doing, that leads to (under some but not all circumstances) apache having a lower tolerance than phpunit for hitTable's allocation.

cgolubi1:
Some futzing around with `memory_get_usage()` tells me these things:
* Almost all of the memory is allocated by `generate_hit_table()` --- there's no evidence that either `list_hits()` or the individual target checks against the various defenders, adds all that much
* At the risk of stating the obvious, the problem here is both the size of the dice and the number of the dice:
    * With the 12 dice in the example, `find_attack()` allocates 33280544 bytes
    * With only the first 6 dice (which includes the `Gg(32):30`), `find_attack()` allocates almost as much, 33273704 bytes
    * With only the first 3 dice (i.e. ending with the `Gg(32):30`), `find_attack()` allocates basically nothing, 164760 bytes
* That doesn't help us much, though --- it's easy to imagine hacks that would make irilyth succeed cheaply at running these checks most of the time (basically all amounting to "don't add all the dice to the hit table at the beginning"), but that wouldn't help in the case in which there wasn't a valid skill attack (e.g. because the target dice were all immune to skill).
* I'm going to keep thinking about this, and i'm also going to collect some data about other buttons so we can compare the expense of irilyth's hit table to that of existing buttons.  I think dwvanstone managing to OOM-kill the test site, demonstrates that we need to do *something* to make this cheaper or less intrusive before we can bring this button to prod.

jl8e:
We have a plan to replace hit tables entirely with a better optimized algorithm and no precomputation, over in #2692. It's waiting on one of us being willing to turn theory into working code.

I'm not sure why the hit table computation would be making the database fall over, unless it's making the DB run out of memory.

cgolubi1:
Fair enough that there's an upcoming solution.  But i think this may be a blocker for irilyth --- i tried instantiating every button currently allowed for prod in the database with maximum side counts and values, to see how big the game-start-time hitTable is for each one, then added irilyth to that list.  Here's the list of all buttons whose hitTable is more than 100kb:

```
  103864 bytes used by Faye
  108128 bytes used by Buck
  142128 bytes used by barswanian
  144392 bytes used by Weylan
  149608 bytes used by Micro
  153832 bytes used by Cristofore
  165968 bytes used by Golo
  177408 bytes used by RavenBlack
  178384 bytes used by Tuxedo Mask
  182960 bytes used by mgatten
  183288 bytes used by Totoro
  190248 bytes used by The Jabberwock
  196456 bytes used by bluebec
  197296 bytes used by BugRoger
  246504 bytes used by Squiddhartha
  256776 bytes used by Noeh
  364328 bytes used by Aysez
  662960 bytes used by Snuff
 1333096 bytes used by kestrel
 1717496 bytes used by IIconfused
 1752296 bytes used by Notorious
 2659152 bytes used by skapheles
19924168 bytes used by irilyth
```

So, if we believe my methodology, irilyth's table size is 7.5x the next largest button (skapheles).  That just... seems like it's going to cause us problems in prod.

cgolubi1:
Here's the (extremely hackish) unit test i used to measure hit table sizes of all our buttons:

```php
    public function testHitTable_many_dice() {
        $interface = new BMInterface(TRUE);
        $buttons = $interface->get_button_data(NULL, NULL, TRUE, NULL);

        $target = BMDie::create_from_recipe("(30)");
        $target->value = 30;
        $game = new TestDummyGame;
        $game->defenderAllDieArray[] = $target;

        $buttons[] = array(
            'recipe' => 'Gg(8) Gg(16) Gg(32) `sptH(1) `sptH(2) `sptH(3) `sptH(5) `sptH(8) spohr(4,9) spohr(9,16) spohr(16,25)',
            'buttonName' => 'irilyth',
        );

        foreach ($buttons as $button) {
            if ($button['recipe'] == '') {
                continue;
            }
            $sk = new BMAttackSkill();
            $dieRecipeArray = preg_split('/[[:space:]]+/', $button['recipe']);
            foreach ($dieRecipeArray as $dieRecipe) {
                $die = BMDie::create_from_recipe($dieRecipe);
                if ($die->max === NULL) {
                    $max = 0;
                    if ($die instanceof BMDieTwin) {
                        $subdice = $die->dice;
                    } else {
                        $subdice = array($die);
                    }
                    foreach ($subdice as $subdie) {
                        if ($subdie->max > 0) {
                            $max += $subdie->max;
                        } elseif ($subdie instanceof BMDieSwing) {
                            $max += $subdie->swingMax;
                        } elseif ($subdie instanceof BMDieOption) {
                            $max += $subdie->optionValueArray[1];
                        } else {
                            error_log(var_export($die, TRUE));
                            throw Exception("help");
                        }
                    }
                    $die->max = $max;
                }
                $die->value = $die->max;
                $sk->add_die($die);
            }
            $mem_before_test = memory_get_usage();
            $sk->find_attack($game, TRUE);
            $mem_after_test = memory_get_usage();
            $mem_added_by_test = $mem_after_test - $mem_before_test;
            if ($mem_added_by_test > 100000) {
                error_log($mem_added_by_test . " bytes used by " . $button['buttonName']);
            }
            $this->assertLessThanOrEqual(3000000, $mem_added_by_test);
        }
```

cgolubi1:
In case anyone has lost the thread with all of these updates, here's my summary of what we've learned:
* dwvanstone was able to knock the fanatics dev site over for O(10m) playing an irilyth game, specifically because apache ran out of memory and had to be restarted
* i think (see questionable methodology above) that irilyth's hit table takes 7.5x as much memory as the next largest button on the prod site (skapheles)
* this seems like a big enough risk to prod site stability that we need to address it before merging the new fanatics
* the specific problem with the irilyth button is the 32-sided stinger die --- without that die, i believe the button would not be out of line with existing buttons (i haven't tested that theory, but could on request)

jl8e:
> * the specific problem with the irilyth button is the 32-sided stinger die --- without that die, i believe the button would not be out of line with existing buttons (i haven't tested that theory, but could on request)

It's mostly the eleven dice, combined with three of them being stingers.

But yeah, I think that irlyth in particular has to be held out until we can replace  hit tables.

blackshadowshade:
I would vote that we disable irilyth's button (much like we disabled mine) with a "has_button_special" flag until we're happy that we have dealt with this issue in a satisfactory manner.

As for hit tables, I'm happy to go back to looking at a way to shortcut the process, but for that, I'll still need a negative konstant responder test case so that I can find out where my previous shortcut broke the logic. Alternatively, we go with Julian's approach and entirely rewrite the hit table code from scratch.

dwvanstone:
Thank you for all of your testing. It makes sense that irilyth's button be delayed.

irilyth:
I don't have a strong feeling about this, despite being personally affected by it, but I would note that the situation in which my button would have eleven dice in play is vanishingly unlikely: It would have to be the fourth or later round of a game in which I lost the first three rounds, and brought in all three reserve dice (a terrible idea), against a button with six or more dice, and then brought in a Warrior die every turn (a terrible idea), while not losing any dice on my opponent's turn (conceivable -- maybe they have six Jolt dice! maybe they make an unsuccessful Trip Attack every turn! -- but nearly impossible in practice). In the playtesting games I've played, I don't think I've ever had more than five or six dice in play at a time, and usually one of the Stinger dice is gone after my opponent's second turn.

Did anyone ever reproduce this issue organically, or only by trying to set up degenerate cases like that?

All that said, I'm still fine with saying "we just can't take the risk", and disabling it.

cgolubi1:
> I would note that the situation in which my button would have eleven dice in play is vanishingly unlikely

Yeah, reiterating what i tried to say right after my testing yesterday: the large number of dice is a contributing factor, but the presence of several stinger dice, one of which is very large, is the bigger problem.  My test showed that your hit table would consume 19924168 bytes (7.5x as much space as the next largest button, skapheles) with all of the reserve dice enabled.  But with only the `spohr(16,25)` you're still at 4819776 (almost 2x skapheles), and with the two bigger reserve dice, you're at 9830600 (almost 4x skapheles).

And it doesn't matter whether you bring in the Warrior dice --- the hit table is used to generate *possible* attacks, and any Warrior die you could bring in is part of a possible attack.  (If the Warriors were the same dice as the Stingers, that might mitigate the hit table size, but they're not.)

jl8e:
> And it doesn't matter whether you bring in the Warrior dice --- the hit table is used to generate _possible_ attacks, and any Warrior die you could bring in is part of a possible attack. (If the Warriors were the same dice as the Stingers, that might mitigate the hit table size, but they're not.)

It occurs to me that the hit table is generating a lot of invalid attacks by attempting to bring in multiple warrior dice, and the immediate problem could be solved by restricting it, but that would be awkward and fiddly, because there isn't enough state in the hit table to easily track it, and the effort spent to change it would be better spent on ditching hit tables, and it's only a solution to lots-of-dice buttons with warrior dice, not the general case.

But if I were trying to patch it, I'd explicitly pull warrior dice out of the hit table calculation, and after the straight hit table check, loop through the warrior dice and do another check of each hit table entry plus the current warrior die.

(This would then have to be repeated in the fire calculations.)

blackshadowshade:
I thought we were already dealing with warrior dice in this way.

jl8e:
Not as far as I recall, and the behavior of irilyth's fanatic supports it. It doesn't know about specific skills at all. All it really knows is that dice can have one or more values that are added in. Dealing with mutually exclusive dice would seriously complicate the code.

Those five warriors are a factor of 32 increase in table size. (The activated reserves are another 8, and the stingers are... complicated. It's probably their current sizes multiplied together.)

blackshadowshade:
I've updated the pull request and rebased. Scipio's flavour text now has an Oxford comma, and irilyth's Fanatic has been disabled for the moment.

blackshadowshade:
Regarding bosco's image, we can update that when the version with transparent background becomes available.

dwvanstone:
I don't think we need a new dev site. There's no testing I want to do before accepting this PR.

blackshadowshade:
Image updated for Bosco312.

blackshadowshade:
No, there will be no push to the testing site. Either the testing site will be entirely redeployed (if you require it), or we wait until staging is up and take a look at it there.

dwvanstone:
We can catch any final issues with Bosco's image if they show up on the staging site. (Even if they go live we can catch them on a future release.) I believe this is ready for the next production release. (I am hopeful for these to go out before 2021 is finished...)

cgolubi1:
Okay, database sanity-check is fine.  Replay testing looks fine.  I think this is ready to merge --- i'm not going to merge it until after the prod release tomorrow, because i don't want to confuse people about whether it will be released this week (it wouldn't be regardless because it hasn't gone to staging, but if i merge it, that will confuse people).  Next week for sure, and apologies for not making 2021!  :>/

dwvanstone:
That's fine - it will be part of the celebration of 2022!!

cgolubi1:
Merging this now.

PR #2708, opened 2021-03-02
Handle database exceptions more cleanly

### LINKED ISSUE #2707: dropped connections on prod site
Since late January, i've been seeing a lot more dropped connections on the prod site.  Here's an example exception:

```
[Tue Feb 02 01:03:00.392551 2021] [:error] [pid 25191] [client X.X.X.X:Y] Caught exception in mysql.inc.php: SQLSTATE[HY000] [2006]  MySQL server has gone away, referer: $SITE/ui/index.html?mode=preference
[Tue Feb 02 01:03:00.393532 2021] [:error] [pid 25191] [client X.X.X.X:Y] PHP Notice:  Undefined variable: conn in /var/www/database/mysql.inc.php on line 46, referer: $SITE/ui/index.html?mode=preference
[Tue Feb 02 01:03:00.393664 2021] [:error] [pid 25191] [client X.X.X.X:Y] PHP Fatal error:  Uncaught Error: Call to a member function prepare() on null in /var/www/engine/BMInterface.php:1603\nStack trace:\n#0    /var/www/engine/BMInterface.php(1709): BMInterface->get_all_games('$PLAYERID', 'NEW')\n#1 /var/www/api/ApiResponder.php(263): BMInterface->get_all_new_games('$PLAYERID')\n#2 /var/www/api/ApiResponder.php(1004):  ApiResponder->get_interface_response_loadNewGames(Object(BMInterface), Array)\n#3 /var/www/api/responder.php(23): ApiResponder->process_request(Array)\n#4 {main}\n  thrown in /var/www/engine/BMInterface.php on line 1603, referer: $SITE/ui/index.html?mode=preference
```

I used to see these maybe once a month; since late January there's been at least one every day, usually more; there's always one around 1am (UTC), often they show up at other times too.  I assume this shows up as an error for players, but i'm not aware that anyone has complained so i can't verify that.

cgolubi1:
* Retry establishing the database connection a few times
* In the worst case, catch exceptions in ApiResponder to avoid returning stacktraces

* Fixes #2707

### DIFF
--- a/src/api/ApiResponder.php
+++ b/src/api/ApiResponder.php
@@ -999,18 +999,26 @@ class ApiResponder {
                 // object, invoke the function on sanitized args,
                 // and return the result
                 $sanitizedArgs = $this->spec->sanitize_function_args($args);
-                $interface = $this->create_interface($sanitizedArgs, $check);
-                apache_note('BMAPIMethod', $sanitizedArgs['type']);
-                $data = $this->{$check['funcname']}($interface, $sanitizedArgs);
-
-                $output = array(
-                    'data' => $data,
-                    'message' => $interface->message,
-                );
-                if ($data) {
-                    $output['status'] = 'ok';
-                } else {
-                    $output['status'] = 'failed';
+                try {
+                    $interface = $this->create_interface($sanitizedArgs, $check);
+                    apache_note('BMAPIMethod', $sanitizedArgs['type']);
+                    $data = $this->{$check['funcname']}($interface, $sanitizedArgs);
+                    $output = array(
+                        'data' => $data,
+                        'message' => $interface->message,
+                    );
+                    if ($data) {
+                        $output['status'] = 'ok';
+                    } else {
+                        $output['status'] = 'failed';
+                    }
+                } catch (Exception $e) {
+                    error_log('Caught unexpected exception in ApiResponder: ' . $e->getMessage());
+                    $output = array(
+                        'data' => NULL,
+                        'status' => 'failed',
+                        'message' => 'Internal error',
+                    );
                 }
             } else {
                 // found a problem with the args, report that
--- a/src/database/mysql.inc.php
+++ b/src/database/mysql.inc.php
@@ -25,23 +25,40 @@ function conn() {
         $pass = 'bmuserpass';
     }
 
-    try {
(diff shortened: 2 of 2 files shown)


cgolubi1:
If you want to see this in action, i have a dev site at `http://54.163.4.93` (sorry for the lack of DNS) that has this change and also sets the connection to NULL upon initialization 50% of the time.  Given five retries, that means an error should propagate to the caller on 1/32 of exceptions.

cgolubi1:
Here's what i'd say about what this does and doesn't accomplish:
* It should stop us from sending stacktraces to API clients, ideally ever
* The retries might or might not be effective in fixing the problem that the prod site has been having; if they're not effective, the additional logging should be helpful

Right now, the exception catching is handled in a kind of subideal way.  Specifically:
* If the connection dies on creation, we fail in `conn()`, then `BMInterface->__construct()` blows up, then the generic catcher in `ApiResponder` catches the exception
* If the connection dies later, we fail in some other function, e.g. `BMInterface->get_all_games()`, which does its own catching
As of now, this is unavoidable because we directly call `$conn->prepare()` in so many different places, and if `$conn` has become NULL, that throws a fatal error which much of our logic doesn't catch.  So we can't intentionally return NULL from `conn()`, because if we do, we know something upstream will blow up in a bad way.

But once i finish #712 and all database interactions are moved to BMDB, we can move all of the database connection retry/fail logic into BMDB, and we won't have to survey so much code checking for whether `$conn` is NULL --- we can just assume we get either a result or a database exception we threw.  So that'll be nice.

blackshadowshade:
Generally, I'm fine with the code changes here.

I wonder if it would be worthwhile adding more information to the new PDOExceptions that you are adding here to mysql.inc.php. Unless you're thinking that the timing of the error is enough to allow us to work out what's going wrong here?

cgolubi1:
Hmm.  If there's more data we have access to, i don't know what it is.  I think i made all the error messages distinct, and they're all looking at the same exception (`$conn` is NULL) at different times.  Given that, i don't have any more ideas about what would be useful.  Do you?

blackshadowshade:
I guess we can always add more diagnostics later if we get a better idea of what's going wrong and a better question to ask.

These diagnostics at least let us know if the connection is being dropped randomly, or if it always happens at the initial formation of the connection.

I'll accept the pull request, and if we need to do more, we can add to this first attempt.

Issue #1097, opened 2014-07-10
buttonweavers.com should redirect to www.buttonweavers.com


AdmiralJota:
Browsers can interpret buttonweavers.com and www.buttonweavers.com as two different sites, and can therefore maintain separate sets of cookies for the two. This can lead to weird effects, such as appearing to lose browser-specific preferences or making the user log in a second time. 

We should somehow force browsers to use one of those two domains canonically, so that visitors who try to visit the wrong one are automatically redirected to the right one, to prevent this from happening.

AdmiralJota:
I've run into this multiple times lately, with people giving URL's to other users which then don't work because they used the wrong form of the domain name.

danlangford:
confirmed this is still an issue. i usually do this on the domain level. the `whois` suggests you may be using route53. in that case i have seen this approach work a couple of times: https://medium.com/faun/redirect-naked-domain-to-www-with-https-on-aws-route-53-e620e4ce989d

also while you are at it, what ever responds to `/` does a redirect to `/ui` and that does a redirect to `/ui/` . it would be good to just have the first one redirect straight to `/ui/`

PR #2712, opened 2021-04-03
Support SSL on AWS-based buttonmen sites

### LINKED ISSUE #224: access to the site should use SSL
We should serve the website via HTTPS as well as HTTP, and we should force access to pages which send passwords back and forth (meaning the Newuser stuff, the Login stuff, and responder.php itself unless we can split out the login functions of responder into a different file) to occur via HTTPS.  The purpose is to protect passwords from being sent in cleartext.

cgolubi1:
* Fixes #224

### DIFF
--- a/deploy/vagrant/manifests/init.pp
+++ b/deploy/vagrant/manifests/init.pp
@@ -53,6 +53,7 @@ node default {
   include "postfix::base"
   include "user::buttonmen-devs"
   include "sudo::buttonmen-devs"
+  include "fqdn::base"
 
   # Node configuration needed for the buttonmen server
   include "apache::server::vagrant"
--- a/deploy/vagrant/modules/apache/manifests/init.pp
+++ b/deploy/vagrant/modules/apache/manifests/init.pp
@@ -16,6 +16,13 @@ class apache::server {
 
   # Monitor the error log
   include "apache::server::feature::monitor-logs"
+
+  # Install and configure letsencrypt (SSL/certbot) for AWS instances 
+  case "${ec2_services_partition}" {
+    "aws": {
+      include "apache::server::feature::letsencrypt"
+    }
+  }
 }
 
 class apache::server::vagrant {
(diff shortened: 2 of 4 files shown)


cgolubi1:
I put a dev site up at https://makeshift.dev.buttonweavers.com/ui/

The only thing that's complicated about this is that in order to configure LetsEncrypt/certbot, you need to know the public FQDN of the site, because that's what LetsEncrypt is going to use to reach out and verify the cert it's providing, not to mention that the FQDN is tied to the provided cert.  That caused two problems:
1. Not all sites have a public FQDN (e.g. local sandboxes, replay sites that i only access via IP, etc), and those don't need HTTPS, but they need to not break
2. Even sites that have a public FQDN don't typically know what it is, and we don't have reverse DNS for dev sites, so they can't easily look it up

I opted to solve this by putting the FQDN in an EC2 instance tag, for sites that have one that we care about (dev sites, staging, prod).  I'll add this tag one-time for staging and prod, and i updated my existing dev site tagging utility script to add it for dev sites.  If that tag can be found, we have certbot use it, and we're off to the races.  If not, we use the dummy sandbox.buttonweavers.com name, and don't try to configure certbot.

I tested this on virtualbox, and on the aforementioned dev site, and it worked on both.

The only thing that's a little irritating is that it takes a few minutes for instance tags to become consistent and thus queryable by a new instance, so typically a new dev site doesn't get its fqdn the first time.  That's not a big deal --- with the current setup, provisioning succeeds (it just doesn't setup SSL), and rerunning `vagrant provision` will make the right thing happen without intervention.  So i'm not going to worry about that.

Folks should check out the dev site and see if anything behaves badly/oddly.  It should work just fine both at https:// and at http://

blackshadowshade:
I've taken a look at this. I can play a move on the https site, view a discussion thread, log out, and log back in---all pages remaining as https://...

Similarly, when I log in under http, everything remains http and is responsive.

I'm happy to try this out.

jl8e:
If I go to buttonweavers.com, not www.buttonweavers.com, I get a certificate mismatch.

blackshadowshade:
Maybe this should be linked to #1097, which deals with somewhat similar issues.

PR #2725, opened 2021-04-25
Don't use the ec2_placement_availability_zone variable unless we're on an EC2 instance

cgolubi1:
I thought i implemented the FQDN stuff in a way that would keep working on virtualbox, but i got my wires crossed and accidentally broke it.  This one-liner fixes it (so virtualbox sites come up, which is the important part, and define the default/fake FQDN `sandbox.buttonweavers.com` and thus don't try to configure HTTPS).


blackshadowshade:
Let's give this a go, it only affects one deployment file and CircleCI passes.

PR #2720, opened 2021-04-14
Added images for Bridge and Tunnel buttons

blackshadowshade:
Fixes #2702.

PR #2727, opened 2021-05-10
Revert "Added images for Bridge and Tunnel buttons"

cgolubi1:
Reverts buttonmen-dev/buttonmen#2720

I think i merged that commit by accident typing commands in the wrong directory of my workspace.  It's a little hard to imagine how that could have happened, but it seems like it did.  Apologies.


cgolubi1:
Hmm, that said: i think we were okay merging the commit.  It's a pity i didn't rebase, but i don't see how i'm going to make less of a mess of the tree by reverting and then unreverting.  So i think i'll leave well enough alone and call it merged.

PR #2730, opened 2021-06-14
Upgrade various modules so that jshint can run/pass

### LINKED ISSUE #2728: grunt broke again
Tests are failing again with another grunt/npm error, see e.g. https://circleci.com/workflow-run/3e0ec22d-d70e-40ac-ad97-1c3a6e4b6605

cgolubi1:
* Fixes #2728 

Simply upgrading various package versions addresses the 404 errors, but yields:
```
Running "jshint:all" (jshint) task

   ../../src/ui/js/Env.js
      3 |  var Env = {};
               ^ 'Env' was used before it was defined.

>> 1 error in 25 files
Warning: Task "jshint:all" failed. Use --force to continue.
```

Setting `"latedef": false` in our preferences would make that error go away, so we could always fall back to that if this doesn't work.

But looking at it, the *only* problem is `Env`, and the only reason we have to check whether `Env` is defined before defining it, is that the JS unit tests use an `Env` property to define the fact that they are running under unit test.  That's kludgy in itself and we could probably do better, but it doesn't really matter --- as far as i can tell, we can initialize `Env` unconditionally, then only define it if it's undefined, and that makes everyone happy:
* jshint passes
* I can still browse to the UI when i bring up a vagrant site
* QUnit tests still pass interactively on my vagrant site and via CircleCI

So afaict this solves the problem, and gets us back into a state where builds can pass.

### DIFF
--- a/src/ui/js/Env.js
+++ b/src/ui/js/Env.js
@@ -1,6 +1,7 @@
 // namespace for this "module"
-if (!(Env)) {
-  var Env = {};
+var Env;
+if (!Env) {
+  Env = {};
 }
 
 // Keycodes for use when handling keyboard events
--- a/util/grunt/package.json
+++ b/util/grunt/package.json
@@ -3,11 +3,11 @@
   "version": "0.1.0",
   "devDependencies": {
     "grunt": "~0.4.1",
-    "grunt-contrib-jshint": "~0.4.3",
+    "grunt-contrib-jshint": "~1.0.0",
     "grunt-contrib-nodeunit": "~0.1.2",
     "grunt-contrib-uglify": "~0.2.0",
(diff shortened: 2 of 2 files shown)


blackshadowshade:
Well, first things first: CircleCI passes!

https://app.circleci.com/pipelines/github/cgolubi1/buttonmen/76/workflows/8e88f660-922c-4e3e-bbc7-053c283cba22/jobs/332

blackshadowshade:
Also, the potential issue I thought of, namely declaring a variable in Javascript as var multiple times appears to be unproblematic: https://stackoverflow.com/questions/21726948/declaring-a-javascript-variable-twice-in-same-scope-is-it-an-issue

So I think that the change looks fine.

Accepting now.

PR #2731, opened 2021-06-17
Convert BMInterfaceGameAction to use BMDB instead of direct database actions

cgolubi1:
* Partially addresses #712


cgolubi1:
This may not pass CircleCI because of formatting/linting type issues --- if it doesn't, i'll fix it tomorrow and rebase.

Assuming it does pass, i think the responder tests do a sufficiently thorough job of updating/polling action log entries, that passing tests can be taken as evidence that this isn't a regression.  Let me know what you think.

blackshadowshade:
CircleCI passes and the code changes make sense to me.

I agree that our code coverage of action logging is pretty extensive, and I'm happy to merge this now.

PR #2729, opened 2021-05-10
Replace gender with pronouns in preferences and profiles

cgolubi1:
* Fixes #2709

PR #2732, opened 2021-06-17
Replace gender with pronouns in preferences and profiles

### LINKED ISSUE #2709: Add pronouns to player profiles, and improve gender options
I see this was discussed back in #1068, but I think the gender field needs a bit of revisiting and supplementing.

Nonbinary people are common enough that I think they should have their own entry in the dropdown.

I also think that "It's complicated" opening up a text box isn't obvious, and something like "Enter your own" would be better.

Whether or not we adjust the gender field, I think it's worthwhile to have a separate pronoun field.

Like gender, it'd be a customizable text field.

Off the top of my head, I'd include the following options:

Unspecified
she/her
he/him
they/them
it/its (An actual robot)
none (use my name)
Custom

I'd probably also add a couple of the more common other pronouns (xie/hir), but that requires a bit more research.

cgolubi1:
* Fixes #2709 
* This PR should be identical to #2729 but rebased and with upstream merged so it can pass tests


cgolubi1:
Sorry for the extra branch; my buttonmen git workflow has a lot of cruft in it from when we started out and didn't really know how to use git, so of course i got my branch into a broken state.  I'm sure i could fix it eventually, but this is much faster for me, and i suspect it isn't any slower for you, since you have the two PRs available to verify that the rebase is the same code.

blackshadowshade:
Looks good to me.

@cgolubi1 , for reference, my workflow when rebasing is:

    git checkout master
    git merge upstream/master
    git checkout <branch_of_interest>
    git rebase -i master
    git push origin HEAD -f

cgolubi1:
Thanks!  That helps.

blackshadowshade:
Just updated workflow to add the requisite "git fetch --all" at the beginning, which is of course important.

Issue #2749, opened 2021-08-15
Missing Swing Die after setting its value


TheOrgg:
http://www.buttonweavers.com/ui/game.html?game=76655

In a Hit List game, I set my X to 4, but it didn't appear in my recipe.

blackshadowshade:
It looks like the die has disappeared completely!

That's definitely a bug.

Can you tell us if the same happens next round too?

TheOrgg:
Will do.  Is there something in the background that can tell you what happened?

jl8e:
Just speculating here:

(2021-08-14 22:20:14) | ElihuRoot set swing values: X=16
(2021-08-14 22:20:13) | TheOrgg set swing values: X=4

Is a race condition possible?

jl8e:
Elaborating a bit more:

Without going diving into the guts of the engine, I think there are two potential race types. One is easy to test, but unfortunately very unlikely to be the cause. The other is a pain to test.

The easy one:

TheOrgg accepts the game, goes into swing die selection. Game state is currently no swings set.
ElihuRoot goes to the game, goes into swing die selection. Game state is currently no swings set.
TheOrgg submits their swing die choice. Game state updates.
ElihuRoot submits their swing die choice. Game state updates, but there's something being set incorrectly because the request was made by a client that thinks there are no swing dice set.

As I said, I think this is very unlikely. I haven't checked, but I'd assume that the client has no idea whether the opponent has set their swings or not, and the request is identical in either case.

The hard one:

I honestly don't know enough about the guts of how the code interacts with the DB, or how the DB behaves (I believe the engine is single-threaded, but is the DB?) to speculate on the mechanics at this level, but it seems not impossible that some of the various DB activities from TheOrgg's request were still pending when ElihuRoot's request was processed.

cgolubi1:
It's definitely possible.  Here's the apache log:

```
X.X.X.X - - [15/Aug/2021:02:19:51 +0000] "POST /api/responder HTTP/1.1" 200 3219 "http://www.buttonweavers.com/ui/game.html?game=76655" "<BROWSER>" BMUser=<ElihuRoot> BMAPI=loadGameData BMStatus=ok
[...]

PR #2750, opened 2021-08-19
Wrapped save_game in a transaction

blackshadowshade:
This pull request attempts to address the race condition that we think is occurring in #2749 by wrapping almost all of BMInterface->save_game() in a transaction.

I'm guessing that we'll just want to let RandomAI test this a bit to see if this changes any behaviour.

### DIFF
--- a/src/engine/BMInterface.php
+++ b/src/engine/BMInterface.php
@@ -888,6 +888,11 @@ class BMInterface {
         // force game to proceed to the latest possible before saving
         $game->proceed_to_next_user_action();
 
+        // start transaction
+        if (!$this->isTest) {
+            self::$conn->beginTransaction();
+        }
+
         try {
             $this->resolve_random_button_selection($game);
             $this->save_basic_game_parameters($game);
@@ -912,12 +917,19 @@ class BMInterface {
             $this->delete_dice_marked_as_deleted($game);
             $this->game_action()->save_action_log($game);
             $this->game_chat()->save_chat_log($game);
+
+            if (!$this->isTest) {
+                self::$conn->commit();
+            }
         } catch (Exception $e) {
             error_log(
                 'Caught exception in BMInterface::save_game: ' .
                 $e->getMessage()
             );
             $this->set_message("Game save failed: $e");
+            if (!$this->isTest) {
+                self::$conn->rollBack();
+            }
         }
     }
 
--- a/src/engine/BMInterface.php
+++ b/src/engine/BMInterface.php
@@ -83,7 +83,11 @@ class BMInterface {
         } else {
             require_once __DIR__.'/../database/mysql.inc.php';
         }
-        self::$conn = conn();
+
+        if (!isset(self::$conn)) {
+            self::$conn = conn();
+        }
+
         self::$db = new BMDB(self::$conn);
     }
 
@@ -889,9 +893,7 @@ class BMInterface {
         $game->proceed_to_next_user_action();
 
         // start transaction
-        if (!$this->isTest) {
-            self::$conn->beginTransaction();
-        }
+        self::$conn->beginTransaction();
 
         try {
             $this->resolve_random_button_selection($game);
@@ -918,18 +920,14 @@ class BMInterface {
             $this->game_action()->save_action_log($game);
             $this->game_chat()->save_chat_log($game);
 
-            if (!$this->isTest) {
-                self::$conn->commit();
-            }
+            self::$conn->commit();
         } catch (Exception $e) {
             error_log(
                 'Caught exception in BMInterface::save_game: ' .
                 $e->getMessage()
             );
             $this->set_message("Game save failed: $e");
-            if (!$this->isTest) {
-                self::$conn->rollBack();
-            }
+            self::$conn->rollBack();
         }
     }
 


src/engine/BMInterface.php
```diff
@@ -888,6 +888,11 @@ protected function save_game(BMGame $game) {
         // force game to proceed to the latest possible before saving
         $game->proceed_to_next_user_action();
 
+        // start transaction
+        if (!$this->isTest) {
```
cgolubi1, line 892: Remind me why we shouldn't do this for test invocations?  I thought the kinds of tests that came through apiResponder used a real database --- so why shouldn't they have transactions too?

cgolubi1:
Anyway, i kicked off a replay_loop, and let's see what happens.

src/engine/BMInterface.php
blackshadowshade: The simple answer is that things failed when I didn't have this conditional in place.

Unfortunately, I have no insight into why this is, I probably don't understand the test responder correctly.

Maybe you understand better?

cgolubi1: I don't, but what i do understand is that the `isTest` flag should do literally one thing: tell the code to use the buttonmen_test database rather than the buttonmen database.  I don't see any reason any of this stuff should be magical in responderTest, and the symptoms here are very weird.  Here i'm running a single responderTest, `test_interface_game_051`, which does a very normal `create_game` with Dead Guy and Gawaine --- nothing should be weird about this.

It fails like so:

```
Caught exception in BMInterface::save_game: There is no active transaction
Caught exception in BMInterface::create_game: There is no active transaction
F                                                                  2 / 2 (100%)

Time: 1.52 seconds, Memory: 10.00Mb

There was 1 failure:

1) responder03Test::test_interface_game_051
API call should succeed:
ARGS: array (
  'type' => 'createGame',
  'playerInfoArray' => 
  array (
    0 => 
    array (
      0 => 'responder003',
      1 => 'Dead Guy',
    ),
    1 => 
    array (
      0 => 'responder004',
      1 => 'Gawaine',
    ),
  ),
  'maxWins' => 3,
  'description' => '',
)
RETURN: array (
  'data' => NULL,
  'message' => 'Game create failed: There is no active transaction',
  'status' => 'failed',
)
--- Expected
+++ Actual
@@ @@
-'ok'
+'failed'

/buttonmen/test/src/api/responderTestFramework.php:986
/buttonmen/test/src/api/responderTestFramework.php:1022
/buttonmen/test/src/api/responder03Test.php:3004

FAILURES!
Tests: 2, Assertions: 28, Failures: 1.
```

I annotated BMInterface with print statements, so i could see whether the transaction never gets started or whether we fall out of it or what --- in these, if the line ends with 1, `self::$conn->inTransaction()` succeeded; if not, it failed:

```
.entering create_game try loop
create_game: insert_new_game() happened
create_game: added players
create_game: set random buttons and custom recipes
create_game: loaded game
create_game: about to save game
in save_game()
save_game: done with proceed_to_next_user_action
save_game: done with beginTransaction: 1
save_game: done with resolve_random_button_selection: 1
save_game: done with save_basic_game_parameters: 1
save_game: done with save_button_recipes: 1
save_game: done with save_random_button_choice: 1
save_game: done with save_round_scores: 1
save_game: done with clear_swing_values_from_database: 1
save_game: done with clear_option_values_from_database: 1
save_game: done with save_swing_values_from_last_round: 1
save_game: done with save_swing_values_from_this_round: 1
save_game: done with save_option_values_from_last_round: 1
save_game: done with save_option_values_from_this_round: 1
save_game: done with save_player_game_decisions: 1
save_game: done with save_player_with_initiative: 1
save_game: done with save_players_awaiting_action: 1
save_game: done with regenerate_essential_die_flags: 1
save_game: done with mark_existing_dice_as_deleted: 1
save_game: done with save_active_dice: 1
save_game: done with save_captured_dice: 1
save_game: done with save_turbo_cache: 1
save_game: done with save_out_of_play_dice: 1
save_game: done with delete_dice_marked_as_deleted: 1
save_game: done with save_action_log: 
save_game: done with save_chat_log: 
save_game: about to commit: 
```

So we exited the transaction while calling `save_action_log()`?  Why?  I don't have an immediate answer, but i definitely think we need to know --- the *point* of unit tests is to catch all the weird things that might happen as we call functions in different ways --- RandomAI literally exercises BMGame logic.  It doesn't help us with the interface at all.

Even if this is a bug in a test harness, committing this is too much of a divergence between test behavior and live behavior --- it'll make our tests much more useless for catching any future problems with how we use the DB.  So i think we need to figure it out.

cgolubi1: Okay, i've learned a little more by instrumenting the part where it creates new PDO connections:

```
About to create new test PDO connection
entering create_game try loop
create_game: insert_new_game() happened
create_game: added players
create_game: set random buttons and custom recipes
create_game: loaded game
create_game: about to save game
in save_game()
save_game: done with proceed_to_next_user_action
save_game: done with beginTransaction: 1
save_game: done with resolve_random_button_selection: (1); (2)
save_game: done with save_basic_game_parameters: (1); (2)
save_game: done with save_button_recipes: (1); (2)
save_game: done with save_random_button_choice: (1); (2)
save_game: done with save_round_scores: (1); (2)
save_game: done with clear_swing_values_from_database: (1); (2)
save_game: done with clear_option_values_from_database: (1); (2)
save_game: done with save_swing_values_from_last_round: (1); (2)
save_game: done with save_swing_values_from_this_round: (1); (2)
save_game: done with save_option_values_from_last_round: (1); (2)
save_game: done with save_option_values_from_this_round: (1); (2)
save_game: done with save_player_game_decisions: (1); (2)
save_game: done with save_player_with_initiative: (1); (2)
save_game: done with save_players_awaiting_action: (1); (2)
save_game: done with regenerate_essential_die_flags: (1); (2)
save_game: done with mark_existing_dice_as_deleted: (1); (2)
save_game: done with save_active_dice: (1); (2)
save_game: done with save_captured_dice: (1); (2)
save_game: done with save_turbo_cache: (1); (2)
save_game: done with save_out_of_play_dice: (1); (2)
save_game: done with delete_dice_marked_as_deleted: (1); (2)
About to create new test PDO connection
in save_action_log(): (); (2)
after load_message_from_game_actions(): (); (2)
after log_game_actions(): (); (2)
save_game: done with save_action_log: (); (2)
About to create new test PDO connection
save_game: done with save_chat_log: (); (2)
save_game: about to commit: (); (2)
Caught exception in BMInterface::save_game: There is no active transaction
Caught exception in BMInterface::create_game: There is no active transaction
```

So that explains it --- `save_action_log()` and `save_chat_log()` each lead to a new conn being created.  That's clearly not going to end well.

cgolubi1: Here you go --- try this diff:

```diff
diff --git a/src/engine/BMInterface.php b/src/engine/BMInterface.php
index 116f746..a4fe0a5 100644
--- a/src/engine/BMInterface.php
+++ b/src/engine/BMInterface.php
@@ -83,8 +83,10 @@ class BMInterface {
         } else {
             require_once __DIR__.'/../database/mysql.inc.php';
         }
-        self::$conn = conn();
-        self::$db = new BMDB(self::$conn);
+        if (!(isset(self::$conn))) {
+            self::$conn = conn();
+            self::$db = new BMDB(self::$conn);
+        }
     }
 
     /**
```

If you make that change, all the responderTest tests pass.  There are still a bunch of failures in non-responder tests --- i'm out of time right now, but let's look into those.

blackshadowshade: Right, that's a great start.

blackshadowshade: I've checked with a similar instrumentation to you, and the issue with the BMInterface tests is now something different. The transactions are apparently completing successfully. I'll keep digging.

blackshadowshade: I reckon it might be a timing issue, somehow.

I'm looking at the first error, which for me occurs in BMInterfaceForumTest->test_create_load_forum_posts().

The first line is

    $overview = $this->object->load_forum_overview(self::$userId1WithoutAutopass);

This returns with

    array(2) {
      ["boards"]=>
      array(0) {
      }
      ["timestamp"]=>
      int(1629510278)
    }

So, somehow, the boards aren't there to load. I don't know exactly what preparation goes on before this, but it looks like the test database is empty.

blackshadowshade: Also, I checked, and self::$userId1WithoutAutopass looks like it has a valid value, so it's not a failure in the creation of the users.

blackshadowshade: Okay, making progress. The PHP error log shows this error:

    Caught exception in BMInterface::save_game: DB select_single_value found no result

blackshadowshade: I think that part of the issue is that self::$conn is an object of class TestDummyPDOConn. I'll try adding function definitions for the transaction-related functions, see if that fixes things.

blackshadowshade: No, that was a confounding factor (that still needed fixing).

blackshadowshade: The problem is that under test conditions, the BMInterface* objects are ending up as TestDummyPDOConn instead of PDO, because they're hitting the !isset(self::$conn) condition and then failing to create a proper PDO connection.

blackshadowshade: The BMDB::$conn has class TestDummyPDOConn instead of PDO, which is causing the problem.

blackshadowshade: And fixed! Just had to shift the condition brackets from Chaos's suggestion, and it all works. Pushing now.

cgolubi1:
Standing up a new replay site for this.  (FWIW, the existing replay site has been spamming me with errors, which is good evidence that there's nothing special about the unit tests --- those action log issues would have blown up on a real site as well):

```
[Sun Sep 05 00:36:37.829911 2021] [:error] [pid 14995] [client 127.0.0.1:33872] Caught exception in BMInterface::create_game: There is no active transaction   
[Sun Sep 05 00:36:38.941197 2021] [:error] [pid 14998] [client 127.0.0.1:33890] Caught exception in BMInterface::save_game: There is no active transaction     
```

cgolubi1:
As far as i can tell, this is doing fine on replay --- i'm having a little bit of trouble clearing the remnants of the RandomAI CustomBM support business out of the replay system, so i'm definitely planning to keep running this until after the end of the merge freeze, but it looks good to me.

cgolubi1:
This has successfully replayed at least 5800 games, and i don't see any issues.  I think we should move forward with it.

I'm curious about whether the rollback works as advertised, but at a glance i don't think we've caught an exception in `BMInterface::save_game()` so far this year, with the exception of:
* exceptions introduced in replay testing by the initial version of this PR
* temporary failures caused by the database going away, which now get retried

I'm inclined to merge this and see what happens in terms of the temporary failure case, rather than try to invent a way to reproduce that issue not-in-prod, which as you may recall i had some trouble with when we initially developed it.

Issue #2762, opened 2021-09-12
Game creation broken by shifted code that checks custom recipes


blackshadowshade:
#2758 broke game creation generally.

We'll need to fix this.

blackshadowshade:
So, using a defined button against another defined button works okay. It appears that this is an issue with random buttons.

PR #2766, opened 2021-09-12
Run most responderTests on a production site_type

cgolubi1:
This is my proposed fix for the fact that no tests alerted us to #2762

Had we had this change in place when i introduced the initial syntax for the CustomBM change, responder tests would have lit up:

```
$ sudo /usr/local/bin/create_buttonmen_databases && sudo -u vagrant phpunit --bootstrap /usr/local/etc/buttonmen_phpunit.php /buttonmen/test/src/api/
...
FAILURES!
Tests: 107, Assertions: 625, Failures: 71, Skipped: 5, Incomplete: 3.
```

### DIFF
--- a/test/src/api/responder00Test.php
+++ b/test/src/api/responder00Test.php
@@ -12,6 +12,10 @@ require_once __DIR__.'/responderTestFramework.php';
 
 class responder00Test extends responderTestFramework {
 
+    public function test_init_database_defaults() {
+        $this->update_config_site_type('production');
+    }
+
     public function test_request_invalid() {
         $args = array('type' => 'foobar');
         $retval = $this->verify_api_failure($args, 'Specified API function does not exist');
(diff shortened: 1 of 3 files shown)


cgolubi1:
If this is approved, i'm going to have to followup with an update to RandomAI to play nicely with this, given that RandomAI takes advantage of responderTest for replay, intentionally uses a development site-type, and does test with CustomBM.  But i think i'll wait to see if you like the approach and want to merge this before worrying about that.

blackshadowshade:
Ah, you got there first. I realised that this was the problem early this morning.

blackshadowshade:
Okay, I've taken a look at the code here and it looks perfect.

I note though that it's going to induce conflicts with all the pull requests that add new buttons, so we'll need to be careful about deconflicting those properly once this has been accepted.

cgolubi1:
Okay, i haven't heard of any new problems in the past two days, so i think we can say that the post-prod-release merge freeze A. has demonstrated its value, and B. is now over.  So feel free to go ahead and merge this if it looks good, and i'll move forward with making replay testing happy with it.

blackshadowshade:
Okay, I'll merge this now and then start working on getting all the other pull requests rebased so that we can deal with the shift in the number of buttons and sets

PR #2770, opened 2021-09-29
save_game should throw an exception when it fails to save the game

### LINKED ISSUE #2769: something not quite right with transactions
Replay testing encountered a failure with the master codebase replaying games recorded on the master codebase, which of course shouldn't happen.

The failure is the game was expected to be in `SPECIFY_DICE`, but was found to be in `START_TURN`:
```diff
1) responder99Test::test_interface_game_00076
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
     'dieBackgroundType' => 'realistic'
-    'gameState' => 'SPECIFY_DICE'
-    'gameActionLogCount' => 2
-    'playerWithInitiativeIdx' => null
-    'activePlayerIdx' => null
+    'gameState' => 'START_TURN'
+    'gameActionLogCount' => 4
+    'playerWithInitiativeIdx' => 0
+    'activePlayerIdx' => 0
     'gameSkillsInfo' => Array (...)
     'gameActionLog' => Array (
         0 => Array (
             'timestamp' => 'TIMESTAMP'
-            'player' => 'responder004'
-            'message' => 'responder004 set die sizes'
+            'player' => ''
+            'message' => 'responder003 won initiative for round 1. Initial die values: responder003 rolled [(4):4, s(8):1, (8):2, (10):2, s(X=12):12], responder004 rolled [%Ho(1):1, pdw(7):7, mhv(13):13, zkt(23):7, (X=4,X=4)!:5]. responder004 has dice which are not counted for initiative due to die skills: [pdw(7), zkt(23)].' 
         )
         1 => Array (
             'timestamp' => 'TIMESTAMP'
-            'message' => 'Game created by responder003'
-            'player' => ''
+            'message' => 'responder003 set swing values: X=12'
+            'player' => 'responder003'
         )
+        2 => Array (...)
+        3 => Array (...)
```

Here's the relevant piece of the generated responderTest file:

```php
        $retval = $this->verify_api_loadGameData($expData, $gameId, 10);

        $_SESSION = $this->mock_test_user_login('responder004');
        $this->verify_api_submitDieValues(
            array(3, 2),
            $gameId, 1, array('X' => 4), NULL);

        array_unshift($expData['gameActionLog'], array('timestamp' => 'TIMESTAMP', 'player' => 'responder004', 'message' => 'responder004 set die sizes'));
        $expData['gameActionLogCount'] = 2;
        $expData['playerDataArray'][1]['waitingOnAction'] = false;

        $retval = $this->verify_api_loadGameData($expData, $gameId, 10);

        $this->verify_api_submitDieValues(
            array(12),
            $gameId, 1, array('X' => 12), NULL);


        $retval = $this->verify_api_loadGameData($expData, $gameId, 10);

        $this->verify_api_submitDieValues(
            array(4),
            $gameId, 1, array('X' => 18), NULL);
```

In other words, responder004 submits die values, and then responder003 submits die values twice.  Whoops.

The apache log confirms that this was a transaction failure thing:
```
[Wed Sep 29 00:01:02.125935 2021] [:error] [pid 13732] [client 127.0.0.1:54196] Caught exception in BMInterface::save_game: SQLSTATE[40001]: Serialization failure: 1213 Deadlock found when trying to get lock; try restarting transaction
```

Here's the issue --- according to the pck file recorded by the initial game, the API returned success, not failure:
```python
# responder004's successful set
>>> d[5]; d[5]['retval'].status; d[5]['retval'].message
{'roundNumber': 1, 'type': 'submitDieValues', 'swing_array': {u'X': 4}, 'retval': <bmapi.BMAPIResponse instance at 0x7f2f7ef47128>, 'option_array': {}}
u'ok'
u'Successfully set die sizes'

# responder003's *failed* set
>>> d[9]; d[9]['retval'].status; d[9]['retval'].message
{'roundNumber': 1, 'type': 'submitDieValues', 'swing_array': {u'X': 12}, 'retval': <bmapi.BMAPIResponse instance at 0x7f2f7ef6d0e0>, 'option_array': {}}
u'ok'
u'Successfully set die sizes'

# responder003's second, successful, set
>>> d[12]; d[12]['retval'].status; d[12]['retval'].message
{'roundNumber': 1, 'type': 'submitDieValues', 'swing_array': {u'X': 18}, 'retval': <bmapi.BMAPIResponse instance at 0x7f2f7eefe290>, 'option_array': {}}
u'ok'
u'Successfully set die sizes'
```

So this isn't a disaster, but it's not right; the database update failed, so responder003 should have received an API failure message.

cgolubi1:
* Fixes #2769 

This isn't a very exciting fix, but it should solve the general problem of preventing the backend from reporting success to players when game save fails (or at least push it one layer up the stack; the calling function could catch and ignore exceptions, but i believe most API functions should return failure when they catch an exception.  Definitely `submit_die_values()` does the right thing, and that's the case for which i filed the issue.

### DIFF
--- a/src/engine/BMInterface.php
+++ b/src/engine/BMInterface.php
@@ -965,6 +965,7 @@ class BMInterface {
             );
             $this->set_message("Game save failed: $e");
             self::$conn->rollBack();
+            throw new BMExceptionDatabase("Game save failed");
         }
     }
 


cgolubi1:
I didn't add a unit test around this, but i did hand-test it to make sure the new behavior is reasonable, by hardcoding an exception into one of the functions save_game calls.  Then when i made an attack, the apache log gets:

```
[Wed Sep 29 23:49:54.307749 2021] [:error] [pid 23232] [client 10.0.2.2:55880] Caught exception in BMInterface::save_game: waaaugh, referer: http://127.0.0.1:8080/ui/game.html?game=1
[Wed Sep 29 23:49:54.310619 2021] [:error] [pid 23232] [client 10.0.2.2:55880] Caught exception in BMInterface::submit_turn: Game save failed, referer: http://127.0.0.1:8080/ui/game.html?game=1
```

and the UI gets:

internal_error

By contrast, with the hardcoded internal error but without this fix, the player sees:

messy_internal_error

So i think the behavior with the fix is better!  Especially since you can tell from the fact that the stack trace string is shown to the player in green, that the API status is `ok`, so we lose there too.

blackshadowshade:
If you're okay with the thrown exception, I'm okay with this fix. Let's do it.

PR #2771, opened 2021-10-01
Remove remaining unnecessary int casts from game/recipe/die code

### LINKED ISSUE #712: all number inputs to the backend should be cast to int immediately

cgolubi1:
* Refactor die parsing slightly so that validate_die_size() does all interaction with the die size as a string, including casting it to an int
* Also remove some bool casts which are not necessary under BMDB

* Fixes #712 (finally!  i think!)

### DIFF
--- a/deploy/circleci/audit_php_files.php
+++ b/deploy/circleci/audit_php_files.php
@@ -41,17 +41,8 @@
     "src/api/ApiSpec.php" => 2,
     // exceptions for the three int types BMDB supports on select
     "src/engine/BMDB.php" => 3,
-
-    // arbitrarily high exceptions because we are not yet limiting casts in these files (#712)
-    "src/engine/BMAttack.php" => 1000,
-    "src/engine/BMDie.php" => 1000,
-    "src/engine/BMDieSwing.php" => 1000,
-    "src/engine/BMGame.php" => 1000,
-    "src/engine/BMPlayer.php" => 1000,
-    // This file is converted to BMDB, but we haven't yet done the exercise of ensuring that
-    // subdie values reconstructed from flags are always ints.
-    // Leave those casts in place until we do
-    "src/engine/BMInterface.php" => 2,
+    // exceptions for validate_die_size(), which parses string die recipe into int sides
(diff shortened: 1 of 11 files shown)


cgolubi1:
I'll kick off some replay testing as soon as i have a chance.  Shadowshade, anything you can see offhand that you think that won't catch?

cgolubi1:
Ah, the UI test failures are real, because this change causes the backend to pass `optionArray` as ints rather than strings.  Let me take a closer look at that.  I think that change is *good*; i just need to check that the UI can handle it.

cgolubi1:
Yeah, this looks fine; i hand-played some rounds of Fraquito vs Zoe, and everything seemed to work just fine.  Note that the responder tests have always put ints in `optRequestArray`, because PHP's equality-checking can't tell the difference (a sentence which really explains a lot of how we've wound up having so much variable type pain over the years, but never mind).

blackshadowshade:
This all looks good to me.

Once you're happy with automated testing, I'm happy to merge this.

cgolubi1:
I never actually did the replay testing i said i'd do on this.  Let's kick that off now.

cgolubi1:
This has replayed 8900 games without issue, and i don't think there's anything in particular we were worried about.  Shadowshade, if you're still fine with this, want to merge it when you have a chance?  (I'm checking in again rather than merging it myself only because it's been a couple of months, not because of any specific concerns.)

blackshadowshade:
This is fine, merging this now.

PR #2775, opened 2021-10-05
Reversed order of open games

### LINKED ISSUE #2774: Open games page should display more recent games first
As discussed recently on the forums. There's the reasonable assumption that people will have seen the older games already, and chosen not to play them at that time. They can always scroll a bit further if they want to.

blackshadowshade:
Fixes #2774.

### DIFF
--- a/src/engine/BMInterface.php
+++ b/src/engine/BMInterface.php
@@ -1834,7 +1834,7 @@ class BMInterface {
                     'INNER JOIN game_player_view AS v_victim ' .
                         'ON v_victim.game_id = g.id AND v_victim.player_id IS NULL ' .
                 'WHERE s.name = "OPEN"' .
-                'ORDER BY g.id ASC;';
+                'ORDER BY g.id DESC;';
             $parameters = array();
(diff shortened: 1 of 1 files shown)


cgolubi1:
Dev site at https://blip.dev.buttonweavers.com/ --- someone check this out and see if it seems fine?

dwvanstone:
I've tried two browsers and I'm unable to connect to https://blip.dev.buttonweavers.com/ .

cgolubi1:
Thanks for pointing it out.  I bet there are still some problems with my new vagrant setup, taking a look.

cgolubi1:
Okay, try now, should be fixed.  This is the usual old problem with needing to run puppet twice to get letsencrypt to do its thing; i should try to solve that at some point, because clearly "i'll just remember" is not a good plan.

dwvanstone:
I'm in and I'm testing!

dwvanstone:
This works as expected. I think it'll be a nice feature.

cgolubi1:
Merging this now.

PR #2776, opened 2021-10-06
Refactor BMUtilityHitTable to use less memory

cgolubi1:
Pass a list of target values at hit table creation time, and only
store combos which (either directly or with assistance) can hit one
of those targets

### DIFF
--- a/src/engine/BMAttackSkill.php
+++ b/src/engine/BMAttackSkill.php
@@ -64,7 +64,7 @@ class BMAttackSkill extends BMAttack {
      *
      * @param bool $includeOptional
      */
-    protected function generate_hit_table($includeOptional = TRUE) {
+    protected function generate_hit_table($includeOptional = TRUE, $hitVals = NULL) {
         if ($includeOptional) {
             $validDice = $this->validDice;
         } else {
@@ -80,7 +80,7 @@ class BMAttackSkill extends BMAttack {
 
         self::strip_excess_plain_zeros($validDice);
 
-        $this->hitTable = new BMUtilityHitTable($validDice);
+        $this->hitTable = new BMUtilityHitTable($validDice, $hitVals=$hitVals);
     }
 
     /**
@@ -124,7 +124,12 @@ class BMAttackSkill extends BMAttack {
             return FALSE;
         }
 
-        $this->generate_hit_table($includeOptional);
+        $dvals = array();
+        foreach ($targets as $t) {
+            $dvals[] = $t->defense_value($this->type);
+        }
+
+        $this->generate_hit_table($includeOptional, $hitVals=$dvals);
         $hits = $this->hitTable->list_hits();
 
         foreach ($targets as $t) {
@@ -225,7 +230,7 @@ class BMAttackSkill extends BMAttack {
         $dval = $defenders[0]->defense_value($this->type);
 
         if (!($this->hitTable instanceof BMUtilityHitTable)) {
-            $this->generate_hit_table();
+            $this->generate_hit_table($hitVals = array($dval));
         }
 
         if ($this->is_direct_attack_valid($attackers, $dval)) {
--- a/src/engine/BMUtilityHitTable.php
+++ b/src/engine/BMUtilityHitTable.php
@@ -43,66 +43,120 @@ class BMUtilityHitTable {
     private $hits = array();
 
     /**
+     * $die_ids is an array of possible unique ids to use in constructing hits
+     *
+     * @var array
+     */
+    private $die_ids = array();
+
+    /**
+     * $hitVals is the array of target hit values used to construct this hit table
+     *
+     * @var array
+     */
+    private $hitVals = array();
+
+    /**
      * Constructor
      *
      * @param array $dice
      */
-    public function __construct($dice, $maxHitValue = PHP_INT_MAX) {
-        // For building hash keys, every die needs a unique
-        // identifier, no matter how many there are, but if there are
-        // more than 36 dice, something is very, very wrong.
-        $ids = str_split("ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789");
-        for ($i = 0; $i < count($dice); $i++) {
-            $die = $dice[$i];
-            $die_id = $ids[$i];
+    public function __construct($dice, $hitVals = NULL) {
+        $this->dice = $dice;
+        $this->hitVals = $hitVals;
+        $this->die_ids = str_split("ABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789");
+        $this->add_dice_to_hit_table($dice, array());
 
-            $this->dice[] = $die;
+        foreach ($this->dice as $dieIdx => $die) {
+            $die->run_hooks('hit_table', array('hits' => &$this->hits,
+                                               'dieLetter' => $this->die_ids[$dieIdx]));
+        }
+    }
 
-            foreach (array_keys($this->hits) as $target) {
-                foreach (array_keys($this->hits[$target]) as $key) {
-                    // We've already been used in this combo
-                    if (FALSE !== strpos($key, $die_id)) {
-                        continue;
-                    }
-
-                    foreach ($die->attack_values("Skill") as $val) {
-                        $newcombo = $this->hits[$target][$key];
-                        $newcombo[] = $die;
-                        // the new key will always be sorted, since we
-                        // process the dice in order
-                        $newkey = $key.$die_id;
-                        $newtarget = $target + $val;
-                        if ($newtarget > $maxHitValue) {
-                            continue;
-                        }
-                        if (array_key_exists($newtarget, $this->hits)) {
-                            // If the same die combo makes a number
-                            // two ways, we just overwrite the old
-                            // entry.
-                            $this->hits[$newtarget][$newkey] = $newcombo;
-                        } else {
-                            $this->hits[$newtarget] = array($newkey => $newcombo);
-                        }
-                    }
-                }
+    /**
+     * Recursively add attacking dice to the hit table
+     *
+     * Every time this function is called, it moves one die from $diceRemaining (the set of to-be-processed dice)
+     * to $prefixDieHits (the set of hits selected from dice).  When processing the die, the function looks at all
+     * possible values that die can contribute to an attack, either as an attacker or as a helper, and recursively
+     * calls itself once for each of those choices.
+     *
+     * When $diceRemaining is empty and all dice have had values selected for $prefixDieHits, the function
+     * determines the total target value of the hit represented by this combination of dice.  It then checks to
+     * see if that total is in $this->hitVals, the set of all target (defender) die values which this hit table
+     * instantiation is trying to hit.  If the target value is one of the ones this hit table is looking for,
+     * we save the combo of attacking die values participating in the attack (discarding helpers; they can and will
+     * be rediscovered later in the process of the actual attack) in $this->hits.  If the target value isn't one
+     * this hit table is looking for, we discard the combo to avoid using a huge amount of memory to store data
+     * about irrelevant attacks.
+     *
+     * @param int $target
+     * @return bool
+     */
+    protected function add_dice_to_hit_table($diceRemaining, $prefixDieHits) {
+        // There are dice remaining to process; process the next one, and recursively call this function
+        if (count($diceRemaining) > 0) {
+            $prefixDieHitsCopy = array();
+            foreach ($prefixDieHits as $prefixDieHit) {
+                $prefixDieHitsCopy[] = $prefixDieHit;
             }
-
-            // Add the unique values the die may provide
-            foreach ($die->attack_values("Skill") as $val) {
-                if ($val > $maxHitValue) {
-                    continue;
-                }
-                if (array_key_exists($val, $this->hits)) {
-                    $this->hits[$val][$die_id] = array($die);
+            $firstDie = NULL;
+            $diceRemainingCopy = array();
+            foreach ($diceRemaining as $dieRemaining) {
+                if (is_null($firstDie)) {
+                    $firstDie = $dieRemaining;
                 } else {
-                    $this->hits[$val] = array($die_id => array($die));
+                    $diceRemainingCopy[] = $dieRemaining;
                 }
             }
+            $prefixLength = count($prefixDieHitsCopy);
+            $prefixDieHitsCopy[] = NULL;
+            $possibleHits = $firstDie->attack_values("Skill");
+            // Values which can be contributed if this die participates in an attack
+            foreach ($possibleHits as $firstDieHit) {
+                $prefixDieHitsCopy[$prefixLength] = array($firstDieHit, TRUE);
+                $this->add_dice_to_hit_table($diceRemainingCopy, $prefixDieHitsCopy);
+            }
+            // Values which can be contributed if this die does not participate in an attack
+            $possibleAssists = $firstDie->assist_values("Skill", array(), array());
+            $possibleAssists[] = 0;
+            foreach ($possibleAssists as $firstDieAssist) {
+                $prefixDieHitsCopy[$prefixLength] = array($firstDieAssist, FALSE);
+                $this->add_dice_to_hit_table($diceRemainingCopy, $prefixDieHitsCopy);
+            }
+            return;
         }
 
-        foreach ($dice as $dieIdx => $die) {
-            $die->run_hooks('hit_table', array('hits' => &$this->hits,
-                                               'dieLetter' => $ids[$dieIdx]));
+        // There are no dice remaining to process; add this combo to the hit table
+        assert(count($prefixDieHits) == count($this->dice));
+        $target = 0;
+        foreach ($prefixDieHits as $dieHit) {
+            $target += $dieHit[0];
+        }
+        if (isset($this->hitVals) && !(in_array($target, $this->hitVals))) {
+            // This combo's target sum isn't needed for the hit table
+            return;
+        }
+
+        $newcombo = array();
+        $newkeyParts = array();
+        for ($i = 0; $i < count($prefixDieHits); $i++) {
+            if ($prefixDieHits[$i][1]) {
+                $newcombo[] = $this->dice[$i];
+                $newkeyParts[] = $this->die_ids[$i];
+            }
+        }
+        if (count($newcombo) == 0) {
+            // This attack contains zero dice
+            return;
+        }
+
+        // Actually add the combo to the hit table, either as a new key or a new combo for an existing one
+        $newkey = implode(".", $newkeyParts);
+        if (array_key_exists($target, $this->hits)) {
+            $this->hits[$target][$newkey] = $newcombo;
+        } else {
+            $this->hits[$target] = array($newkey => $newcombo);
         }
     }
 
--- a/src/engine/BMUtilityHitTable.php
+++ b/src/engine/BMUtilityHitTable.php
@@ -138,10 +138,16 @@ class BMUtilityHitTable {
             return;
         }
 
+        // The logic up to this point collected both direct and indirect die contributions,
+        // for the purpose of finding exactly those combinations which could under some circumstances
+        // hit a target of interest.  However, we want to install into the hit table the
+        // target sum from the direct contributions only, because that's what find_attack() will expect
         $newcombo = array();
         $newkeyParts = array();
+        $directTarget = 0;
         for ($i = 0; $i < count($prefixDieHits); $i++) {
             if ($prefixDieHits[$i][1]) {
+                $directTarget += $prefixDieHits[$i][0];
                 $newcombo[] = $this->dice[$i];
                 $newkeyParts[] = $this->die_ids[$i];
             }
@@ -153,10 +159,10 @@ class BMUtilityHitTable {
 
         // Actually add the combo to the hit table, either as a new key or a new combo for an existing one
         $newkey = implode(".", $newkeyParts);
-        if (array_key_exists($target, $this->hits)) {
-            $this->hits[$target][$newkey] = $newcombo;
+        if (array_key_exists($directTarget, $this->hits)) {
+            $this->hits[$directTarget][$newkey] = $newcombo;
         } else {
-            $this->hits[$target] = array($newkey => $newcombo);
+            $this->hits[$directTarget] = array($newkey => $newcombo);
         }
     }
 
--- a/test/src/engine/BMAttackSkillTest.php
+++ b/test/src/engine/BMAttackSkillTest.php
@@ -142,8 +142,8 @@ class BMAttackSkillTest extends PHPUnit_Framework_TestCase {
         $this->assertTrue($sk->validate_attack($game, array($die1), $def));
 
         // test case where the amount of help is explicitly specified
-//        $this->assertFalse($sk->validate_attack($game, array($die1), $def, array('helpValue' => 3)));
-//        $this->assertTrue($sk->validate_attack($game, array($die1), $def, array('helpValue' => 1)));
+        $this->assertFalse($sk->validate_attack($game, array($die1), $def, array('helpValue' => 3)));
+        $this->assertTrue($sk->validate_attack($game, array($die1), $def, array('helpValue' => 1)));
 
         $target->value = 2;
         $this->assertTrue($sk->validate_attack($game, array($die1), $def));
--- a/test/src/engine/BMAttackSkillTest.php
+++ b/test/src/engine/BMAttackSkillTest.php
@@ -142,8 +142,8 @@ class BMAttackSkillTest extends PHPUnit_Framework_TestCase {
         $this->assertTrue($sk->validate_attack($game, array($die1), $def));
 
         // test case where the amount of help is explicitly specified
-        $this->assertFalse($sk->validate_attack($game, array($die1), $def, array('helpValue' => 3)));
-        $this->assertTrue($sk->validate_attack($game, array($die1), $def, array('helpValue' => 1)));
+//        $this->assertFalse($sk->validate_attack($game, array($die1), $def, array('helpValue' => 3)));
+//        $this->assertTrue($sk->validate_attack($game, array($die1), $def, array('helpValue' => 1)));
 
         $target->value = 2;
         $this->assertTrue($sk->validate_attack($game, array($die1), $def));


cgolubi1:
Shadowshade, jl8e, what do y'all think of this?  Is there some algorithmic reason this won't work?

I don't *think* it should have appreciably slower runtime than the previous version; calling a function recursively shouldn't be a lot worse than doing the same effectively-recursive iteration over `$this->hits` for each die.  And i'm pretty sure it uses much, much less memory.

And as far as i can tell, it works; obviously i'll need some replay testing, but calling `assist_values()` directly seems to DTRT for both fire dice and the fake AVTesting skill used by the skill attack tests.  It did fail on the test that sends `helpValues` directly --- is that something *any* live code ever does, or is it just something that was inserted to aid some form of unit testing?  It seems to me that real skills should use hooks to define assists, precisely so that all code paths can deal with them identically.  So if `helpValues` isn't used by any real skills, i think we should get rid of it and expect to be able to define future skills cleanly.  But tell me if i have this one by the wrong end.

blackshadowshade:
I remember the last time I tried to alter this code, the failures we saw were to do with negative konstant contributions, so we *really* need to have a responder test for that case.

As for the contents of the changes here, I'll have to take a closer look.

blackshadowshade:
I think running replay testing on this change is probably the first step. That picked up the issues with my previous changes, and we'd need it anyway.

blackshadowshade:
Good to see that the failing test was signalling something useful.  :)

cgolubi1:
> Good to see that the failing test was signalling something useful. :)

Yeah, definitely!  I'll probably stick a comment on it that explains a little more of what it does, though Future Me may just remember this experience.

blackshadowshade:
Super Germ would also be another candidate for checking speed.

cgolubi1:
Oh, i need more coffee.  I thought you were talking about speed attacks.  Yeah, sure, i'll snag wall-clock times on some game executions.  I actually wonder if that's something i could get RandomAI to complain about in general.

jl8e:
Super germ's not actually that great, as it only has 8! possible attacks. Stinger buttons are the way to go, because each possible stinger value has to be evaluated. In this code, Fire  will have the same effect.

Since hit tables were an attempt to trade memory for time, any of the buttons that make huge hit tables will also really stress the timing.

cgolubi1:
Okay.  This isn't going to be exceptionally scientific, but my plan is to have RandomAI record 100 games each of which chooses both buttons randomly from among (skapheles, Aysez, Super Germ, Notorious, The Jabberwock, Totoro, bluebec), and then i'll replay those games on both the control replay site and the site with this branch, and see how long the replays relatively take to run all 100 games.

I do want to throw out there that my expectation is that this new site will be faster.  That's gathered (again, extremely unscientifically) from the existing replay_loop logs, which have timings in them (at least at the level of "how long does it take between starting to replay 100 games and starting the next thing?"), from which i got the following head-to-head comparisons, where `CTL` is the number of seconds it took replay-cgolubi1-master to play the file, `BNC` is the number of seconds it took `replay-cgolubi1-hittable_failure_modes` to play the file, and `TARBALL` is the filename containing the games:

```
 CTL BNC GAME TARBALL
 --- --- -------------------
 521 511 fc7ef9daaee66c55a03c66bab3b77c44e8e26f5a.games.20211007.012419.tar.bz2
 537 503 fc7ef9daaee66c55a03c66bab3b77c44e8e26f5a.games.20211007.015009.tar.bz2
 536 518 fc7ef9daaee66c55a03c66bab3b77c44e8e26f5a.games.20211007.021606.tar.bz2
 508 476 fc7ef9daaee66c55a03c66bab3b77c44e8e26f5a.games.20211007.024116.tar.bz2
 534 496 fc7ef9daaee66c55a03c66bab3b77c44e8e26f5a.games.20211007.030645.tar.bz2
 537 512 fc7ef9daaee66c55a03c66bab3b77c44e8e26f5a.games.20211007.033237.tar.bz2
 539 507 fc7ef9daaee66c55a03c66bab3b77c44e8e26f5a.games.20211007.035838.tar.bz2
 529 502 fc7ef9daaee66c55a03c66bab3b77c44e8e26f5a.games.20211007.042420.tar.bz2
 509 492 fc7ef9daaee66c55a03c66bab3b77c44e8e26f5a.games.20211007.044954.tar.bz2
 534 509 fc7ef9daaee66c55a03c66bab3b77c44e8e26f5a.games.20211007.051522.tar.bz2
```

Like i said, this is not exactly a rigorous attempt to control variables between sites, but i think the fact that these logs think the new site is a little bit faster, provide some evidence that it isn't a whole lot slower.

And that doesn't surprise me.  I am not a software optimization expert by any stretch, but i don't think there's any significant overhead to starting a new function in modern programming languages.  And i don't think my solution is any more recursive than the old one --- it's just that the recursion is visible because it looks like a recursion, rather than looking like a linear iteration over dice and a second iteration over hits, which obfuscated the fact that hits was getting larger and larger on each iteration.  And all else equal, it ought to save us at least a little bit of time to be carrying around a much smaller memory object in the form of the hits table itself.

Anyway.  I'll replay these games with the above list of stingery and/or large buttons, and let y'all know what i find.  And then i'll go back to replaying random games and looking for regressions (up to 5800 games now, so i'm cautiously optimistic that there's not a major logic snafu here, though as noted my target is 4x that many).

blackshadowshade:
Sounds like a good plan to me.

blackshadowshade:
On the side, I think it's great that this central part of the code (hit tables) will end up having contributions from each of the three primary backend contributors to this project. In this sort of case, I think many minds do help to solve a problem better in an iterative manner.

Now, to work out how to deal with konstant and fire.  XD

blackshadowshade:
Looks good to me.

cgolubi1:
So, i got a bunch of successful replays, then a failure, as shown here, with the error thrown being:
```
<b>Fatal error</b>:  Maximum execution time of 30 seconds exceeded in <b>/var/www/engine/BMUtilityHitTable.php</b> on line <b>100</b><br />
```

I'm curious how this particular game would fare on the master branch, and will try it out as soon as i can.  But i also lost track of this testing, so i'm going to rebuild the replay site and collect some more games.

Shadowshade: RandomAI never tries invalid attacks, but i think the timeouts would happen, as they do in this case, when the site is calculating possible valid attacks.  I don't think hit tables are used when checking the validity of the selected attack, so i don't think that code path should be changed.  No?

ferrer_vs_custombm

cgolubi1:
Oh, i guess i should say: RandomAI never tries invalid attacks, but it *does* randomly pick an attack type the server said should be valid, and look for an attack of that type.  So that means that if the backend were reporting "you have a valid skill attack" when that was not true, eventually RandomAI would blow up because it was told there should be a skill attack and it couldn't find one.

cgolubi1:
Huh, as far as i can tell, this game works fine on the default branch site.  So that's interesting, and definitely something i'll have to understand before moving forward.

blackshadowshade:
I think it may generally be a useful idea to be attempting invalid attacks as part of the testing, especially when we're mucking around with things that underlie skill attacks.

Not necessarily a blocker for this pull request, just a passing thought that I had.

cgolubi1:
This game yielded `PHP Fatal error:  Maximum execution time of 30 seconds exceeded in
/var/www/engine/BMUtilityHitTable.php on line 133` on the first load after setting swing dice to W=10:

```
|      20 |         3 |     11000 | F(W) F(W) F(W) F(W) F(W) F(W) F(W) F(W) F(W) F(W) F(W) | F(W) F(W) F(W) F(W) F(W) F(W) F(W) F(W) F(W) F(W) F(W) |        0 |                  1 |                  1 |            0 |             0 |              0 |        0 |                0 | 2022-01-09 14:56:52 |                  0 |                0 |                   1 |               0 |
|      20 |         5 |       169 | (4) (10) (12) (20) r(4) r(12) r(20) r(20)              | NULL                                                   |        1 |                  0 |                  0 |            0 |             0 |              0 |        0 |                0 | 2022-01-09 14:56:52 |                  0 |                0 |                   1 |               0 |
```

Again, i need to check if this game is playable on master.

blackshadowshade:
That's a remarkably degenerate button recipe, 11 fire dice is just asking for trouble.

blackshadowshade:
However, if we wrote the skill attack detection correctly, similar to my attempt previously, we would short-circuit further calculation when we hit a valid attack, which would simplify things considerably.

cgolubi1:
Yeah, maybe that game wouldn't work on the default branch either --- i need a better rig to do head-to-head comparisons here so that it's easy to find out whether this branch is introducing problems, or just failing to solve them.  I'm trying to understand some of the deadlock and OOM issues that *aren't* caused by hittable, so i can get back to testing this.  Things are just a little noisy right now in terms of inexplicable failures, so i wasn't having a lot of luck assessing this PR for correctness.

cgolubi1:
I've basically forgotten the status of my hypothesis about this PR.  I *think* where we are is:
* There are three things we're interested in proving about this code change:
    * It's not a logical regression
    * Its performance is at least as good as that of the old code in all places
    * It successfully prevents irilyth's fanatic from causing the site to hang
* Status of each of those:
    * I claimed it's not a logical regression on the strength of 2600 games, which is a reasonable number of games, but for something core, i'd want to see many more games than that.  Since it's trivially easy to stand up a replay site and forget about it for a week, which should yield many more than 2600 games, that seems like the obvious first thing to do
    * I was trying to demonstrate that it had better performance using some somewhat contrived examples which appeared to show that, but i have to reread more carefully to determine how i feel about that experiment.  I definitely don't have any kind of performance test rig, so any attempt to do this kind of head-to-head performance comparison is going to involve a lot of handwaving and extra code.  I should dig back into that.
    * I found at least one example of a game which crashed with a timeout on the new code, which was discouraging --- the last comment i left was that i need to determine whether this game (involving the button `F(W) F(W) F(W) F(W) F(W) F(W) F(W) F(W) F(W) F(W) F(W)`) is playable on master.  I imagine it isn't, but my next step (which i never took) was finding that out.  I need to keep in mind that if the codebase can handle irilyth's fanatic, it doesn't necessarily need to be able to handle any possible button ever, to meet our needs
* So i think the next step is to kick off item 1, which is easy, and then sit down for 20 minutes and propose the actual test plan to take this across the line.

cgolubi1:
I got a mysql OOM-kill on the replay site --- the game was `IIConfused` mirror battle, and it went to 200 rounds and then was cancelled.  Nothing else to report, restarting the test.

cgolubi1:
The replay site has now played 11100 new and replayed games, with the only issue being the IIConfused mirror battle OOM mentioned above.  No reason not to let it keep running and get more, but that's a lot of games.  I think if something was wrong with the logic, we would have seen it by now.

cgolubi1:
Anyway, i've switched the site to playing irilyth vs random novel games --- let's see under what circumstances exactly the thing we want to unblock causes crashes, and let that lead us to what data we need to gather next to assess items 2 and 3 from my list above.

cgolubi1:
Okay, so i'm trying my head-to-head comparison idea again, but this time i'm comparing normal replay logs taken from the buttonmen-master replay loop, vs normal replay logs recorded on this site.  I scripted the comparison, and it looks like i currently have 110 data points.

On average, it took master 519.3 seconds to replay a file (there's a tar command in there, i think a database reload, and then playing 100 games).  On average, it took the branch code 511.9 seconds.  So that's an improvement of 7.4 seconds (or about 1.4%).

But the distribution is weird --- for 107 of the 110 logs, the two sites are within 50 seconds of each other, and for that subset the average improvement is 11.4 seconds/log (so, a bit better).  There are 3 logs where the difference is more than 50 seconds, and for all of those, the branch did worse:

```
7e1f44265521d038513a19882b798f6b31c9e03e.games.20230128.060000.tar.bz2 538 599 61
7e1f44265521d038513a19882b798f6b31c9e03e.games.20230128.071718.tar.bz2 523 789 266
7e1f44265521d038513a19882b798f6b31c9e03e.games.20230129.061409.tar.bz2 527 603 76
```

So at a glance, that looks a little alarming, especially the one where the new code took a massive 266 seconds longer than the old code.

I compared it to the `266_wildcard` replay loop.  That loop also has a couple of cases in which the new branch was over 50 seconds slower than the reference branch, including one several-hour whopper corresponding to the OOM from this morning.  So definitely there are a lot of factors impacting relative speed, and a lot of them are very random and probably have more to do with vagaries of EC2 instances, than with the code.

I would also note that `266_wildcard` outperforms master by more than the hittable change does --- and obviously Wildcard isn't a performance improvement.  I would also note that `266_wildcard` only has 27 data points, vs the hittable change's 110.  Sooo, maybe the small number of data points is relevant, or maybe this is a bogus way to measure the performance of a site change.

So basically in terms of the question of "[This branch's] performance is at least as good as that of the old code in all places", i've got... well, it doesn't seem obviously worse.

cgolubi1:
The other thing i'm doing is i stood up a fresh replay site of the master branch, and i'm running irilyth vs random games there as well --- if it doesn't crash master either, it's not a good test of whether anything has improved.

cgolubi1:
Let me summarize my thinking here, because i feel like nothing i've done related to performance testing of BM to this point has been especially either scientific or helpful.  So, i'm going to run down a few dimensions:

Performance controls:

I haven't looked into the MySQL level at all, but at the PHP level, there are a few controls that we can modify, all of which seem potentially useful for replay-testing and/or long-term site use:
* `max_execution_time`: This is a PHP per-script limit on how much wall-clock time a PHP process can spend in its logic (time spent in system calls doesn't count).  It defaults to 30 seconds, and is associated with the `<b>Fatal error</b>:  Maximum execution time ...` error message.  Hypothetically, we could reduce it for use in testing, to try to catch regressions earlier/in a more controlled way.
* `memory_limit`: This is a PHP per-script limit on how much memory a PHP process can allocate.  It defaults to 128M, and is associated with the `<b>Fatal error</b>: "Allowed memory size ...` error message.  Hypothetically, we could reduce it for use in testing, to try to catch regressions earlier/in a more controlled way.
* `slowlog`: This is PHP/FPM functionality to log a stacktrace when a query takes more than a specified number of seconds.  In tests, setting this is slightly at odds with using `max_execution_time` (because neither can be set lower than 1 second, and in general we think if anything in our system is taking 1 second in a single-threaded test, that's already bad news), but it could in principle help catch excessive time waiting for MySQL queries, etc.  It could also be useful to set in production, to get a report of slowness impacting us in the field.

Limits of replay testing:

I want to call this out really explicitly, because i think it contradicts perceptions others (including me 24 hours ago) may have of what we can and can't do with replay testing:

As of now, i *cannot* cause a single replay test to fail with a backend performance error for *any* case involving named buttons or backend-legal CustomBMs.  I *cannot* make the backend fail with irilyth, i *cannot* make the backend fail with IIconfused, i *cannot* make the backend fail with a CustomBM full of big stinger dice.  I think that's surprising --- i think people believe i can do that (and i've been able to do it in the past, before we put guardrails around CustomBM, solved some of the zero-sided die problems, etc).  And to be clear, i'm not talking about this branch --- i'm taking about the base, buttonmen-master, codebase.  I can't make a single replay test cause it to fail with a backend performance error.

Here are some things i can do, which are much more limited than the above:
* I can cause a "Maximum execution time" PHP error (and fast!) using a mirror battle of the recipe `ng(99) ng(99) ng(99) ng(99) ng(99) ng(99) ng(99) ng(99)` - to do that, i have to relax the limit that a recipe can only have up to 200 sides, because that recipe has 800 sides (fine, 792).
* Running a 200-round `IIconfused` mirror battle, i can cause `test_log_games` (the python script that runs RandomAI) to be OOM-killed.  Presumably the reason is that when the game ends, RandomAI queries that final action log with all the rounds in it, and processes it in memory in an inefficient way.
* Running a test of irilyth vs the pathological CustomBM in the first bullet, i can cause a hang --- but, again, it's not the backend hanging, it's `test_log_games`, because RandomAI has to do all the math to find a valid skill attack too, and it doesn't even have a hit table.  I can come in via the UI while `test_log_games` is chewing up CPU, and take an attack --- the backend is just fine.

There are still things i'm trying (e.g. i currently have a game running with Giant vs CustomBM `F(W) F(W) F(W) F(W) F(W) F(W) F(W) F(W) F(W) F(W) F(W)`, which is pathological but legal, and it's progressing but slowly --- but again, the CPU crunching is happening in `test_log_games`, not in PHP or MySQL.  So i don't think that's going to turn out to include a backend problem

A corellary of the fact that i cannot cause a single replay test to fail with a backend performance error for *any* case involving named buttons or backend-legal CustomBMs, is that i don't believe we have real data about either of these two questions:
* Will the irilyth fanatic cause the site to hang when played against any named buttons or legal custom buttons?
* Does the hittable refactor in this PR perform better than the base code under any real-life or contrived conditions?

Hang on, if you can't crash replay testing, why do you talk about replay testing crashes so much?

Yeah, so, the table i'm collecting in #2801 is a table of cases in which replay testing ran for a long time, and after awhile it crashed.  And in general, if i take an individual game that i identified (correctly or not) as the trigger in one of those events, and replay that game in isolation, it doesn't crash.  So definitely it's not consistent for any set of buttons i've identified, see above.

Given that, it seems overwhelmingly likely to me that MySQL OOM is caused, not by an individual pathological game, but by buildup of state which isn't cleaned up.  And it seems pretty likely to me that the same is true for Apache OOM / "Max execution time" errors seen in the wild.  That is, single games aren't causing these things.  Poor garbage collection/poor separation of state for multiple games/poor handling of resource limitations hit within the AWS virtual resource space, is what's causing them.

Well, this is a fine pickle.  Should we just give up?

Maybe!

No, okay, seriously.  I think the following set of actions are worth taking:
1. Run slowlog with a 1-second limit on replay sites, and if it seems to give interesting results and to be harmless, roll it out to staging and production too, with reporting on the log contents --- the promise of getting outside info about what PHP code causes the database to bog down, sounds *highly* valuable.
2. I've been testing against buttonmen-master using a modified PHP that has `max_execution_time` of 1 second.  Try running that more --- if we really think nothing in these replay games should ever take more than a second of actual PHP execution time, in isolation, let's test that theory by just continuing to run that way.
3. Experiment with lowering `memory_limit` for replay PHP as well (i haven't played with this at all yet).  If we can get reproducible failures with irilyth's fanatic at a higher level than we get reproducible failures with other named buttons, that will give us real data that will help us assess the risk of enabling the button (and assess whether the hittable mitigation helps).  I don't know that any of that will actually work, but it seems promising and i haven't tried it.

I definitely don't think we should be too eager to merge this PR.  We don't have real evidence that it helps with an actual problem, and i was very pleased with the solution when i wrote it, but that was awhile ago, and at the end of the day i'm not that invested in having us adopt it *unless* it helps with an actual problem.

I also don't think we should stall forever on irilyth's fanatic.  Maybe we give ourselves H1 2023 (until the end of June) to either convince ourselves that it's safe or get real data that it's dangerous, and if we haven't learned anything by then, just merge it?  Sure, we risk that it does something disastrous in prod and we have to disable it and mark some games broken, but that's life.  I think at some point we have to either bring the data or move forward with the button, and definitely my personal memory of the extent to which we had brought the data, was not really backed up by the evidence i could find.

cgolubi1:
I've started a hand-modified version of the buttonmen-master replay site which has all of `max_execution_time=1s`, `memory_limit=4M` (2M blew up calling `listButtons`, so far no problems with 4M), and `slowlog` enabled.  I'll let that run for a day without messing with it, and see how it goes.

blackshadowshade:
That's definitely progress! Great work figuring out what we should be testing!

cgolubi1:
Well, this is exciting.  So i got a baseline of 5000 games on the hittable refactor branch, with max_execution_time=1 and memory_limit=4M.  No problems.

Then i started playing irilyth games, and, remember how we were like "this branch probably trades off memory size for execution time, but probably it's fine."  Well: it does.  Numbers so far:
* I couldn't get a single game to complete with max_execution_time=1, including some games which failed out on the first move, never mind the first round.  5/5 games hit "Maximum execution time of 1 second exceeded"
* With max_execution_time=2, 5/7 games failed, most but not all of them after adding at least one reserve die.
* With max_execution_time=4, 5/12 games failed, all of them after adding 1-2 reserve dice
* With max_execution_time=8, 5/64 games failed, all of them after adding 1 or usually 2 reserve dice
* I'm still running with max_execution_time=16.  So far so good (but only 23 games in).

So, even if this is the only thing up with this PR, i think we have to ask some real questions about whether this is a good tradeoff. 

Why it's probably not a good tradeoff:
* This PR increases complexity, so if it's a wash, we shouldn't do it.
* It seems like irilyth is more expensive relative to other buttons in time than in memory --- in memory, we had to allow 4x the amount other buttons needed; in time, we have to allow 16x the amount other buttons need.

Why it might be a good tradeoff:
* Maybe letting an expensive call tax the service on CPU/execution time, is less harmful to other players than letting an expensive call tax the service on memory.  With big memory usage, we risk the OOM-killer coming in and disrupting games with other buttons.  If using irilyth's fanatic means you get a slow response from the server, but other games are fine, and any timeouts only affect the game containing irilyth, maybe that's a better overall experience in terms of what happens when we enable this button.

It's also worth calling out: these replay tests run on a t3.micro EC2 instance, which has 1.0 GB of memory and 2 vCPUs (but it's a burstable model, so you're only allowed to use them 10% of the time, which might well be relevant).  Production is a t3.small, which has 2.0 GB of memory (and a similar CPU situation, though you can use the vCPUs 20% of the time).  So it's quite possible that memory impacts would be less impactful on prod than on replay testing.

Anyway, i'll sleep on this and see what i think of to try next.

cgolubi1:
Okay, so, i have upgraded (downgraded?) my opinion on this PR as written to "no way."

What happened is that with a max execution time of 16 seconds, it played 166 games and then failed with a max execution time error.  And then i restarted it, and it failed again on game 3.  And i don't think that's a coincidence --- i think the burstable CPU on these instances means that if we're driving the CPU hard, we're more likely to ask for CPU at a time when we're out of credits, and have things bog down until we get a timeout.

Whereas the memory we get is the memory we get --- it doesn't have a bursting model.

Given that that's the hardware we're running the site on, trading relatively predictable memory for relatively unpredictable CPU is just going to be a maintenance nightmare.  We definitely shouldn't do it.

I don't strongly understand why the rewrite is so much more CPU-intensive than the old way.  So maybe it'd be worth tracing some code and convincing myself that there's not a dumb bug, or maybe i should just cut bait.

If we *can't* use this PR at all, what's our plan for getting irilyth's fanatic playable?  Do we have one?

One idea that i don't know how to test the impact of, is to wonder if we should *reduce* the `memory_limit` parameter in production.  The (very speculative) hypothesis is:
* The vast majority of API calls are perfectly happy with 4M
* Actually allocating 128M (the current max) to a call destabilizes the overall server, and causes impact to everything using the server

So if we reduced the max allowed to e.g. 32M, then very slow calls might time out, and players might have to retry those games or we might eventually have to intervene, but by erroring out those greedy calls, we prevent impact on other games and server use.

irilyth:
> Given that that's the hardware we're running the site on

Is this a given? it's not hard to resize instances, and these days I think you can even move between instance classes; https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-resize.html has some docs. It would probably be possible to dry-run it on a dev site and/or the staging site.

If that's not appealing for other reasons, that's fine, just wanted to check. We gave up on T-class instances at work a while ago because we didn't want to deal with the unpredictability, and we recently found some old ones that we hadn't replaced yet, so the idea was close to mind.

cgolubi1:
> Is this a given?

My short answer is: yes, because of cost.

My long answer starts with: i'll briefly elaborate for the curious, but reminder that this is a digression --- we only care about the increase in dependence on CPU stability *if* we adopt this PR, and i've put forward several reasons not to adopt this PR, and only one weak reason to (and no one else has put in any arguments about whether or not to adopt this PR, after my testing this week).  So i think we should avoid getting sidetracked onto talking about hardware migration *unless* we actually decide we want to adopt this PR.  I'm much more interested in whether anyone still thinks this PR is worth putting time into, in light of my test results from this week.

Anyway, for the curious:
* T-class instances are a lot cheaper per instance-hour than other classes.
* Right now we count on being able to spin up replay instances without it being very expensive, to avoid introducing logic regressions into the site.  We could fix that by completely rewriting how we do regression testing, at the cost of a lot of developer time not spent on adding new features or getting to beta.
* We could move prod to a more expensive instance, not a huge deal, maybe even staging, but to some extent, that doesn't really matter --- if we are doing our logic break testing on instances that are known to be much more brittle than prod and behave differently than prod in terms of brittleness, then that's going to get in the way of understanding whether new code is safe to bring on prod.

Keep in mind the actual context of this PR we're considering: this PR is trying to solve a problem we've historically had with *memory* consumption, by turning memory consumption into CPU consumption.  If we choose not to adopt this PR, we won't turn memory consumption into CPU consumption.

So the fact is, this PR would increase our dependency on the brittle part of the instances we use now.  We could solve that problem with money i don't really want to spend.  We could also solve it with tens of hours of work rearchitecting how we do testing to avoid the problem that this PR introduces.  *But we could also solve it by not adopting this PR.*

The question of instance type was introduce as a side comment about one of several reasons this PR probably isn't a good tradeoff, so i don't think we need to worry about the instance type thing until we have an actual argument that adopting this PR would be desirable.  I've made several arguments for why it wouldn't, and only one for why it would, namely: maybe it's less disruptive to other players to CPU-kill PHP than to memory-kill it.  And i have a proposed mitigation for that one without this PR (lower the per-call memory limit on the site), which i think is testable, and i think we could establish through loaded dev testing whether it would unblock the irilyth fanatic.

So afaict, there's no reason to consider adopting this PR.  It won't get us closer to being able to merge irilyth.

jl8e:
> I don't strongly understand why the rewrite is so much more CPU-intensive than the old way. So maybe it'd be worth tracing some code and convincing myself that there's not a dumb bug, or maybe i should just cut bait.

It's probably not a bug. The fundamental problem is that there's potentially n! comparisons to see if a skill attack can hit. The hit tables trade this for a constant memory cost, but there's still an n! in there.

> If we _can't_ use this PR at all, what's our plan for getting irilyth's fanatic playable? Do we have one?

Yes, probably. I worked out a new heuristic method a while back, over in #2692 , but it needs a lot of implementation.

I can't prove it's faster, but it ought to be a lot faster, and it has no significant memory overhead.

cgolubi1:
Meanwhile:
* I did get a failure testing irilyth on the base site, in round 2/2/0 of irilyth vs ABCGi.
* I realised that my testing wasn't applying `memory_limit` to the cli, which i thought i didn't care about, but phpunit is of course a CLI.  So that means the replays weren't using a memory limit
* I added it, and got a consistent failure of the specific `irilyth` vs `ABCGi` game i'd recorded at 16M, then raised the limit to 32M and it succeeded.
* There are lots of ways that test was insufficient that i don't feel like going into because i don't want to fix them all on the spot, but for my testing purposes, it would be terrific if 1. specific games that caused hangs consistently caused hangs (which really should be the case for memory-related hangs --- there shouldn't be a lot of variability there), and 2. the same games caused hangs in initial RandomAI play as in phpunit replay.  So i'm pleased by the mild corroboration that that might be the case, because it'll make a lot of things easier going forward.

cgolubi1:
> It's probably not a bug. The fundamental problem is that there's potentially n! comparisons to see if a skill attack can hit. The hit tables trade this for a constant memory cost, but there's still an n! in there.

Thanks.

> Yes, probably. I worked out a new heuristic method a while back, over in https://github.com/buttonmen-dev/buttonmen/issues/2692 , but it needs a lot of implementation.

Yeah, all i can say about that at a glance is that hopefully the work i'm doing this week gets us much closer to a way to compare these algorithms head-to-head in real conditions.  (And of course replay-testing is already an effective way to find out whether we've changed the logical behavior of the code, but if tests fail unpredictably for performance reasons, it's hard to get to those logic checks.)

I would say if y'all want to pull the thread of whether we could be close enough to a hit table replacement to try that approach, i can pull the thread of whether putting limits on PHP could let us enable irilyth in a way that won't hang the site (and what that outcome would look like for games involving irilyth, so we can decide if we like that tradeoff, of course).

irilyth:
> My short answer is: yes, because of cost.

That makes sense; and thanks for all the details, some of which I'd misunderstood while not reading closely enough.

cgolubi1:
Okay, another irilyth failure with 16M, down 0/2/0 vs `SC (The Deuce)`, with the `+(20)` selected and two reserve dice.  So, let me bump up the memory again to 32M, for both the initial play and phpunit, and see if i can get some games to succeed.

cgolubi1:
Okay, so i've successfully played 6000 novel games with the irilyth fanatic with the limit at 32M.  All of the phpunit tests have failed, presumably because the memory limit applies to the entire phpunit process, not to any individual unit test.

cgolubi1:
Alright, i updated my replay site creation scripts to put the execution time and memory limits in place for initial game play (but not for cli/phpunit, since i need to do a bit more work to get that done).

With that, i think it's time to stop using this PR as scratch space for some combination of issues related to performance regression testing and enabling irilyth's fanatic, and create some new issues, then close this out.  I'm going to tag the new issues beta, to keep the prioritization on this, but we can untag them if we disagree.  The new issues are:
* Apply memory and execution time limits to tests run by phpunit
* Enable irilyth's fanatic (with the proposed implementation of limiting the memory used by each PHP process on the live site to 32M)

cgolubi1:
Okay, done!  Closing this out.

cgolubi1:
Oh, sorry, one more: i want to enable the `slowlog` on all the sites, so we get notifications of slow MySQL queries.  I didn't see any slowlog entries yet on the sites i was testing on, so i'm going to need to do some dedicated experimenting with it and make sure it works.

PR #2784, opened 2021-11-20
ubuntu upgrade to 20.04

danlangford:
I wanted to explore the idea of an ubuntu upgrade. I have no idea how this might impact various systems but I wanted to have a discussion and at least see how the tests would pass. 

16.04 setup had some inconsistencies. the AMI is 64bit while the vbox setup is 32bit. 16.04 is an LTS but its getting fairly old and falling off of some of the ubuntu radar. more and more I need to drill down to older releases or legacy pages to find data. 

this attempts to upgrade to the latest LTS 20.04. There is another LTS 6 months away 22.04 

p.s. should my random PRs be targeted to `dev` instead of `master`? what is the proper PR etiquette here?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
   build:
     docker:
-      - image: ubuntu:16.04
+      - image: ubuntu:20.04
     steps:
       - checkout
       - run:
--- a/deploy/vagrant/Vagrantfile
+++ b/deploy/vagrant/Vagrantfile
@@ -10,11 +10,11 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   # please see the online documentation at vagrantup.com.
 
   # Every Vagrant virtual environment requires a box to build off of.
-  config.vm.box = "ubuntu/xenial32"
+  config.vm.box = "ubuntu/focal64"
 
   # The url from where the 'config.vm.box' box will be fetched if it
   # doesn't already exist on the user's system.
-  #config.vm.box_url = "http://files.vagrantup.com/xenial32.box"
+  #config.vm.box_url = "http://files.vagrantup.com/focal64.box"
 
   # Create a forwarded port mapping which allows access to a specific port
   # within the machine from a port on the host machine. In the example below,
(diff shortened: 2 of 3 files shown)


blackshadowshade:
The last time we upgraded our Ubuntu version, we did knowing that we wanted to upgrade the associated PHP version, which then meant that we needed to make the large number of changes to move from PHP 5.2 to 7.0.

I'm guessing that this upgrade will mean moving up from PHP 7.1 to 8.0, which may not be trivial.

blackshadowshade:
Also, all pull requests are targeted at master, but we don't merge them unless they are trivial or they have been tested on at least one dev site.

blackshadowshade:
I've checked, and by default, the release comes with PHP 7.4 or PHP 8.0.

danlangford:
> I've checked, and by default, the release comes with PHP 7.4 or PHP 8.0.

im not familiar enough with the PHP ecosystem to know if 7.0 to 7.4 is considered a big and disruptive jump. I mostly come from java where any release under the same MAJOR version number is absolutely trivial and always 100% backwards compatible.

danlangford:
answering my own question here looks like PHP will make backwards incompatible changes in MINOR releases. 

> This new minor version brings with it a number of new features and a few incompatibilities that should be tested for before switching PHP versions in production environments.

https://www.php.net/manual/en/migration71.incompatible.php
https://www.php.net/manual/en/migration72.incompatible.php
https://www.php.net/manual/en/migration73.incompatible.php
https://www.php.net/manual/en/migration74.incompatible.php

blackshadowshade:
Yup, I've taken a look at those changes, and PHP 7.4 looks like it should work without changes to the codebase. Just checking out 8.0 now.

blackshadowshade:
Okay, definitely potential issues with the migration to 8.0.

- The ability to call non-static methods statically has been removed.
    - I don't think that we have any of these left, but if they're there, we should tidy them up anyway
- each() has been removed. foreach or ArrayIterator should be used instead.
    - I can't remember if we use each(), but it's possible.
- Inheritance errors due to incompatible method signatures (LSP violations) will now always generate a fatal error. Previously a warning was generated in some cases.
    - It's possible we have some of these in the daughter BMDie* and BMSkill* classes
- The precedence of the concatenation operator has changed relative to bitshifts and addition as well as subtraction.
    - I hope this doesn't affect us, but I'm not certain.

None of these are huge, certainly not as problematic as the shift from 5.2 to 7.0.

danlangford:
looks like a fairly significant puppet update

Issue #2786, opened 2021-12-29
push new features out to staging.buttonweavers.com


cgolubi1:
This is a tracking ticket for all updates to the buttonmen staging site.

PR #2768, opened 2021-09-29
Fix get_friendly_time_span regression

cgolubi1:
* Fixes #2767

PR #2759, opened 2021-09-11
Migrate BMInterface.php to BMDB

cgolubi1:
* Partially addresses #712

PR #2755, opened 2021-09-11
Add a cron job to test whether the database site_type matches the javascript site_type

cgolubi1:
There's not an issue for this, but i'm tired of getting bitten by it.

PR #2787, opened 2021-12-29
Push changes from master to staging on 2021-12-29

cgolubi1:
This update brings these approved pulls to the staging branch:

* #2750: Wrapped save_game in a transaction
* #2766: Run most responderTests on a production site_type
* #2755: Add a cron job to test whether the database site_type matches the javascript site_type
* #2759: Migrate BMInterface.php to BMDB
* #2752: Migrate BMInterfaceGame.php to BMDB
* #2768: Fix get_friendly_time_span regression
* #2770: save_game should throw an exception when it fails to save the game
* #2641: Fixed zero-sided option die min value

Database updates:
```
UPDATES=
SCRIPTS=
```

Once this pull is merged, i will load the changes on staging.buttonweavers.com as part of #2786.


cgolubi1:
Staging site is updated.  I'm still having a little bit of trouble accessing the staging site, which i think is a result of the IP address changing when AWS made me move to a different instance.  Working on that now

cgolubi1:
Ugh, except i'm an idiot --- i have an elastic IP for staging, so i didn't want to do any of those things, i instead wanted to associate that existing EIP to the new instance.  Working on this

cgolubi1:
Okay.  Cool, cool.  I think things are working, once DNS catches up now that i've unwedged it again.

So my new overall plan is to mostly-automate the process of deployments to staging and production, so that running a deployment is on the order of 5 minutes of work and zero decisions.  Then i'll deploy weekly, going to staging on Wednesday or Thursday, and to prod over the weekend.  That way we won't have changes hanging out for weeks or months after we've forgotten how far they've been pushed, and the bar will be lower for avoiding regressions because we can fix them quickly.  Maybe this is overambitious, but i gotta do *something*, and i think trying to mostly automate deployments is the lowest-hanging-fruit thing to do.

cgolubi1:
For staging testers: everything in this deployment is an invisible change, which is presumably part of why it's been hanging out for so long.  The save_game stuff is designed to avoid the kind of race conditions that led players to complain about a couple of recent game corruption things.  And the BMDB stuff is cleanup of how we interact with the database and should be invisible.  And we don't have zero-sided option dice in any extent buttons; they're for a future button.

cgolubi1:
Sigh, sorry for the bother; the previous deployment actually didn't take effect because of the rsync issue described in #2788.  *Now* there should be new code on the staging site.

PR #2788, opened 2021-12-29
Newer vagrant versions require additional AWS config

cgolubi1:
* This forces folder sync to use rsync, rather than attempting SMB sync

See https://github.com/mitchellh/vagrant-aws/issues/365 for what happened when i tried running with a newer vagrant without this configuration.

I have no reason to believe this won't work for anyone else, but IMO this doesn't need any other testing or validation, because i think i'm the only person who's actively using `setup_aws_vagrantfile`


cgolubi1:
Hmm, i thought this solved my problem, but i'm still missing something.  Marking it incomplete for now.

cgolubi1:
Okay.  This does work: there's just a second change needed --- in the older version of vagrant, if you did `vagrant provision`, it would rsync folders as the first step of that provision.  Now if you just run provision, it won't bring new config into the instance --- you need `vagrant rsync && vagrant provision`.

Anyway, this change is orthogonal to that, and will make my life a little easier if approved.

blackshadowshade:
If you're happy with this, I'm happy to let it go through.

PR #2790, opened 2021-12-30
Clean up how puppet variables are referenced in templates

cgolubi1:
This one isn't *so* critical, but it's a noop which cleans up an annoying warning that shows up when i run vagrant provision:

```
==> default: Running provisioner: puppet...
==> default: Running Puppet with init.pp...
==> default: Warning: Variable access via 'puppet_hostname' is deprecated. Use '@puppet_hostname' instead. template[/tmp/vagrant-puppet/modules-71904193657bd9e6560c6f15cd46d302/postfix/templates/main.cf.erb]:22
==> default:    (at /usr/lib/ruby/vendor_ruby/puppet/parser/templatewrapper.rb:77:in `method_missing')
==> default: Notice: Compiled catalog for ip-10-153-208-25.ec2.internal in environment production in 1.03 seconds
```

(The warning shows up in the circleci output too, so we should be able to use circleci to verify that we got all the relevant ones.  It only shows the warning once, for the first occurrence it parses, but it will show it if there are any occurrences in any templates.)

### DIFF
--- a/deploy/vagrant/modules/apache/templates/site_default.erb
+++ b/deploy/vagrant/modules/apache/templates/site_default.erb
@@ -9,7 +9,7 @@
     ## Set environment variables for buttonmen database access
 
     # Database 1 (primary)
-    SetEnv DB1_HOST <%= database_fqdn %>
+    SetEnv DB1_HOST <%= @database_fqdn %>
     SetEnv DB1_PORT 3306
     SetEnv DB1_NAME buttonmen
     SetEnv DB1_USER bmuser1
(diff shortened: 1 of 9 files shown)


cgolubi1:
Marking this as incomplete, because i'm still having more core puppet problems, and i don't want to worry about this one until next week.

cgolubi1:
Okay, this should be fine to review, but not high-priority.

blackshadowshade:
I can confirm that I see the warning in another CircleCI run (https://app.circleci.com/pipelines/github/cgolubi1/buttonmen/120/workflows/c2b97a42-55c3-4be7-8c2f-193136d4f05d/jobs/381) at the beginning of the "Run puppet" step.

and the warning is not present in the current CircleCI run (https://app.circleci.com/pipelines/github/cgolubi1/buttonmen/118/workflows/395307aa-a0cb-45b8-8dbc-e08a78e35967/jobs/378).

CircleCI passes, so I'm happy with this.

PR #2797, opened 2021-12-30
Install mod_pagespeed in lieu of custom timestamp-based versioning

### LINKED ISSUE #2288: JS caching results in blank pages
The new JS caching scheme doesn't work quite correctly with the "Go to your next pending game" link.  For the first several games i played after updating the site this morning, clicking that link gave me a blank page.  Shift-reloading fixed the problem each time.

The issue, per firebug, was that it was trying to load `Api.<old_datestamp>.js`.  Something in the next pending game link code path must cache the old value of that URL directly.

The behavior recovered after awhile, and now i'm no longer seeing it.  So i think the problematic script itself must also eventually time out of cache.

cgolubi1:
* Fixes #2288 

My virtualbox testing of this looked pretty good --- mod_pagespeed basically takes ownership of browser-side caching of pages, and by default serves the pages from its cache, which it refreshes every 5 minutes on a per-page basis.  So that should mean that when we do a deployment, the new pages come in over the course of the 5 minutes following the rsync, and then everyone has them, and nobody should be either getting old pages or seeing blank pages after that 5 minutes.

(In theory, you can force a cache refresh, but i couldn't get it to work as documented in a reasonable amount of futzing around, and IMO as long as the 5-minute refresh actually happens, and it seems substantiated, as in, there are per-file metadata blobs with expiration timestamps under `/var/cache/mod_pagespeed`, and i verified multiple times that if i change the site type in `Config.js`, the updated header color becomes visible to my browser exactly when that timestamp says it should be, and that time is within 5 minutes.  I think that's good enough for our purposes.)

dev site is up at https://makeshift.dev.buttonweavers.com/


cgolubi1:
Assuming the qunit failure isn't real and the retry succeeds, i'll stand up a dev site so y'all can check this out.

cgolubi1:
Failure is persistent.  Let me look into this locally.

cgolubi1:
Hah, i think the test failure is just because i changed the site_type in the committed Config file while i was futzing with this.  Verifying and fixing.

cgolubi1:
Okay!  dev site is up at https://makeshift.dev.buttonweavers.com/

Obviously caching is only interesting if something is changing, so i installed these cron jobs:

```
0 * * * * /home/chaos/set_site_type production
20 * * * * /home/chaos/set_site_type staging
40 * * * * /home/chaos/set_site_type development
```

where that's a simple script that changes the site type in both Config.js and the database.  So it *should* be the case that from 0-5 to 20-25 minutes past each hour the site should have the production banner, from 20-25 to 40-45 minutes past each hour it should have the staging banner, and from 40-45 to 0-5 minutes it should have the dev banner.  And you shouldn't have to force-reload or anything to see the new banner.  (The five minute range is because of mod_pagespeed's cache, as described in the PR header comment.)

cgolubi1:
Note that you don't have to force-reload, but you do have to reload --- i don't think testers can leave the monitor running and see the header change colors, because the monitor is activity within one page --- it's not a reload of any kind, cached or not.

cgolubi1:
Standing up a replay site too --- i don't see how this change could break replay testing, but if i don't have to guess, i don't have to guess wrong.

blackshadowshade:
This looks like a nice potential solution to the problem. I'll take a look at this over the next couple of hours.

blackshadowshade:
Looks good to me, I'm seeing the favicon change every 20 minutes when I come back to the page with a normal load.

blackshadowshade:
So, @cgolubi1, I'm happy for this to be merged whenever it fits into your deployment schedule.

cgolubi1:
Sounds good!  Tagging ready_to_merge, and let's get it in next week along with the fanatics.

cgolubi1:
Meanwhile, others should  try out the dev site, and if anything is cause for concern, report it.

dwvanstone:
I've captured a few dice - so far everything looks fine. here.

cgolubi1:
Alright, i'll go ahead and merge this now.

PR #2798, opened 2021-12-31
Detect a class of race conditions in which the game state has changed between load and save

### LINKED ISSUE #2780: Die reappeared after being captured; a different one disappeared
In http://www.buttonweavers.com/ui/game.html?game=78570 my button was Sally and tasha's was BabyBM, whose dice are (7) o(8,8) (11) (21) HG(V). 

In the first round I got initiative and captured tasha's (21); the game log shows that it happened, and the list of buttons I captured showed the (21). 
Except the (21) was still there among tasha's dice -- she used it to attack me on her turn -- while the o(8,8) was not, or at least the log says nothing about it Ornerying afterward, and it certainly was gone when I took my second turn -- on which I captured the (21) a second time, and again the game log and list of captured buttons verified that I had captured a second (21).

cgolubi1:
* Fixes #2780


cgolubi1:
I can do some replay testing on this, but first i want Shadowshade's take on whether my argument in #2780 makes sense, and whether this seems like it would help (or at least not hurt).

blackshadowshade:
Something about this seems wrong to me. I can't exactly put my finger on it, but adding an extra load in the save logic may well hurt, since it slows down the save. Also, using an assert is almost certainly the wrong way to deal with this error.

Let me take another look when it's not so hot and I can actually think about the problem some more.

cgolubi1:
Sounds good.  For what it's worth, my expectation was that the catch block would catch the assert and roll back the transaction.  If that's not the way asserts work in PHP, then i agree we should throw a different type of exception --- that's just a one-line change, not fundamental to the idea.

Anyway, let me know what you think --- i agree there's no rush on this one.

blackshadowshade:
Okay, I've had another look at this, and I think I know which part seems wrong to me.

Why would we use the game state when we could use the last_action_timestamp?

The problem with using game state is that it does not necessarily change after an action takes place. However, the last_action_timestamp should.

Does that sound right to you?

cgolubi1:
Ah, okay.  I didn't think of using last_action_timestamp; i agree i can't think of a reason it wouldn't be preferable.  I'll try it out!

PR #2792, opened 2021-12-30
Undo accidental removal of FQDN config

cgolubi1:
Testing done: without this change, a build of a fresh AWS site fails with:

```
==> default: Error: Could not find dependency Exec[fqdn_populate_etc_file] for Exec[apache_certbot_setup] at /tmp/vagrant-puppet/modules-01f9aab119b6a4b76e30e7c1575836eb/apache/manifests/init.pp:96
```

With this change, it succeeds.  Sorry again!

PR #2789, opened 2021-12-29
Uninstall mysql-server, rather than installing it, on RDS sites

cgolubi1:
* Also cleanup the current IP of the RDS dev site so it can be used for testing

PR #2752, opened 2021-09-06
Migrate BMInterfaceGame.php to BMDB

cgolubi1:
* Partially addresses #712

PR #2799, opened 2021-12-31
Push changes from staging to production on 2021-12-31

cgolubi1:
This update brings these approved pulls to the production branch:

* #2750
* #2766
* #2755
* #2759
* #2752
* #2768
* #2770
* #2641
* #2788
* #2789
* #2792

Database updates:
```
UPDATES=
SCRIPTS=
```


cgolubi1:
I had to reload the site to get vagrant to be able to sync folders again, so that wasn't ideal, but i kind of feel like the only way out is through.  (That should be a one-time thing, and obviously if it's not, i'll debug that rather than continue to do it.)

cgolubi1:
I'm going to leave the merge freeze in place for a day so that if we broke anything, we'll have the decks clear to get a fix through, and then start merging things for next week's release.

If we're doing updates every week, two days between staging and production is probably too long.  Seems like we could go to staging on Thursday and prod on Friday, and that would in fact be plenty of time for whatever staging testing people are going to do, which is likely to be pretty minimal and that's fine.

PR #2802, opened 2022-01-04
Improve apache log grep to only look at PHP errors

cgolubi1:
I realised recently that all the errors recorded by our site PHP code contain the string `[:error]`, whereas errors thrown by other things within apache are `[core:error]` or `[authz:error]` or `[mpm_prefork:notice]` or whatever.  I only want to see the PHP errors, because the purpose of this log grep is to detect code/database problems.

Testing:
* ran this script against a couple of sites to make sure it was valid, including a replay site that had a bunch of database connection failures (and they did all show up as expected)
* did a bunch of poking through my e-mail archives of log error greps to validate the assertion that of the spurious errors i've been excluding, only `" not found or unable to stat"` shows up under `[:error]`


blackshadowshade:
Fine by me.

PR #2806, opened 2022-01-07
Push changes from staging to production on 2022-01-07

cgolubi1:
This update brings these approved pulls to the production branch:

* #2790
* #2796
* #2797
* #2706
* #2775
* #2771
* #2666
* #2802

Database updates:
```
UPDATES=deploy/database/updates/02346_2021fanatics_01.sql
deploy/database/updates/02664_corpo_world_01.sql
deploy/database/updates/02664_corpo_world_02.sql
SCRIPTS=
```


cgolubi1:
Fascinating.  On the prod site, the buttonset ID 86 is already in use, for the set Big Top, which is ID 76 on staging.  Well, i installed CyberCorp as ID 76, and i think we're okay.  Not sure if that's worth ever sorting out.  Definitely not right now.

blackshadowshade:
We hardcode the button set IDs, so it's probably worthwhile cleaning this up so that the database matches what we have in the codebase.

cgolubi1:
> We hardcode the button set IDs, so it's probably worthwhile cleaning this up so that the database matches what we have in the codebase.

I agree, but it's enough of a hassle that maybe it would be better to update the codebase to match prod (and make staging etc match that).  E.g. there is a player who has Big Top as their favorite buttonset, the buttons are attached to the set, etc.  My suggestion is that i futz around on staging to make it match prod, and then we make the codebase match them both.

blackshadowshade:
Yep, I don't mind changing the codebase to match the database, just as long as the two are consistent.

cgolubi1:
Filed #2807 so we don't lose this

PR #2811, opened 2022-01-12
Added btn_special flag to CustomBM to hide it from the live site

### LINKED ISSUE #2810: CustomBM is visible on the live site
At http://www.buttonweavers.com/ui/forum.html#!threadId=1019&postId=23491, Scipio noticed that the CustomBM set and button are visible on the live site.

Decide what to do about this.

If we decide to hide them, perhaps just setting the button_special flag to TRUE might be enough, because this allows CustomBM to be used on dev sites, but not on the live site.

blackshadowshade:
Fixes #2810


blackshadowshade:
When creating a test site for this, remember that we need the site to look like the live site, not a dev site.

cgolubi1:
This looks good to me, and i tested it locally with both development and production set as site type.  Merging it.

Issue #2827, opened 2022-01-25
internal error when withdrawing game between other players via UI


cgolubi1:
If i:
* login as player A
* create a game between players B and C
* withdraw the game i created

I see an internal error:

<img width="1150" alt="internal_error_while_dismissing" src="https://user-images.githubusercontent.com/2898207/150891374-26ca1d17-5faa-46dc-8523-188091c00159.png">

The message in the apache log is:

```
Caught exception in BMInterface::dismiss_game: Found non-set value in DB when expecting integer
```

cgolubi1:
I found this in the prod site apache log, with game 81374 --- this was my best guess as to how to reproduce the issue, and i think it's right.

cgolubi1:
Okay, this is really two bugs:
* The way a game you withdraw gets automatically dismissed, is that the game *UI* sends a `dismissGame` API call after a successful `reactToNewGame` reject call.  Two things about that:
    * That's a little kludgy in any event, because this happening in the UI means if you withdraw a game via the API, it doesn't get dismissed.  It doesn't sit right with me to have the dismiss behavior be different depending on whether you came in via the UI or the API --- seems undocumented and unexpected.
    * However the actual bug is that it calls `dismissGame` whether or not you're not one of the participants in the game, and `dismissGame` isn't valid if you're not a participant
* Second bug: if you call `dismissGame` with a game ID that's not yours, you should get a sensible error, not an internal error

cgolubi1:
I submitted the fix for the `dismissGame` call because it was the easy fix --- the UI result still doesn't look great in this case:

<img width="1150" alt="at_least_its_not_an_internal_error" src="https://user-images.githubusercontent.com/2898207/150895243-9102b29c-ba90-4f71-8f5c-51664cee4035.png">

Anyone have a preference for how we fix the first bug?  My preference is to rip the followon call out of the Game.js UI, and to have the `reactToNewGame` API function, internally, dismiss the game if the person withdrawing it is a participant.  That way the UI doesn't need to do anything silly, and people who withdraw their own games via UI or via API will get consistent behavior (the game will be dismissed).

But this is all sort of ringing a bell with me; did we design the inconsistent behavior on purpose because we think dismissing is a UI thing, so we don't want to make assumptions about it if you don't actually have the game loaded right now this second?  If we did this on purpose, then i guess the fix is to add logic to `Game.js` to do something different if the person withdrawing the game is or isn't a participant.

PR #2828, opened 2022-01-25
Throw a clean error when dismissing a game in which you're not a participant

cgolubi1:
* Partially addresses #2827 

Note: i added the test before fixing the code, and verified that the test would catch the internal error:

```
chaos@ubuntu-xenial:/buttonmen/src$ sudo /usr/local/bin/create_buttonmen_databases && sudo -u vagrant phpunit --bootstrap /usr/local/etc/buttonmen_phpunit.php --group fulltest_deps /buttonmen/test/src/api/
buttonmen already exists
buttonmen_test already exists - recreating it
Creating buttonmen_test
Populating buttonmen_test
mysql: [Warning] Using a password on the command line interface can be insecure.
PHPUnit 5.1.3 by Sebastian Bergmann and contributors.

.Caught exception in BMInterface::dismiss_game: Found non-set value in DB when expecting integer
F                                                                  2 / 2 (100%)

Time: 1.33 seconds, Memory: 10.00Mb

There was 1 failure:

1) responder00Test::test_request_dismissGame
--- Expected
+++ Actual
@@ @@
-'Game 1 does not exist, or you are not a participant'
+'Internal error while dismissing a game'

/buttonmen/test/src/api/responderTestFramework.php:985
/buttonmen/test/src/api/responder00Test.php:1010

FAILURES!
Tests: 2, Assertions: 60, Failures: 1.
chaos@ubuntu-xenial:/buttonmen/src$
```

### DIFF
--- a/src/engine/BMInterfaceGame.php
+++ b/src/engine/BMInterfaceGame.php
@@ -1631,7 +1631,7 @@ class BMInterfaceGame extends BMInterface {
                 'SELECT s.name AS "status", m.was_game_dismissed ' .
                 'FROM game AS g ' .
                 'INNER JOIN game_status AS s ON s.id = g.status_id ' .
-                    'LEFT JOIN game_player_map AS m ' .
+                    'INNER JOIN game_player_map AS m ' .
                     'ON m.game_id = g.id AND m.player_id = :player_id ' .
                 'WHERE g.id = :game_id';
             $parameters = array(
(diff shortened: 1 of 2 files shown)


cgolubi1:
So this should take care of the bug in which `dismissGame` fails with an internal error --- i wonder if we should do an audit of LEFT JOIN calls in general, because definitely this is a case in which, if the right side of the JOIN doesn't match, we want to return nothing, not a result containing NULLs.  I wonder if there are other such calls.  But that's a side note.

blackshadowshade:
The unit test here seems a bit odd to me. I would have thought that it would make more sense to attempt to dismiss the game as a non-participant while the game is still active, as opposed to when it has already been dismissed.

What about shifting those last few lines up a bit, to before the successful dismiss actually happens?

cgolubi1:
I mean, it doesn't matter, because dismissing the game is a per-player action.

I'm happy to move it, because it doesn't matter.

blackshadowshade:
You have it right, I was thinking this action did more than it did.

I'm happy to merge this change.

Issue #2537, opened 2020-02-28
the prod site hangs while the backup is running


cgolubi1:
I was catching up on games on the prod site just now, and i clicked "next game", and the site didn't respond.  I tried a couple of times, no response, so i hit force-reload, and got the header only, no page load.  I noticed the time was just after midnight UTC, so i logged onto the site to take a look, and sure enough, the backup was running.

When the initial backup finished (after about 3 minutes), the site immediately loaded.  (The backup is still being compressed now, and i don't see any issue with site use.)

I assume this happens every day.  It doesn't seem totally ideal for the site to hang for 3 minutes with no information every day.

cgolubi1:
I wonder if we could detect "you can't get a database connection because the backup is running" somehow, and just put up an informative "Site backup is running, wait a few minutes" type message.  I don't know if that would be a better user experience or not.

irilyth:
I bumped into this tonight.

blackshadowshade:
I believe that this issue should have been dealt with. Anyone able to confirm or deny?

cgolubi1:
It's easy to reopen, and it links to the issue which we think should have fixed it, and since we think we fixed it, we're no longer working on it.  So i don't think there's any harm to closing it.

blackshadowshade:
I've just checked this, and while the load delay on pages is longer than when the backup is not running, it's still less than 10 seconds, which is a huge improvement on the previous behaviour.

PR #2845, opened 2022-02-17
Use the --single-transaction flag to make backups less disruptive

### LINKED ISSUE #2808: Deadlock found when trying to get lock
A standard replay loop hit this failure:

```python
Starting game 79
  File "./test_log_games", line 71, in <module>
    c.log_test_game(i, button1, button2)
  File "lib/random_ai.py", line 1803, in log_test_game
    self.next_game_action()
  File "lib/random_ai.py", line 1787, in next_game_action
    if state == 'START_TURN': self._game_action_start_turn()
  File "lib/random_ai.py", line 1668, in _game_action_start_turn
    self.loaded_data['playerDataArray'][1])
  File "lib/random_ai.py", line 1581, in _game_action_start_turn_player
    retval and retval.message or "NULL"))
  File "lib/random_ai.py", line 1683, in bug
    raise ValueError, message
ValueError: API submit_turn(79, 0, 1, {'playerIdx_0_dieIdx_2': 'false', 'playerIdx_0_dieIdx_0': 'false', 'playerIdx_0_dieIdx_1': 'true', 'playerIdx_1_dieIdx_1': 'true', 'playerIdx_1_dieIdx_0': 'false', 'playerIdx_1_dieIdx_2': 'false'}, Power, 2, 1641600061, {}) unexpectedly failed: Internal error while submitting turn
```

The log message is:

```
[Sat Jan 08 00:01:02.227924 2022] [:error] [pid 18424] [client 127.0.0.1:50644] Caught exception in BMInterface::save_game: SQLSTATE[40001]: Serialization failure: 1213 Deadlock found when trying to get lock; try restarting transaction
[Sat Jan 08 00:01:02.235067 2022] [:error] [pid 18424] [client 127.0.0.1:50644] Caught exception in BMInterface::submit_turn: Game save failed
```

The game state looks like:

deadlock

Rerunning the same attack manually, it succeeded with no problem:

deadlock_rerun

cgolubi1:
* Fixes #2808 (as far as i know)

### DIFF
--- a/deploy/vagrant/modules/buttonmen/templates/backup_database.erb
+++ b/deploy/vagrant/modules/buttonmen/templates/backup_database.erb
@@ -23,7 +23,7 @@ else
   MYSQL_ARGS="-u root"
 fi
 
-mysqldump ${MYSQL_ARGS} $DBNAME > $BACKUPFILE
+mysqldump ${MYSQL_ARGS} --single-transaction $DBNAME > $BACKUPFILE
 if [ ! -s "$BACKUPFILE" ]; then
   echo "after backup, backup file is missing or zero size"
   exit 1


cgolubi1:
My initial replay smoke test was to use the default mysqldump settings, and run the backup once a minute (rather than once a day).  With that configuration, it took 2 minutes for the replay loop to fall over with an internal error.  So the problem is reproducible.

Now i'm planning to run with the new settings, and leave that "backup once a minute" cron job installed.  I may run into problems if a backup takes more than 1 minute to run (which is only a problem caused by my test rig for this PR, not related to the issue itself; we've never seen a backup take more than a day to run), but ideally i won't see any problems with replay_loop falling over due to transaction deadlocks.  So far, it's played 40 games over 5-6 minutes, and things look fine, but obviously i want more data.

Do y'all want a dev site running a backup every minute so you can try out the user experience while a backup is running?  Or are you happy to merge if the replay testing seems good and try it out on prod when we get there?  Either way is fine with me.

blackshadowshade:
I think it would be useful to try using the site during a backup, yes. I guess the idea would be for pairs of testers (or a tester with multiple accounts) to try playing turns and setting focus values as quickly as possible. My thought is that we're looking for any problems like #2780.

cgolubi1:
Hmm, i don't think this has anything to do with #2780 from a correctness standpoint.

blackshadowshade:
In that case, I'm happy for us to merge this, it's easy enough to back out. We can test on staging, and we'll get a sense of responsiveness there, as long as we know when the backup is going to happen.

blackshadowshade:
Go ahead and merge this when you're happy with your replay testing, @cgolubi1.

cgolubi1:
My testing ran overnight without falling over, so i think this is at least provisionally better.

blackshadowshade:
Just noting that this pull request will probably fix #2537.

PR #2863, opened 2022-09-21
Don't cache JS files at the AJAX level

### LINKED ISSUE #2818: Pages do not appear to be updated after deployment
At the moment, I'm looking at the live site and I cannot see the changes to broken links that were implemented in #2783.

In #2797, we implemented mod_pagespeed to handle caching of JS pages. My expectation was that the new pages should show the deployed changes, but I'm not seeing this yet.

cgolubi1:
* Fixes #2818 
* Dev site up at: https://temporary.dev.buttonweavers.com

### DIFF
--- a/src/ui/about.html
+++ b/src/ui/about.html
@@ -25,7 +25,6 @@ should run when served either as text/html or application/xhtml+xml.
       $.ajax({
         url: 'js/Loader.js',
         dataType: 'script',
-        cache: true,
         success: function() {
           Loader.loadScripts([
             'js/About.js'
--- a/src/ui/active_players.html
+++ b/src/ui/active_players.html
@@ -25,7 +25,6 @@
       $.ajax({
         url: 'js/Loader.js',
         dataType: 'script',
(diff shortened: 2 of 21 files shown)


cgolubi1:
So this just does the last thing we talked about on #2818.

For testing, i added a line with a datestamp to the hardcoded text in Overview.js, and made a cron job that runs every minute and does:

```
/bin/sed -i -e "s/Page last updated .............../Page last updated $(date +%Y%m%d.%H%M%S)/" /var/www/ui/js/Overview.js
```

So the date you see on the logged-in overview page is hardcoded into the JS file, and if that date is recent (it's in UTC), that means you're getting a recent version of the JS file.  Does that prove we're solving the actual problem actual users are actually having?  Who can possibly say.

blackshadowshade:
Well, the change itself appears minor enough that I see no problems with it. At worst, we'll be back to having longer page load lags without solving the problem.

I'm happy to try this.

PR #2868, opened 2022-12-23
Made changes to enable upgrade to PHP 7.4.33 and MySQL 8.0.31

blackshadowshade:
I've just installed a new version of WAMP on the computer that I'm using to travel to Europe and the U.S.

After managing to get everything running (which was a bit of a headache, since things have changed since the wiki page was written), I noted that unit tests were failing catastrophically and the database_initialise.php script was also failing.

The major issues are that:
- PHP 7.4.x is stricter than previous versions in the way that it deals with non-existent array elements, throwing an error now (see the first point at https://www.php.net/manual/en/migration74.incompatible.php)
- MySQL 8.0.28 upwards changes the default encoding of UTF from a 3 byte to a 4 byte representation (see https://dev.mysql.com/doc/refman/8.0/en/charset-unicode-utf8mb3.html), which causes some of our VARCHAR(255) keys to exceed the maximum key length allowed

The changes that I have made are enough to get the unit tests passing, and games creatable and playable, but there are probably instances of the above compatibility changes that I did not stumble across with my casual testing. I anticipate that there may be other areas where we end up with issues with attempted non-existent array element access. This should not hold up acceptance of this particular pull request, where all changes (except maybe the AMP database access port change) should be backward compatible. However, when we do move up to a newer version of PHP / MySQL, we'll need to do extensive testing.

### DIFF
--- a/deploy/amp/database_initialise.php
+++ b/deploy/amp/database_initialise.php
@@ -32,7 +32,7 @@
 
                 // recreate database
                 print "Creating database $dbname<br>";
-                $query = "CREATE DATABASE `$dbname`;".
+                $query = "CREATE DATABASE `$dbname` CHARACTER SET utf8mb3 COLLATE utf8mb3_unicode_ci;".
                          "CREATE USER :user@:host IDENTIFIED BY :pass;".
                          "GRANT ALL ON `$dbname`.* TO :user@:host;".
                          "FLUSH PRIVILEGES;";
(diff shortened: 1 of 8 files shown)


blackshadowshade:
Passing CircleCI: https://app.circleci.com/pipelines/github/blackshadowshade/buttonmen/384/workflows/8ab101d9-3f71-496d-adc1-0ed22e696520/jobs/760

cgolubi1:
I don't have a problem with this as long as it doesn't blow up in some minimal replay testing.  Trying it out now.

cgolubi1:
This passed 600 replay and novel games, so i think if something is going to blow up, it's not something overwhelmingly obvious.  Merging.

PR #2871, opened 2023-01-26
Add replay support for locally replaying the games that were just generated

cgolubi1:
As described by the new header comments, local replaying is useful functionality because it will catch the case in which the new code works, but can't be responder-tested without changes to `responderTest` code.  Right now, we have to catch that manually, or the hard way after pushing the code, but with this flag, e.g. we can test some wildcard games and see the evidence that responderTest can't yet handle Wildcard, without manual hackeration:

```
About to execute: sudo -u ubuntu sh -c "phpunit --bootstrap /usr/local/etc/buttonmen_phpunit.php --group fulltest_deps /buttonmen/test/src/api/ 2>&1" | tee /srv/bmgames/chaos-test/replay_state/ip-172-31-33-91/1ab6dda9cea5063c2b10e9e6042092b7a62a160c.games.20230126.225110.tar.bz2.output
PHP Parse error:  syntax error, unexpected 'suit_black' (T_STRING), expecting ')' in /buttonmen/test/src/api/responder99Test.php on line 6954
```

### DIFF
--- a/tools/api-client/python/replay_loop
+++ b/tools/api-client/python/replay_loop
@@ -3,17 +3,31 @@
 # This loop is designed for use on a VM which is recording games
 # for use in replay-testing of new code, or on a VM which is
 # replay-testing new code against recorded games.  It does:
-# 1. Play 100 new games using random_ai, generating replay logs in
-#    an internal log format, and caching it in pickle files
-# 2. Translate the 100 games into responderTest's PHP replay format
-# 3. If games are being recorded for future use, tar those generated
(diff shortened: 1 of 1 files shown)


cgolubi1:
I also added direct support for specifying button names to test in a given version --- it's still a code edit, but it's setting a variable rather than remembering the name of the random_ai.py variable to set, which i always get wrong the first eight times.

blackshadowshade:
This only touches the replay loop, I'm fine with changes to that.

PR #2859, opened 2022-08-05
Added tutorial on how to play BM

blackshadowshade:
Fixes #2843.

Depends on #2837.

This is adding a lightly-edited version of @jl8e's tutorial to our site. Thanks very much for the work, @jl8e!

PR #2872, opened 2023-01-27
Changed "Forum" to "Forums" in header text

### LINKED ISSUE #2870: Header text should be "Forums" instead of "Forum"
The link takes the user to Button Men Forums. I think the header should match and say "Forums" instead of "Forum".

blackshadowshade:
Fixes #2870.


blackshadowshade:
This will conflict with the tutorial pull request #2859, so this should probably be added after that one. I'll deconflict it then.

blackshadowshade:
No conflict here.

cgolubi1:
This seems fine.  Merging.

PR #2875, opened 2023-01-27
If local replay tests fail, exit right away so game files are retained

cgolubi1:


blackshadowshade:
These changes only affect the replay loop, I'm fine with that.

PR #2888, opened 2023-02-11
Hardcode a known-working version of miniconda

### LINKED ISSUE #2887: vagrant/puppet regression related to wget
It looks like the new python tests are failing in master branches, because conda is failing to be installed, so it's not installing the python envs.  Here's an example failure on cgolubi1:master: https://app.circleci.com/pipelines/github/cgolubi1/buttonmen/179/workflows/01df0f79-fb2d-41c7-a4b0-a37d43b946a5/jobs/444

I don't fully grok the problem --- it looks like puppet is also failing to install phpmd because of a DNS issue, and i'm not yet sure if that's corellated or is a red herring.

cgolubi1:
If this passes CircleCI, that's evidence that it works around the recent miniconda change and fixes #2887 (though presumably they'll eventually remove the version we're hardcoding, so we'll have to go back to using a newer version eventually).

### DIFF
--- a/deploy/vagrant/modules/buttonmen/manifests/init.pp
+++ b/deploy/vagrant/modules/buttonmen/manifests/init.pp
@@ -121,26 +121,28 @@ class buttonmen::server {
 
 class buttonmen::python-api-client {
 
+  $buttonmen_pyclient_miniconda_version = "py310_22.11.1-1"
+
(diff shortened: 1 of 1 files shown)


blackshadowshade:
I can confirm that with this change, the python2 tests now run successfully. Merging now.

PR #2667, opened 2020-12-10
Python clean up for both python 2 and 3

danlangford:
when trying to run the python i found some scripts only ran with 2 and others only ran with 3. i found some simple improvements to make, some compatibility issues to address, some style clean up to perform. the code now more generally conforms to PEP8. i think there are a couple issues this could address

edit: see the latest take on this PR here https://github.com/buttonmen-dev/buttonmen/pull/2667#issuecomment-972614387

PR #2896, opened 2023-03-04
Expand python unit testing to cover bmutils and several of the scripts

cgolubi1:
I think this pulls a good number of other pieces from #2667 in a way that's not intrusive to replay testing.

Basically what this PR does is add very bare-bones "is it on?" unit testing to bmutils and to *most* of the convenience API client utilities (everything except `monitor` and `interactive_mode.py`), and plug it into a wrapper that circleci runs on python27 and on python39.

So it fixes a couple of python3 incompatibilities in various things, but mostly it unblocks the way for people to add more functionality to the API, and more scripts, and use unit testing to verify that they are more-or-less compatible with both python versions.

The only testing i did is that the unit tests pass, but i think that's sufficient for what we're doing here --- replay testing doesn't use any of these scripts (except bmapi.py, which we handled in a previous PR).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -50,14 +50,16 @@ jobs:
           command: /usr/local/bin/phantomjs --web-security=false /usr/local/etc/run-jscover-qunit.js http://localhost/test-ui/phantom-index.html
       - run:
           name: Run python2 client unit tests
-          command: /opt/conda/envs/python27/bin/python ./test/tools/api-client/python/lib/test_bmapi.py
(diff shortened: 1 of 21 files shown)


cgolubi1:
Successful circleci run: https://app.circleci.com/pipelines/github/cgolubi1/buttonmen/186/workflows/31093c2b-5843-457b-9cb0-e9c555c61a6d/jobs/450

blackshadowshade:
CircleCI passes, and this only touches Python files, so I'm happy to approve this.

PR #2899, opened 2023-03-05
Do an update-vs-rebuild database test within circleci

cgolubi1:
It's gotten a bit time-consuming and annoying to do this check by hand and verify that i've done it.

I believe this script will cause circleci to do the check for me, so people can see their own results.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,6 +33,9 @@ jobs:
           name: Generate checkstyle.xml using PHP_CodeSniffer
           command: php /etc/php/7.0/deploy-includes/phpcs.phar --report=checkstyle --report-file=./build/logs/checkstyle.xml --standard=./deploy/circleci/checkstyle/buttonmen.xml ./src/api ./src/engine
       - run:
+          name: Test consistency of mysql database update files
+          command: /usr/local/bin/branch_database_rebuild_test
+      - run:
           name: Run unit tests using PHPUnit and generates junit.xml and clover.xml
           command: phpunit --bootstrap ./deploy/circleci/phpunit_bootstrap.php --log-junit build/logs/junit.xml --coverage-clover build/logs/clover.xml --debug test/
       - run:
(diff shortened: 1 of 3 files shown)


cgolubi1:
Here's the circleci success: https://app.circleci.com/pipelines/github/cgolubi1/buttonmen/189/workflows/127bfa5b-c500-46d2-a614-1b8d8b73d8f4/jobs/455

And you can see how the test looks when there's no database change at: https://circleci.com/api/v1.1/project/github/cgolubi1/buttonmen/455/output/111/0?file=true&allocation-id=6404060efb6d4d110d67ea37-0-build%2F33C0D75F

I ran the contained commands by hand on a CircleCI container i'd logged into, using a copy of shadowshade's Assailer/Harbinger PR, and they seemed to work reasonably.

blackshadowshade:
I think it would be a good idea to skip the database comparison if there are no new update files in the pull request, and I don't believe that's currently in the proposed code---correct me if I've missed something.

cgolubi1:
Why?  I could, but it's more logic that i could get wrong, and the test takes 0s according to CircleCI's self-reported metric.

blackshadowshade:
So, what does the output look like when there is a difference between the databases? Does this cause CircleCI to fail appropriately?

cgolubi1:
Here, this should show it: https://app.circleci.com/pipelines/github/cgolubi1/buttonmen/190/workflows/1edc9f39-2722-46e6-906c-e6eb06d06d10

(Obviously don't merge until i revert this.)

blackshadowshade:
Very good, that fails with a full diff and:
 
    Exited with code exit status 1
    CircleCI received exit code 1

Go ahead and revert the change, and I'll accept this pull request.

cgolubi1:
Reverted!  Assuming the build succeeds, should be back to the original change.  Thanks.

blackshadowshade:
Looks good to me! Accepting this now.

Issue #2908, opened 2023-05-08
migrate away from vagrant to deploy buttonmen


cgolubi1:
We're using vagrant to manage prod/staging/dev/replay buttonmen sites on EC2 instances (we use docker-based circleci to run regression tests).  This has served us okay so far, but, since we committed to the vagrant+instance path, docker+container has kind of become the done thing.  There are a couple of specific reasons we might want to migrate:
* Support for vagrant itself is okay, but the `vagrant-aws` plugin, which we totally rely on, is not being maintained.  We should move away from that dependency.
* Building on containers (and moving to a container-based platform like ECS) would ultimately let us do some cool things, mostly around automating parts of the replay/dev testing lifecycle, and taking advantage of support for dormant containers (so that i would only have to pay for dev sites when someone was actively using them).  We'd have to do a bit of work to get that stuff, but moving to containers is step 1 of that work

This issue is to make and implement a plan for migrating site installs from vagrant+EC2 to docker+ECS.

Problems that need to be addressed (this list is incomplete):
* Problems which are NOT blockers for migrating staging and prod:
    - [x] Support replay sites - #2952
    - [ ] When ECS replaces a dev site, the database is lost
    - [ ] When ECS replaces a dev site, DNS needs to be updated
    - [ ] Fix the sed -i thing that yields: deploy/vagrant/manifests/init.pp-e
    - [ ] Make a way to periodically run: docker system prune (or detect that this is needed)
    - [x] Setup cloudwatch logging of container startup (awslogs?) - #2929
    - [ ] Look into using a real process manager in the container / understand support for clean shutdowns on deployment - e.g. https://aws.amazon.com/blogs/containers/graceful-shutdowns-with-ecs/
    - [x] When a new container comes up, it should e-mail people who may need to SSH in to notify them about its public IP address - #2941
    - [x] Mount /var/log/apache2 from the EFS volume so we retain apache logs across containers - #2942
 * Problems which ARE blockers for migrating staging and prod:
    - [x] Merge #2925
    - [x] Support for associating an EIP with a site (via NLB, probably) - #2927
    - [x] When ECS replaces a site with an EIP, rerun certbot - #2935
    - [x] When ECS replaces a site with a remote database, buttonmen_db.cnf is lost - #2930
    - [x] Support attaching an FS to a site so that backups can be stored persistently - #2935
    - [x] Fix the private-IP-specific mysql grants for bmuser1 on the staging and prod DBs (done for staging)
[...]

PR #2919, opened 2023-11-28
ECS deployment bugfix: make sure tasks exist before describing them

cgolubi1:
Partially addresses #2908 

Each dev branch is its own ECS service (containing a single task).  Without this change, the first time i create a dev cell for a new branch, the service is created before the task is created, and the script blows up trying to describe the empty tasks list.

I tested this change on both an ECS service with existing tasks, and an ECS service with no existing tasks.


blackshadowshade:
Merging this now.

PR #2920, opened 2023-12-28
Add support for ECS sites with remote databases

cgolubi1:
This partially addresses #2908 - specifically it adds support for using an ECS site with a remote database, which is required by staging and prod

* As part of this PR, i cleaned up the places in the existing vagrant config where we hardcode locations and names of specific buttonweavers sites, because all of that will be handled by the config file in the future
* I also made the container startup able to handle the case in which the container doesn't have mysqld installed, which is the case for remote-db sites
* I also fixed an unrelated bug where puppet failures didn't cause the overall container deployment to fail fast, leading to harder-to-diagnose problems using the image

I tested this by creating a dev site using a remote DB and this codebase, and then rebuilding that dev site using a local DB.  It's currently up at https://2908-remote-db.cgolubi1.dev.buttonweavers.com/ (with the local DB), and it should be uninteresting.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,6 +10,7 @@ RUN mkdir -p build/logs
 RUN bash ./deploy/vagrant/bootstrap.sh
 
 # Run puppet
-RUN puppet apply --modulepath=/buttonmen/deploy/vagrant/modules /buttonmen/deploy/vagrant/manifests/init.pp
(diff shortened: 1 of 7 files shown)


blackshadowshade:
Okay, the new dev site looks okay and the files look like they have to do with setup. Looks good so far!

PR #2921, opened 2024-01-03
Script to teardown AWS resources for a dev site

cgolubi1:
* Part of development work for #2908 

This is just a script for me to use to teardown a dev site once we're done testing with it --- it cleans up ECS tasks (which cost money), ECS services (which are confusing to have lying around in the cluster), and ECR repositories (which cost money).  It doesn't bother to remove DNS entries or ECS task definitions, which i believe are not billed at any scale we're likely to hit, and which aren't in the way.

Before tearing anything down, it downloads the most recent DB backup of the dev site, in case we need it for anything later.

[I tested it by using it to tear down the dev-cgolubi1-2908_remote_db site.]

### DIFF
--- a/deploy/docker/teardown_buttonmen_site
+++ b/deploy/docker/teardown_buttonmen_site
@@ -1,0 +1,236 @@
+##### teardown_buttonmen_site
+# Tear down all resources for the buttonmen site represented by this
+# working directory
+# The caller is responsible for:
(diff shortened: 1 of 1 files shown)


blackshadowshade:
This is a new file, appears to be entirely to do with teardown. Merging this now.

PR #2925, opened 2024-02-17
Move some static post-site configuration to be built into the docker image

cgolubi1:
Related to #2908 but does not complete it

Specific things this PR moves into the docker image build:
* Populate /usr/local/etc/bmsite_fqdn at docker build time
* Hardcode site type in set_buttonmen_config since it's knowable at build time
* Run set_buttonmen_config at build time (since this will be correct for local test sites that use the autogenerated DB, and for sites which connect to a remote DB which will already have the correct type)
* Set an environment variable containing the site FQDN in the root cron job, so that cron e-mails can quickly be associated with their site

The purpose of this PR is to reduce the number of steps which need to be done on a container after it is replaced by ECS.  Since ECS treats containers as truly ephemeral, any site which relies on post-replacement configuration will become unavailable at unexpected times.

This PR doesn't fully solve the problem --- here's what i'm aware of that remains unsolved (i'll add these things to #2908 to be addressed later):
* Load of a database for use by a dev site isn't handled, so any dev site will have its database reset to clean state (without e.g. staging users) on ECS rebuild.
* We don't populate `/usr/local/etc/buttonmen_db.cnf`, so administrative access (e.g. backups) will fail on sites with remote databases after ECS rebuild.
* We don't rerun certbot, so SSL will be broken on sites after an ECS rebuild.  The fix for this will be split into sites with EIPs, vs dev sites:
    * For dev sites, we can't rerun certbot because the DNS is wrong, and DNS updates are handled outside of ECS.  So there's not much point addressing the certbot part until/unless we can address the DNS part.
    * For sites with EIPs, we can't run certbot at docker build time, but we can run certbot as part of container startup, since the container will be starting up behind the NLB which is attached to the EIP certbot wants.  I'll handle this case when i implement EIP support.


cgolubi1:
This dev site was built using this code - https://2908-dockerize-config.cgolubi1.dev.buttonweavers.com/ - it is a dev site with a local DB.

cgolubi1:
Here's the successful circleci run: https://app.circleci.com/pipelines/github/cgolubi1/buttonmen/205/workflows/e0eea153-5301-4ca2-846c-d41e5540ce8b

cgolubi1:
Actually, that circleci run isn't good:

```
Error: /usr/local/bin/set_buttonmen_config returned 1 instead of one of [0]
Error: /Stage[main]/Buttonmen::Server/Exec[buttonmen_set_config]/returns: change from notrun to 0 failed: /usr/local/bin/set_buttonmen_config returned 1 instead of one of [0]
```

(I now know why CircleCI ignores errors - i filed #2926 to fix that because it's not part of this project.)  But let me put in a one-line modification to this PR to address that.

cgolubi1:
New revision solves that problem:

```
Notice: /Stage[main]/Buttonmen::Server/Exec[buttonmen_set_config]/returns: executed successfully
```

but checks fail, so that's interesting.

cgolubi1:
Ah, the failure isn't actually a surprise --- there's a UI test, `test_Config_is_loaded()`, which expects the site type to be production.  Sure, that's fine.

blackshadowshade:
CircleCI passes. I note 4 errors in the puppet run that relate to phpmd and phpdocumentor, these are peripheral.

Files all pertain to deployment.

Merging now.

cgolubi1:
> I note 4 errors in the puppet run that relate to phpmd and phpdocumentor, these are peripheral.

Yeah, these are why i filed #2926 rather than just fixing it --- the fix is approximately a one-liner, but fixing it would make all of our runs start failing.

blackshadowshade:
Yup, just thought I would note it explicitly so that we have something to search for in the future if necessary.

PR #2927, opened 2024-02-18
Add support for attaching containers to an NLB so they can be used with an elastic IP

cgolubi1:
Partially addresses #2908 

Note: the creation and configuration of the NLB itself is not part of this code.  The NLB has to have:
* A mapping in the AZ in which the ECS service's subnet is located, attached to the EIP whose DNS FQDN is the `bmsite_fqdn` used in the configuration here
* Two attached port forwardings, one for each of two target groups, one on port 80 and one on port 443.  (The ARNs of the two target groups must be the `nlb_arn_port_80` and `nlb_arn_port_443` used in the configuration here.)

Configuring the ECS service to use the target groups, means that new containers will be registered with those target groups, and unhealthy container deregistered.

I also fix a regression with the `set_buttonmen_config` change in my previous PR --- obviously in the case in which we're attaching a remote DB, we can't set_buttonmen_config while creating the docker image, because the DB isn't attached yet at that time.  So we have to do it after launching.  I believe it could be done at container startup time, but i'll sort that out later --- i've added it as a TODO on #2908.

Testing: i checked for regressions using a local-database/no-EIP container (which i've now torn down).  This site is now backed by an ECS service attached to this dev EIP (and a remote database, just to keep things tidy), and was created using the code in this PR: https://rds.dev.buttonweavers.com/ui/

### DIFF
--- a/deploy/docker/buttonmen_ecs_config.json
+++ b/deploy/docker/buttonmen_ecs_config.json
@@ -1,17 +1,24 @@
 {
   "production": {
+    "bmsite_fqdn": "FIXME",
     "network_subnet": "FIXME",
     "network_security_group": "FIXME",
+    "nlb_arn_port_80": "FIXME",
+    "nlb_arn_port_443": "FIXME",
     "remote_database_fqdn": "FIXME",
     "remote_database_admin_pw": "FIXME"
   },
   "staging": {
+    "bmsite_fqdn": "FIXME",
     "network_subnet": "FIXME",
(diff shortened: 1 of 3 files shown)


blackshadowshade:
CircleCI passes: https://app.circleci.com/pipelines/github/cgolubi1/buttonmen/209/workflows/642dc538-0f75-42d7-8167-f04413678d8e/jobs/475

Files changed are all to do with deployment.

Merging now.

PR #2934, opened 2024-03-08
Implement `cartesian` tool for projects of getting sets of buttons to game counts

agathazeren:
It's left as a draft because i need to eliminate the commenting out of the future imports. (probably either setting up virtualenv and using an older version of python or updating the codebase, but maybe it's something else that I have to research.)

### DIFF
--- a/tools/api-client/python/cartesian
+++ b/tools/api-client/python/cartesian
@@ -1,0 +1,133 @@
+#!/bin/python
+
+import configparser
+import argparse
+import os
+import sys
+import json
+import time
+
+# Import Button Men utilities.
+
+sys.path.append('lib')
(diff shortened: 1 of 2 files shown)


blackshadowshade:
I reckon this could do with a little bit of commenting so that it's obvious what it does---both for others and for future you.  :)

agathazeren:
I agree. I'm not sure when I will get this sorted, but I'll try to get to it in not too long.

blackshadowshade:
Just noting that the correct way to be bringing in updates into your branch is to rebase your branch instead of merging our changes. This means that your branch becomes relative to our current code status and that we don't have merge issues.

PR #2941, opened 2024-05-05
Send a hello world e-mail whenever a new non-sandbox site is started

cgolubi1:
Partially addresses #2908 

I've left the dev site running; you should be able to use the e-mail which was sent to buttonmen-help to get the IP address to SSH to this dev site's container.

### DIFF
--- a/deploy/docker/startup.sh
+++ b/deploy/docker/startup.sh
@@ -50,5 +50,8 @@ fi
 # Set site type so it's correct in UI JS
 /usr/local/bin/set_buttonmen_config
 
+# Send hello world e-mail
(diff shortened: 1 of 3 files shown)


cgolubi1:
(Obviously the dev site isn't a perfect analogue, because dev sites aren't behind load balancers, so you can just get the address from the FQDN.  But there's no reason this method won't work for a site behind an LB, so i think the example should be good enough.)

blackshadowshade:
I'm happy to give this a go, we'll see if there are any edge cases that occur in practice. Changes look reasonable.

PR #2942, opened 2024-05-05
Mount /var/log/apache2 from EFS so http logs will be retained across container replacement

cgolubi1:
I created a dev site for this, and it looks good:

```
$ ls -l /var/log/apache2
lrwxrwxrwx 1 root root 89 May  5 20:10 /var/log/apache2 -> /mnt/efs/2908-retain-apache-logs.cgolubi1.dev.buttonweavers.com/apache_logs/ip-10-0-0-205
```

```
$ ls -lart /var/log/apache2/
total 68
-rw-r-----  1 root adm      0 May  5 19:52 other_vhosts_access.log
drwxr-xr-x 17 root root  6144 May  5 20:10 ..
drwxr-x---  2 root adm   6144 May  5 20:10 .
-rw-r-----  1 root adm   2372 May  5 20:15 error.log
-rw-r-----  1 root adm  51675 May  5 20:17 access.log
```

And the contents of the access and error logs look normal.

### DIFF
--- a/deploy/docker/startup.sh
+++ b/deploy/docker/startup.sh
@@ -10,8 +10,11 @@ set -x
 /etc/init.d/ssh start
 /etc/init.d/postfix start
 
-## Remote filesystem setup
+# Host identity variables
 FQDN=$(cat /usr/local/etc/bmsite_fqdn)
+HOSTNAME_LOCALIP_PART=$(/bin/hostname | awk -F\. '{print $1}')
+
+## Remote filesystem setup
 MNT_DIR="/mnt/efs/${FQDN}"
 if [ ! -e "${MNT_DIR}" ]; then
   mkdir ${MNT_DIR}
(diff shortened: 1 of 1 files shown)


deploy/docker/startup.sh
```diff
@@ -26,6 +29,18 @@ fi
...
 
+# Replace the /var/log/apache2 in the image with a remotely-mounted one
+# Add a per-container piece because logs may be written during deployments
+APACHE_LOG_DIR="${MNT_DIR}/apache_logs/${HOSTNAME_LOCALIP_PART}"
+if [ ! -e "${APACHE_LOG_DIR}" ]; then
+  mkdir -p ${APACHE_LOG_DIR}
+  chown root:adm ${APACHE_LOG_DIR}
+  chmod 750 ${APACHE_LOG_DIR}
+fi
+mv /var/log/apache2/* ${APACHE_LOG_DIR}/
```
cgolubi1, line 40: Oh, huh, this is a ticking time bomb for when a private IP gets reused, which won't happen on dev sites, but will eventually happen on prod and staging.  Lemme make this `mv -f`.  (That's not very harmful to do because we only care about recent logs, and private IPs won't get reused very quickly; they just eventually will, and a deployment will fail mysteriously and we'll be sad.)

deploy/docker/startup.sh
```diff
@@ -26,6 +29,18 @@ fi
...
+# Add a per-container piece because logs may be written during deployments
+APACHE_LOG_DIR="${MNT_DIR}/apache_logs/${HOSTNAME_LOCALIP_PART}"
+if [ ! -e "${APACHE_LOG_DIR}" ]; then
+  mkdir -p ${APACHE_LOG_DIR}
+  chown root:adm ${APACHE_LOG_DIR}
+  chmod 750 ${APACHE_LOG_DIR}
+fi
+mv -f /var/log/apache2/* ${APACHE_LOG_DIR}/
+rmdir /var/log/apache2
+ln -s ${APACHE_LOG_DIR} /var/log/apache2
```
blackshadowshade, line 42: So, in this process of moving, removing the directory, and creating the symbolic link, is there a chance that Apache may wish to log something happening? And if it doesn't find a directory to log to, what happens then? Does it create a directory, which would cause the symbolic linking to fail? Or does it fail silently? Or something else?

blackshadowshade: It looks like ln -Ff might be able to do what we need to handle the case where the directory already exists.

cgolubi1: That shouldn't be a problem --- this script is run at container startup, and apache isn't running yet.  If you look at the rest of deploy/docker/startup.sh, you'll see that apache gets started later, around line 60.  The only reason logfiles exist is that they're created when apache is installed as part of docker image creation; they shouldn't be being written to at this point.

blackshadowshade: Very good, approving this now.

PR #2947, opened 2024-05-15
Script to print the IP address of this buttonmen site's launched container

cgolubi1:
This is just a utility for me to use, but it'll help with running post-deploy database updates for future code pushes, which is on the checklist for #2908.

Testing:

```
$ env AWS_PROFILE=bm_deploy /Users/chaos/games/buttonmen/miniconda3/bin/python ./deploy/docker/get_buttonmen_site_ipaddr 
54.165.109.196
```

That matches what i get from DNS (but note that DNS only works because the container hasn't been replaced since DNS was set, and won't work for staging or prod --- this should):

```
$ host 2908-get-ipaddr-script.cgolubi1.dev.buttonweavers.com
2908-get-ipaddr-script.cgolubi1.dev.buttonweavers.com has address 54.165.109.196
```

### DIFF
--- a/deploy/docker/get_buttonmen_site_ipaddr
+++ b/deploy/docker/get_buttonmen_site_ipaddr
@@ -1,0 +1,157 @@
+##### get_buttonmen_site_ipaddr
+# Retrieve the IP address of the ECS container running the site,
+# and print it to STDOUT
+# The caller is responsible for:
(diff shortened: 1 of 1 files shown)


blackshadowshade:
File has nothing directly to do with game code, is only to do with the deployment. Merging now.

PR #2952, opened 2024-05-23
Minimal container support for replay sites

cgolubi1:
* Partially addresses #2908 

This adds support for a "replay" site type, which can be invoked with a flag (`-p`) to `deploy_buttonmen_site`.

* replay sites have a simpler startup than dev/staging/prod sites --- they don't setup HTTPS, they don't have the option of using a remote database, etc
* i made some changes to replay_loop to work more consistently in the constrained container environment:
    * restart mysql and apache between test runs, for consistency
    * log information about novel games that were played, for debugging

I've tested:
* dev sites still work, and a dev and replay site can be deployed for the same branch at the same time without conflicts
* replay testing generally works, and in particular ran for 22 hours without interruption

### DIFF
--- a/deploy/docker/buttonmen_ecs_config.json
+++ b/deploy/docker/buttonmen_ecs_config.json
@@ -29,5 +29,11 @@
     "log_group": "FIXME",
     "dns_update_script_path": "FIXME",
     "load_database_path": "FIXME"
+  },
+  "replay": {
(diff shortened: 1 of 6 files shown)


blackshadowshade:
Looks okay to me. Happy to give it a go.

PR #2954, opened 2024-05-23
Fix mechanism for finding the current git commit ID during archived replay testing

cgolubi1:
I think this broke because i was previously checking out my own primary branch and then overlaying the branch under test, and now i'm checking that branch out directly.

Regardless, it's clearly more portable to ask git for the commit ID than to try to figure it out by looking at files in .git.  And with this change, `replay_loop -a` (archive) doesn't blow up.


blackshadowshade:
Happy to give this a go. Merging now.

Issue #2967, opened 2024-07-25
stuck game on prod site


cgolubi1:
https://www.buttonweavers.com/ui/game.html?game=103527 shows: "Can't figure out what action to take next"

cgolubi1:
According to the database, the game is in state 46, aka `CHOOSE_TURBO_SWING`.  So that explains why it's stuck --- a game shouldn't be able to get into that state and the UI can't deal with it --- but not how it got there.

cgolubi1:
The matchup is Boot2daHead (`d(1) Gz(Z) !p(V) !n(Y,Y) oF(C)`) vs Aysez (`s(U) z(U) %(U) n(U) g(U) t(U)`).

cgolubi1:
According to the action log, the last thing that happened was a skill attack in which `[d(1):1, p(V=7)!:6, n(Y=4,Y=4)!:3]` attacked `%(U=30):13`.

The attack was fire-assisted by `oF(C)`, which was turned down from 5 to 2.

That should all be fine.  We've played lots of replay games with Boot2daHead over time; i guess it's possible that none of those games has ever tried a fire-assisted attack where both turbo dice participate in the attack, but that strikes me as unlikely.

Hmm.  What is *this* though:

```
mysql> select * from game_turbo_cache where game_id=103527;
+---------+---------+------------+
| game_id | die_idx | turbo_size |
+---------+---------+------------+
|  103527 |       2 |        127 |
|  103527 |       3 |         20 |
+---------+---------+------------+
2 rows in set (0.02 sec)
```

I don't know much about how this cache is used, but that first row doesn't look good.

cgolubi1:
Okay, i got it.  I reproduced this in https://2928-mysql-80.cgolubi1.dev.buttonweavers.com/ui/game.html?game=262 --- it looks like when you enter turbo values during an attack that needs firing, those values aren't checked against the valid bounds.  So i typed V=127 into the turbo box, submitted the skill attack, turned down fire dice, and here we are.

(I can do that again and get screencaps if it's not completely obvious what i'm talking about, but hopefully it is.)

That's the bug: the skill attack should have been rejected because the turbo selections submitted with that attack were invalid.  Never mind that they're going into the cache and not being applied immediately, they should have been checked and rejected.

cgolubi1:
It looks like unsticking the game may not be trivial:

```
mysql> select * from game_turbo_cache where game_id=262 and die_idx=2;
[...]

PR #2972, opened 2024-07-31
Avoid putting the game in a broken state if invalid turbo settings are submitted for fire-assisted attack

cgolubi1:
* Partially addresses #2967 

This is the simplest possible fix for the root cause.

Background:
* Fire-assisted attacks are performed in two parts, first setting up the attack for fire turndown, and then the fire turndown happens and the attack is committed
* Setting turbo values always happens after the attack is resolved.
* Because in the UI the turbo values are set on the first screen, we cache those values from `submitTurn`, and then get them out of the cache and apply them when we adjust fire dice.

The reason we get games into a broken state, is simply that the fire-adjusted path has its own implementation of applying turbo values, and it doesn't check the return value from `set_turbo_sizes()`, such that if `set_turbo_sizes()` fails, we just plow on ahead, and importantly, we plow on ahead to `save_game()`, committing the broken state in the DB.

This is the simplest possible fix --- it checks the return value, and tells the player if it failed.  And it does *work* (i'm attaching some screencaps below), because if the player abandons the fire turndown, that abandons the cached turbo values, and the player can start over.

It's not the most user-friendly --- it would be marginally more friendly if we rejected the turbo values on the initial screen, rather than making the player back their attack out.  However, i think we should stick with this, because:
* This is the first time we've heard of a player actually running into this case, and we're several years into supporting Turbo+Fire
* Detecting the problem on the first screen would require something different would require something like the `BMAttackTrip::attacker_apply_turbo()` workflow where we optimistically apply the turbo changes on cloned dice just to test them.  That's more complexity and more opportunity to introduce errors; since the player *can* rescue themselves with this fix, it doesn't seem worth it.

[dev site: https://2967-invalid-turbo-swing-fire.cgolubi1.dev.buttonweavers.com/ui/ ]

### DIFF
--- a/src/engine/BMInterfaceGame.php
+++ b/src/engine/BMInterfaceGame.php
@@ -1532,13 +1532,18 @@ class BMInterfaceGame extends BMInterface {
                 $game->proceed_to_next_user_action();
 
                 if (!is_null($game->turboCache)) {
-                    $this->set_turbo_sizes(
+                    if (!$this->set_turbo_sizes(
                         $playerId,
                         $game,
                         $roundNumber,
                         'ignore',
                         $game->turboCache
-                    );
+                    )) {
+                        $this->set_message(
+                            'Setting turbo dice failed.  Cancel Fire turndown and retry the initial attack.'
+                        );
+                        return NULL;
+                    }
                 }
 
                 $this->save_game($game);


blackshadowshade:
This looks like a pretty simple fix. I figure we fire up a dev site for it and take a look.

cgolubi1:
Demonstrating this fix, in six screencaps:

01_invalid_turbo
02_fire_turndown
03_error_message
04_back_to_turbo_screen
05_fire_turndown_again
06_success

blackshadowshade:
Looks good. If we can get testers on the dev site to replicate that behaviour, I'm happy to approve this as-is.

cgolubi1:
Dev site is up at https://2967-invalid-turbo-swing-fire.cgolubi1.dev.buttonweavers.com/ui/ --- please take a look!

blackshadowshade:
Still looking for testers to give this a go.

PoshFrosh:
> Dev site is up at https://2967-invalid-turbo-swing-fire.cgolubi1.dev.buttonweavers.com/ui/ --- please take a look!

@blackshadowshade @cgolubi1 
I learned about this on discord and gave it a shot. Game 502:

image

image

image

image

image

image

dwvanstone:
I've tested this as well -- including the case of fire overshooting on a power attack (with turbo). It looks good.

blackshadowshade:
Thank you for testing, both of you!

cgolubi1:
@blackshadowshade what else does this need before you'd be good to merge it?

blackshadowshade:
Thank you for the reminder, merging now.

PR #2973, opened 2024-07-31
allow https remote repos + add gitignore

kierajreed:
A simple change to let deploys work with https remotes on github, which I use instead of ssh remotes.

Also added a gitignore to allow storing temp files in the repo directory without cluttering up git stuff.


deploy/docker/deploy_buttonmen_site
```diff
@@ -26,7 +26,8 @@ LOCALHOST_IPV4 = '127.0.0.1'
 def get_subprocess_output(cmdargs):
   return subprocess.check_output(cmdargs).decode()
 
-REPO_MATCH = re.compile('^origin\s+git@github.com:([\w-]+)/buttonmen.git \(fetch\)$')
+REPO_MATCH_SSH = re.compile('^origin\s+git@github.com:([\w-]+)/buttonmen.git \(fetch\)$')
+REPO_MATCH_HTTPS = re.compile('^origin\s+https://github.com/([\w-]+)/buttonmen.git \(fetch\)$')
```
cgolubi1, line 30: Can you make these changes to `get_buttonmen_site_ipaddr` and `teardown_buttonmen_site` as well?  You won't need them, but i've been keeping them in sync, in lieu of having had ten seconds to do anything halfway intelligent with modules.

cgolubi1:
@kierajreed - can you make the same changes to the two other scripts in this directory that contain the same regexps?  I'm happy to try this out and approve it if so.

.gitignore
```diff
@@ -0,0 +1 @@
+tmp/
```
cgolubi1, line 1: Another drive-by comment: the red circle means this file is missing a final newline.  Can you fix that in the next revision?

(We have code checks that make sure we're not committing source files without final newlines, but they don't check 100% of the things in the source directory, so this new file you're adding doesn't seem to be checked.)

cgolubi1:
Closing this out because the critical fix was made in #3006.  Feel free to open a new PR with a gitignore if you want one!

PR #2985, opened 2024-09-01
Log error message on API response json_encode() failure

### LINKED ISSUE #2983: we don't get a log message if JSON encoding fails
This is something i discovered during testing around #2974 but is worth addressing independently.

The ApiResponder.php code which returns API responses to the user ends with:

```
            header('Content-Type: application/json');
            echo json_encode($output);
```

JSON encoding fails silently, so if it fails for whatever reason, the user will get red error text, e.g. `Internal error when calling loadForumThread`, but nothing shows up in the error log.  This means people could be getting that kind of error message all the time, and we have no way of knowing unless they complain.

Proposed fix:

```
            header('Content-Type: application/json');
            echo json_encode($output);
            $last_error = json_last_error();
            if ($last_error) {
                error_log("API call with spec: " . var_export($args, TRUE) . " yielded a JSON error: " . $last_error);
            }
```

If we made that change, we'd be notified via error log whenever someone didn't get their API response.

cgolubi1:
* Fixes #2983 

Testing:
* I did a bit of manual testing, which i'll describe in a comment
* There's a dev site so people can poke around --- this should be a noop; we're not aware of anything that happens on a normal site that should lead to blank pages, so i really just want a smoke test of normal site activity: https://2983-json-encode-logging.cgolubi1.dev.buttonweavers.com
* I stood up a replay site, because testing of novel games uses the API, so if anything goes wrong during gameplay, it should show up in those logs

### DIFF
--- a/src/api/ApiResponder.php
+++ b/src/api/ApiResponder.php
@@ -1050,6 +1050,13 @@ class ApiResponder {
         } else {
             header('Content-Type: application/json');
             echo json_encode($output);
+            $last_error = json_last_error();
+            if ($last_error) {
+                $last_error_msg = json_last_error_msg();
+                error_log(
+                    'Response to API call ' . var_export($args, TRUE) . ' failed json_encode(): ' . $last_error_msg
+                );
+            }
         }
     }
 


cgolubi1:
Manual testing on a local docker site:
* Various UI actions, including creating and playing some games, posting and viewing some forum posts
* Simulated a JSON error by hardcoding a bad item into the output:

```
            $output['data']['foobar'] = NAN;
```
Then loaded a forum page, and verified that i got no output on the page, and the error log got:

```
[Sun Sep 01 19:22:54.167891 2024] [:error] [pid 1216] [client 172.17.0.1:36850] Response to API call array (\n  'type' => 'loadPlayerName',\n) failed json_encode(): Inf and NaN cannot be JSON encoded, referer: http://127.0.0.1:8080/ui/forum.html
```

cgolubi1:
The replay site has played 1500 games, and hasn't logged any error messages or seen any failures, so i think that's all we're going to get from that smoke test.

cgolubi1:
CaptConstruct did some smoke-testing on the dev site.  Per discord:

> I did some stuff in my profile, joined a few open games, selected die sizes, flipped through some of the other pages.

cgolubi1:
I'm going to take the replay site down.

Shadowshade, when you get a chance, let me know if this seems okay to merge, if you have concerns about the code, or if there's anything else you want to see tested.

blackshadowshade:
This looks fine to me. Merging now.

Issue #2989, opened 2024-10-04
the backend allows players to pass when they have valid moves


cgolubi1:
Splitting this out of #2687 because it's a separate bug, and is pretty urgent.

The backend allows players to submit `Pass` attacks via the API under at least some circumstances when they have valid attacks they could make.  Here's a screencap of an attack which should not have succeeded, but did.  I'm including the responder test as a comment.

<img width="1164" alt="pass_allowed_incorrectly" src="https://github.com/user-attachments/assets/766d519d-a327-40a8-8d0a-4259396b77e2">

cgolubi1:
Here's a new responderTest to validate the fix:

```
<?php
require_once 'responderTestFramework.php';
class responder99Test extends responderTestFramework {

    /**
     * @depends responder00Test::test_request_savePlayerInfo
     * @group fulltest_deps
     */
    public function test_interface_game_00064() {

        // responder003 is the POV player, so if you need to fake
        // login as a different player e.g. to submit an attack, always
        // return to responder003 as soon as you've done so
        $this->game_number = 64;
        $_SESSION = $this->mock_test_user_login('responder003');


          $gameId = $this->verify_api_createGame(
              array('bm_rand' => array(1, 4, 4, 12, 4, 10, 3), 'bm_skill_rand' => array()),
              'responder003', 'responder004', 'Cheathem', 'Mutton Ben', 3,
              '', NULL, 'gameId', array()
          );

        $expData = $this->generate_init_expected_data_array($gameId, 'responder003', 'responder004', 3, 'SPECIFY_DICE');
        $expData['gameSkillsInfo'] = $this->get_skill_info(array('Mighty', 'Ornery', 'Poison', 'Rush', 'Shadow'));
        $expData['playerDataArray'][0]['button'] = array('name' => 'Cheathem', 'recipe' => 'Ho(1) s(6) o(10) o(14) s(X)', 'originalRecipe' => 'Ho(1) s(6) o(10) o(14) s(X)', 'artFilename' => 'cheathem.png');
        $expData['playerDataArray'][1]['button'] = array('name' => 'Mutton Ben', 'recipe' => 'p(8) #(12) s(20) (X) (X)', 'originalRecipe' => 'p(8) #(12) s(20) (X) (X)', 'artFilename' => 'muttonben.png');
        $expData['playerDataArray'][0]['swingRequestArray'] = array('X' => array(4, 20));
[...]

PR #2992, opened 2024-10-12
With small probability, try an invalid attack, and make sure it fails

cgolubi1:
I created this for use in reproducing the bug in #2989, and wanted to commit it.

It's not perfect, and it doesn't directly test any new cases since the fix for the outstanding issue should be pretty solid.  But it establishes a framework for expecting `submitTurn` to fail, and that's something i should be able to build on to do a lot more fuzz-testing of things that shouldn't work, so i'd like to get it in the codebase.

### DIFF
--- a/tools/api-client/python/lib/random_ai.py
+++ b/tools/api-client/python/lib/random_ai.py
@@ -725,6 +725,7 @@ class LoggingBMClient():
     # Any state that's needed so we can tailor future random decisions based on past ones
     self.decision_state = {
       'cancelled_fire_attacks': 0,
+      'expected_start_turn_status': 'ok',
     }
     self.log = []
 
@@ -757,6 +758,10 @@ class LoggingBMClient():
   def _list_all_idx_combos(self, list_len, combo_len):
     return [x for x in itertools.combinations(range(list_len), combo_len)]
 
+  def _invalid_attack_okay(self):
+    if self.decision_state['expected_start_turn_status'] == 'failed': return True
+    return False
+
   def _look_for_attacker_defender_combo(self, attacker_dice, defender_dice, n_att, n_def, validate_fn):
     attacker_combos = self._list_all_idx_combos(len(attacker_dice), n_att)
     defender_combos = self._list_all_idx_combos(len(defender_dice), n_def)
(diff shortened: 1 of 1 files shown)


blackshadowshade:
I acknowledge that this is random_ai only so I'm not going to block these changes. However, the code itself doesn't completely make sense to me, I'll add comments to let you know where.

tools/api-client/python/lib/random_ai.py
```diff
@@ -1405,6 +1414,24 @@ def _should_allow_fire_cancellation(self):
...
+    funcPrefix = '_game_action_start_turn_find_attack_'
+    return [ x[len(funcPrefix):] for x in sorted(dir(self)) if x.startswith(funcPrefix) ]
+
+  # Given the list of valid attack types, choose one
+  # With low probability, choose an invalid attack type, and expect it to fail.
+  def _choose_attack_type(self, validAttackTypes):
+    invalidAttackTypes = self._all_known_attack_types()
+    if random.random() < 0.01:
+      invalidAttackType = self._random_array_element(invalidAttackTypes)
+      if invalidAttackType not in validAttackTypes:
```
blackshadowshade, line 1427: So, I get that for some low probability (less than 1%), we'll get an invalid attack type here. However, I think that the code (or perhaps the variable names) are somewhat inaccurate.

invalidAttackTypes never actually ends up holding an array of invalid attack types. Nor is invalidAttackType guaranteed to hold an invalid attack type here.

If I were to rewrite this code to match the variable names, I'd expect to have to do a set subtraction of valid attack types from all attack types.

cgolubi1: Ah, you're right, i just need to change the variable name to `allAttackTypes` to make it match the logic (or do something like your suggestion, but that's probably more elegance than i need for this).  Good catch.

cgolubi1:
> I acknowledge that this is random_ai only so I'm not going to block these changes. However, the code itself doesn't completely make sense to me, I'll add comments to let you know where.

New revision addresses this comment lazily, by making the variable names match the behavior and adding a comment to describe the behavior.

blackshadowshade:
Yeah, that's better, thanks. Accepting now.

PR #2995, opened 2024-10-14
Support replay for failed submitTurn entries

cgolubi1:
Sorry for the bother - i was in too much of a hurry to merge #2992 and didn't test it for whether the games it created could be replayed.  Rookie mistake, and i even have a flag to `replay_loop` to support that testing, i was just in a hurry.

Anyway, this PR fixes that one so when `submitTurn` fails expectedly, PHP-based replay testing correctly expects that failure as well.  I've tested it using `replay_loop -l`, where i've played 1200 novel games, replayed them all, and also replayed 1200 games from the archive, and those all passed.

### DIFF
--- a/tools/api-client/python/lib/random_ai.py
+++ b/tools/api-client/python/lib/random_ai.py
@@ -416,17 +416,30 @@ class PHPBMClientOutputWriter():
       entry['attacker_indices'], entry['defender_indices'],
       entry['all_attackers'], entry['all_defenders'],
       entry['attacker'], entry['defender'])
-    php_turbo_vals = 'array(%s)' % ', '.join(['%s => %s' % (k, v) for k, v in sorted(entry['turbo_vals'].items())])
-    self.f.write("""
-        $this->verify_api_submitTurn(
-            %s,
-            '%s',
-            $retval, %s,
-            $gameId, %s, '%s', %d, %d, '', %s);
+    if entry['retval'].status == 'ok':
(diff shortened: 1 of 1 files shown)


tools/api-client/python/lib/random_ai.py
```diff
@@ -416,17 +416,30 @@ def _write_entry_type_submitTurn(self, entry):
...
+        randstr, self._php_get_message(entry['retval']), php_attack_str,
+        entry['roundNumber'], entry['attackType'], entry['attacker'], entry['defender'],
+        php_turbo_vals))
+    elif entry['retval'].status == 'failed':
+      self.f.write("""
+          $this->verify_api_submitTurn_failure(
+              %s,
+              '%s',
+              $retval, %s,
+              $gameId, %s, '%s', %d, %d, '');
```
cgolubi1, line 437: I'm not offhand sure *why* `verify_api_submitTurn()` takes a `turboVals` arg and `verify_api_submitTurn_failure()` doesn't, but that is the state of play in responderTest, and this code's job is to match what responderTest expects, so that's what i did.

blackshadowshade: I've taken a look and verify_api_submitTurn*() are only used in the responder tests, so I think this doesn't matter much in the greater scheme of things.

blackshadowshade:
This change is to random_ai only.

I see a CircleCI success here: https://app.circleci.com/pipelines/github/cgolubi1/buttonmen/264/workflows/ff54109b-ad05-4968-82b6-25bdc9da8b3a/jobs/532

Accepting this pull request now.

PR #2996, opened 2024-10-18
Add some more test cases to RandomAI

cgolubi1:
* sometimes choose Default attacks, using dice which have already been chosen as valid for some attack type
* occasionally choose Surrender attacks
* occasionally intentionally submit an invalid attacker/defender combo
* after any submitTurn call fails, check that the API-reported game state is unchanged

These are additions i've had in mind to make for awhile, and the relevant code was top-of-mind because of the "invalid pass" fix.  I've run about 2000 novel and replayed games, and nothing has turned up, so i think it's okay to put this up for PR.


blackshadowshade:
Changes are only to random_ai, looks okay to me. CircleCI passes. Accepting now.

PR #3000, opened 2024-10-27
Explicitly handle the case where a player requests a non-existent forum thread

### LINKED ISSUE #2993: internal error when someone loads a nonexistent forum thread
If you browse to e.g. https://staging.buttonweavers.com/ui/forum.html#!threadId=10000 you get the error:

```
Error from loadForumThread: Forum thread loading failed
```

and the apache log gets an internal error:

```
Wrong number of records returned for forum_thread.id = 10000, referer: https://staging.buttonweavers.com/ui/forum.html
```

This shouldn't be an internal error; we should give the player a friendly "no such thread" error message, and we shouldn't log anything to apache error log.

cgolubi1:
* Fixes #2993 

The player-visible change here is trivial --- a marginally friendlier message.  The purpose of the fix is to make it so that a player requesting a non-existent forum thread doesn't cause an error log message, which it shouldn't, because only internal errors should cause error log messages.

I don't think this needs a dev site --- i think the extra responder test stub plus the screencap i'm going to put in a comment demonstrates that it works, and it's not worth more discussion than that.  But let me know if you think more is needed!

### DIFF
--- a/src/engine/BMInterfaceForum.php
+++ b/src/engine/BMInterfaceForum.php
@@ -238,7 +238,11 @@ class BMInterfaceForum extends BMInterface {
             );
             $rows = self::$db->select_rows($query, $parameters, $columnReturnTypes);
 
-            if (count($rows) != 1) {
+            if (count($rows) == 0) {
+                $this->set_message('The requested forum thread does not exist');
+                return NULL;
+            }
+            if (count($rows) > 1) {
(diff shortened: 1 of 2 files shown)


cgolubi1:
Here's what i get on my local test site when i try to load forum thread 10000:

3000_new_forum_error_message

blackshadowshade:
Looks good to me. Merging now.

PR #3006, opened 2024-11-04
Minor fixes to docker deployment scripts

cgolubi1:
* Support parsing either SSH or HTTPS git remotes
* Fix escaping of regexps for newer python versions
* Import boto3 just-in-time because it's not needed for local containers

These changes are to support blackshadowshade's testing, and also partially cover #2973.


blackshadowshade:
CircleCI passes. Merging now.

PR #3009, opened 2024-11-04
Set file modes correctly when copying files from /buttonmen to /var/www

### LINKED ISSUE #3007: vagrant inherits /var/www permissions from git working directory
There are a couple of vagrant commands that rsync files from `/buttonmen` into `/var/www`:

```
  exec {
    "buttonmen_src_rsync":
      command => "/usr/bin/rsync -a --delete /buttonmen/src/ /var/www/",
      require => [ Package["rsync"], Package["apache2"] ];
      
    "buttonmen_uitest_rsync":
      command => "/usr/bin/rsync -a --delete /buttonmen/test/src/ui/ /var/www/test-ui/",
      require => Exec["buttonmen_src_rsync"];
  }   
```

Those commands don't set permissions, they just take whatever `/buttonmen` has, and `/buttonmen` is synced from the dev's git working directory.  So if the dev's working directory doesn't happen to have global `rx` permissions, the containerized site can't be accessed (and if the dev's working directory has write permissions, those would propagate to the site files as well).

cgolubi1:
* Fixes #3007 

I tested that `deploy_buttonmen_site -l` still works for me with this change.

Shadowshade, want to test that it solves your deployment problem before accepting?  With this change, you should be able to deploy a container site and it should be accessible via web UI, without any hand-modification needed.

### DIFF
--- a/deploy/vagrant/modules/buttonmen/manifests/init.pp
+++ b/deploy/vagrant/modules/buttonmen/manifests/init.pp
@@ -32,6 +32,11 @@ class buttonmen::server {
       content => template("buttonmen/backup_database.erb"),
       mode => 0555;
 
+    "/usr/local/bin/sync_buttonmen_web_files":
+      ensure => file,
+      content => template("buttonmen/sync_web_files.erb"),
+      mode => 0555;
+
     "/usr/local/bin/set_buttonmen_config":
       ensure => file,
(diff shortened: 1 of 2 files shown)


blackshadowshade:
I've tried this out on my setup and the site comes up correctly on http://localhost:8080/ first try. Accepting this now.

PR #3018, opened 2024-12-03
Handle what amounts to a division-by-zero error when the divisor is 'None'.

### LINKED ISSUE #3017: Python 'head-to-head' script has an unhandled exception if two buttons have never played
The `head-to-head` script in the Python api-client tools has some code meant to handle a situation where two buttons have never played, but it's currently throwing an exception when this comes up:
```
[19:57:54] irilyth@azazel:/home/irilyth/src/github/irilyth/buttonmen/tools/api-client/python
+$ ./head-to-head button Georgia Bauer
  File "./head-to-head", line 143, in <module>
    print_record(bmconn, args.type, args.protag, opponent)
  File "./head-to-head", line 116, in print_record
    winpct = (wins / games) * 100
TypeError: unsupported operand type(s) for /: 'int' and 'NoneType'

```
This could be a Python 3 change, or possibly an API change, where `games` is now coming back as `None` rather than `0`.

irilyth:
Fixes #3017.

With this change, the command that currently throws an exception instead says
```
[20:05:30] irilyth@azazel:/home/irilyth/src/github/irilyth/buttonmen/tools/api-client/python
+$ ./head-to-head button Georgia Bauer
Georgia vs Bauer: 0 - 0 (0.00%)

```
as expected and desired.

### DIFF
--- a/tools/api-client/python/head-to-head
+++ b/tools/api-client/python/head-to-head
@@ -114,7 +114,7 @@ def print_record(bmconn, type, thing1, thing2):
 
   try:
     winpct = (wins / games) * 100
-  except ZeroDivisionError:
+  except (ZeroDivisionError, TypeError):
     winpct = 0
 
   print("{0} vs {1}: {2} - {3} ({4:.2f}%)".format(thing1, thing2, wins, losses, winpct))


cgolubi1:
Sure, this seems fine.  In theory, it could mask other `TypeError` causes, but that doesn't feel worth worrying about; i'm going to go ahead and accept it.

Issue #2953, opened 2024-05-23
make replay testing more user-friendly


cgolubi1:
The goal of replay testing is to throw compute time rather than human time at finding logic bugs / breaking changes in new code, even if the tester doesn't know what kind of bug they're looking for.  That would be easier/more reliable if i sanded down some current rough edges in the replay test rig.  Specific things i have in mind:
* Flags:
    * There are things i commonly want to do, such as restrict which buttons get used in novel games, which i do by editing `replay_loop` on the replay site.  Those should be CLI flags.
    * `replay_loop` should have help/usage text describing all the CLI flags.
* Automate a mix of behavior:
    * There are behaviors that are not worthwhile to test on every loop iteration, because either they'll fail or they won't, but should be tested more than zero times.  One example is replay-testing of the novel games created by the current iteration of `replay_loop` --- that would catch problems like introduction of unmodelled randomization that are rare, but important.  Currently, those behaviors are tested only if a particular CLI flag is selected --- instead they should always be tested a small percentage of the time.
    * The mix of behaviors that `replay_loop` tests by default should be documented for quick reference, so we can easily know what's been tested for a particular branch.

cgolubi1:
Another thought to integrate into the above list: it would be great to have less manual work (ideally none) needed to turn random_ai generated tests into responder tests.

cgolubi1:
A silly bug: when a game runs to 200 rounds (e.g. Echo vs IIconfused) and is cancelled, RandomAI falls over.

I think what's happening is simply that the python process gets OOM-killed when it's trying to pull the entire game action log into memory and write the final game state.

PR #3019, opened 2025-01-04
When restarting mysqld during replay testing, retry the start a few times

cgolubi1:
Partially addresses #2953

This addresses an issue in which sometimes mysqld takes more than 2 seconds to stop, so it hasn't stopped yet when the start is attempted, so the start fails, and the replay loop falls over and has to be manually restarted.

I've verified via a replay site that this functionality works in the happy path, and that if i substitute a broken "start" command, the failure case works --- so i didn't break the ability to run test games.  I'm letting it run to see if it helps with the problem, so we could wait a week to approve it, but also it's harmless, so if you want to just go ahead and approve it, i think that's fine too.

### DIFF
--- a/tools/api-client/python/replaytest/replay_loop
+++ b/tools/api-client/python/replaytest/replay_loop
@@ -136,7 +136,15 @@ def save_state(state):
 def restart_mysqld():
   os.system('sudo killall -u mysql')
   time.sleep(2)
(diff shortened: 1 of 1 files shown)


blackshadowshade:
Looks fine to me. Merging now.

PR #2131, opened 2016-11-01
Tournaments

blackshadowshade:
This pull request is going to be a full implementation of single elimination tournaments, an initial attempt at addressing #462.

To do
- [x] Responder tests
- [x] RandomAI compatibility (code will be submitted in a separate pull request)
- [x] Clearing the current UI with testers
  - [x] Reduce number of tables to three: Closed, Open, Active
  - [x] Move open tournaments that you have joined to the bottom of the Open table
  - [x] Display tournament number (+ verb View or Join + colour?) in the leftmost column
  - [x] For open tournament pages that you are not part of, change wording to Join Tournament
  - [x] For open tournament pages that you are a part of, make it easy to change your button 
  - [x] For active tournaments, put games in chronological order 
  - [x] irilyth and devious to review whether issues raised in https://github.com/buttonmen-dev/buttonmen/pull/2131#issuecomment-1402294093 have been addressed
  - [x] Consider workflow when initially visiting a tournament page
  - [x] Fix error message when joining a tournament without selecting a button
  - [x] Check absence of open / close triangles on Tournaments page
  - [x] Add number of rounds to tournament page
  - [x] Consider how to deal with game messages generated when a tournament changes rounds or finishes
  - [x] Make tournament winner more prominent
- [x] Opening new issues regarding usability and evolution of the UI
  - [x] Have some functionality to view a list of completed tournaments (e.g., Tournament History page)
- [x] QUnit tests for the most important (possibly immutable) parts of the UI
- [x] CircleCI
- [x] Add tournament API methods to python API client (copy code in https://github.com/buttonmen-dev/buttonmen/pull/2131#issuecomment-2440494395)
- [x] Add additional responderTest helper functions to `test/src/api/responderTestFramework.php` (diff in https://github.com/buttonmen-dev/buttonmen/pull/2131#issuecomment-2440510258)
- [x] Code cleanliness issues raised by cgolubi1 in testing, which should be either addressed or responded to:
  - [x] Why is "Newgame" now a global in jshintrc?
  - [x] Rename the BMInterfaceTournament function `update_tournament()`, because that function does *not* implement the `updateTournament` API call, so it's confusing for them to have the same name.
- [x] Bugs found via cgolubi1's API testing:
  - [x] Backend allows creation of tournaments with any number of players, should only allow 4, 8, 16, 32 like the UI does
  - [x] Backend should set a message if the player tries to create a tournament with an unreasonable integer number of wins, such as 0 or 6
[...]

PR #3027, opened 2025-02-16
Refactor: set the message that game data was successfully loaded in a more portable place

cgolubi1:
In particular, this change is to set the "Loaded data for game $gameId." message only in the public method `load_api_game_data()` which is directly invoked by the API, not in the internal method `load_game()`.

Taking a step back: the context in which we're having so much trouble with `$this->message` over in #2131 is that there are two different high-level patterns of the message we want to set for the user:
* Hi, API caller, i did something for you; i'm using the API message to say in more words "this succeeded" or "this failed", and maybe to summarize some easily-attainable details, but there's nothing more to say about it.  (This is the case for both loading game data and creating a game, and lots of other things.)
* Hi, API caller, you asked me to take a step on processing a game, and that step may have had an arbitrary number of follow-on effects, and i want to tell you about all of the ones that are relevant to you.  The only way to find out what they are is to capture that information as we proceed through the state transitions.  (This is the case for submitTurn.)

This confusion comes to a head with tournaments because tournaments introduce logical actions which are taken as a result of `submitTurn`, so as far as the site is concerned, they're part of the `submitTurn` workflow.  But as far as the player who called `submitTurn` is concerned, they are not.  From that player's POV, if their turn causes the end of a game which causes the next round of the tournament which causes the creation of many more games, which causes armies to be deployed and wars to be won and lost... none of that matters, they just want to know what happened during the game as a result of their attack.

So the long-term fix is to do a bigger refactor of how we set `$this->message` which does a better job of allowing internal methods to propagate pieces of user-friendly message text upwards, and public functions to decide what message to send.

And in the immediate term, to unblock tournaments, we can take the most limited version of that basic approach for specifically the pieces of message text that are causing problems with tournaments right now.  That means:
* Public interface methods shouldn't invoke other public interface methods, because the public interface method is designed to be called directly from the API, period.
* `$this->message` should be set in the public interface method.

For the `load_game()` case, that second one is all that's needed, and that's what this PR does, as a proof-of-concept that it's easy and harmless to make that change, and i believe it will actually address the relevant problem in the PR.

For the `create_game()` case, this is showing up a slightly more complicated issue, which is that tournament game creation is calling the *public* `create_game()` method.  So Shadowshade should do a refactor that moves all the logic of `create_game()` into a protected method, so you can have an actual public method that just makes the API call and returns it with a friendly message.  That should be simple, and should get us to "public interface methods shouldn't invoke other public interface methods."

### DIFF
--- a/src/engine/BMInterface.php
+++ b/src/engine/BMInterface.php
@@ -297,6 +297,7 @@ class BMInterface {
             $data['creatorDataArray']['creatorName'] =
                 $this->get_player_name_from_id($data['creatorDataArray']['creatorId']);
 
+            $this->set_message("Loaded data for game $gameId.");
             return $data;
         }
         return NULL;
@@ -366,8 +367,6 @@ class BMInterface {
             $this->load_drawn_cards($game);
 
             $this->recreate_optRequestArrayArray($game);
-
-            $this->set_message($this->message."Loaded data for game $gameId.");
 
             return $game;
         } catch (Exception $e) {


cgolubi1:
Successful circleci run: https://app.circleci.com/pipelines/github/cgolubi1/buttonmen/307/workflows/a1cd5287-2ab7-4aa5-9d2f-f84d4e9ce0b7

blackshadowshade:
Okay, this makes a lot of sense.

I'll approve this now on the strength of the CircleCI success, and thus the passing responder tests.

PR #3031, opened 2025-03-23
Use strlen when testing validity of API string argument length

### LINKED ISSUE #3030: API erroneously accepts overlong Unicode game descriptions
As described in https://github.com/buttonmen-dev/buttonmen/pull/2131#issuecomment-2745964557 it's possible to create a broken game by filling the UI box with Unicode characters.

This shouldn't be possible, because the API should reject overlong strings in a Unicode-aware way.  But something ain't right in how it's implemented, because it is allowing the creation of games, whose descriptions throw an internal error on load:

```
Response to API call array (\n  'type' => 'loadGameData',\n  'game' => '879',\n  'logEntryLimit' => 10,\n  'automatedApiCall' => false,\n) failed json_encode(): Malformed UTF-8 characters, possibly incorrectly encoded
```

cgolubi1:
* Fixes #3030 

The new responder test checks provide evidence that the fix works.  Without the fix, a similar test shows that game creation succeeds, but the description is corrupted:

```
1) responder00Test::test_request_createGame
--- Expected
+++ Actual
@@ @@
-'αααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααα'
+'ααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααα?'

/buttonmen/test/src/api/responder00Test.php:345

FAILURES!
Tests: 2, Assertions: 74, Failures: 1.
```

I also tested manually and found that with the fix, 127 alphas succeed and 128 alphas fail cleanly.  Without the fix, i see a variety of unhelpful behavior as described in comments to #3030

### DIFF
--- a/src/api/ApiSpec.php
+++ b/src/api/ApiSpec.php
@@ -1031,7 +1031,7 @@ class ApiSpec {
      */
     protected function verify_argument_of_type_string($arg, $argtype = array()) {
         if (is_string($arg)) {
-            $length = mb_strlen($arg, mb_detect_encoding($arg));
+            $length = strlen($arg);
             if (isset($argtype['maxlength']) && $length > $argtype['maxlength']) {
                 return FALSE;
             }
--- a/test/src/api/responder00Test.php
+++ b/test/src/api/responder00Test.php
@@ -329,6 +329,33 @@ class responder00Test extends responderTestFramework {
         $this->assertEquals(array('gameId'), array_keys($retval['data']));
         $this->assertTrue(is_numeric($retval['data']['gameId']));
         $this->assertEquals("Game " . $retval['data']['gameId'] . " created successfully.", $retval['message']);
+
+
+        // Check game creation using unicode descriptions
+	// A maximum-length unicode description should succeed,
+	// and loadGameData should report the same description that was submitted
+        $valid_description = 'ααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααααα';
+        $retval = $this->verify_api_createGame(
+            array(1, 1, 1, 1, 2, 2, 2, 2),
+            'responder003', 'responder004', 'Avis', 'Avis', 3, $valid_description, NULL, 'data'
+        );
+        $this->assertEquals('ok', $retval['status'], 'Game creation should succeed');
+        $real_game_id = $retval['data']['gameId'];
+        $retval = $this->verify_api_success(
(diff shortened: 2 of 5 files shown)


blackshadowshade:
I'm fine with the fix per se. I just wonder whether we should rename 'maxlength' to 'maxsize' to make it clear that we're talking about a number of bytes instead of a number of characters.

blackshadowshade:
(I'm not going to block the pull request based on that request if you think it's a bad idea)

cgolubi1:
I guess i would say it doesn't super matter --- this is an internal function used by the API specification; it's not a player-visible part of the API.

That said, i'm happy to change the variables in ApiSpec to maxbytes and minbytes, to make that clear.

Worth flagging: this new code is definitely *not* consistent with the length validation the UI does --- and i'm not all that it was consistent before this update either.  So there are UI uses of maxlength that we probably want to revisit, so we can give UI players a friendlier error.

However, i don't think we should futz with that now; i think we should get this PR over the fence, because i'd rather have players see "Argument (description) to function createGame was invalid", than be able to create games which block their overview page from loading, which is a live bug we have right now.

cgolubi1:
Update: i looked at ApiSpec a bit, and sorry to be a coward about it, but the problem is that we're using `maxlength` two different ways: one is maximum length of a string (the case we changed here to make it correctly Unicode-friendly), and one is maximum length of an array (i.e. maximum number of elements).

So it's not just a question of copy/paste --- if i changed all the strings to `maxbytes`, i would have to carefully parse out which ones are the strings and which ones are the arrays, and that would make the PR more error-prone, at which point it would definitely need dev testing.

I guess i could change everything to `maxsize`, but i don't like `maxsize`; i don't think it's especially clearer than `maxlength`:
* Argument against "size": there's an implicit "in what unit" that we're not answering in the variable name
* Argument for "length": after all, we're using the function `strlen()` [length] to get the number of bytes of the string, so there's a semantic argument that "length" is correct

In conclusion: philosophically, i like the idea of using "bytes" for strings and "entries" or "elements" for arrays, but i don't think it's worth making this PR a lot more complicated to approve for that small benefit.  Someone is going to have to return to this topic in the future to make a player-visible improvement to the UI length measurements, and they can think about these variable names in `ApiSpec` when they do so.

blackshadowshade:
Fair enough, we'll deal with this in the future if we need to then.

PR #3037, opened 2025-05-21
When an SSH- or SCP-based command fails, the whole deployment should fail

cgolubi1:
This has been a papercut for me in standing up dev sites for awhile: sometimes a dev site fails to come up healthy, usually due to intermittent MySQL issues, and the deployment doesn't fail immediately, but just rolls through.  So i have to check for hidden failures on SSH commands, which is tedious.

This PR won't actually address the underlying flakiness problem --- i'll presumably need a new fix for that once i learn more about it --- but it'll make it easier to know when i have to retry.

I tested this by standing up a dev site, which was successful and loaded its database (i could login to it).  I've torn it down now.


blackshadowshade:
Looks good to me.

CircleCI passes: https://app.circleci.com/pipelines/github/cgolubi1/buttonmen/313/workflows/54752ca1-a2ec-4179-b617-6f3a8299afb9

Merging now.

PR #3043, opened 2025-06-07
Make the OS-default version of phantomjs work

### LINKED ISSUE #3042: circleci failing because phantomjs can't be unpacked
The "unpack phantomjs" step of CircleCI testing persistently fails.

Step details:
```
#!/bin/bash -eo pipefail
/bin/tar xjf /usr/local/src/phantomjs-1.9.7-linux-x86_64.tar.bz2 --strip-components=1
```

Failure:
```
bzip2: (stdin) is not a bzip2 file.
/bin/tar: Child returned status 2
/bin/tar: Error is not recoverable: exiting now
```

cgolubi1:
* Fixes #3042 

The problem was that the downgraded version of phantomjs we were previously installing became unavailable online.

The reason we had to downgrade in the first place was https://github.com/ariya/phantomjs/issues/14376 --- basically the phantomjs project and debian/ubuntu had a disagreement about how to support headless things, so newer phantomjs versions couldn't run on ubuntu.

However there's an easy fix, per https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=817277 - just install a fake X server and run phantomjs under it.  With that change, the Ubuntu-default version works.


cgolubi1:
Here's the successful check: https://app.circleci.com/pipelines/github/cgolubi1/buttonmen/317/workflows/5ec01694-b41e-4a2e-a057-5ad38f15c337

And here's a failed check with this code but an intentionally-broken version of `test_VerifyReset.js` - i did this to make sure the test successes were actually being checked: https://app.circleci.com/pipelines/github/cgolubi1/buttonmen/316/workflows/7fa9b8ed-165f-4666-a663-a720ca20b84b/jobs/586

deploy/vagrant/modules/javascript/templates/run-jscover-qunit.js.erb
```diff
@@ -133,9 +133,6 @@ page.open(system.args[1], function(status){
                 } catch (e) { }
                 return 10000;
             });
-            page.evaluate(function(){
-                jscoverage_report('js-coverage');
-            });
```
cgolubi1, line 138: This hasn't run successfully in a long time, so i deleted it while i was here to get rid of a spurious error.

deploy/vagrant/modules/javascript/templates/install_phantomjs.erb
```diff
@@ -1,13 +0,0 @@
...
-set -e
-set -x
-
-PHANTOMJS_VERSION_NAME="phantomjs-1.9.7-linux-x86_64"
-
-cd /tmp
-/usr/bin/wget https://bitbucket.org/ariya/phantomjs/downloads/${PHANTOMJS_VERSION_NAME}.tar.bz2 -O /tmp/${PHANTOMJS_VERSION_NAME}.tar.bz2
-sleep 10
-/bin/tar xjf ${PHANTOMJS_VERSION_NAME}.tar.bz2
-cp ${PHANTOMJS_VERSION_NAME}/bin/phantomjs /usr/local/bin/phantomjs
```
cgolubi1, line 13: This was no longer used, so i deleted it to avoid confusion around how we're installing phantomjs.

blackshadowshade:
I've taken a look at the CircleCI output for the failing and passing runs. I note that at the end of "Run QUnit tests", it says:

    ReferenceError: Can't find variable: QUnit
    ReferenceError: Can't find variable: QUnit
    Blocked a frame with origin "file://" from accessing a frame with origin "null".  The frame requesting access has a protocol of "file", the frame being accessed has a protocol of "about". Protocols must match.

Any idea what's going on there?

cgolubi1:
Not really: there's some sort of regression from an earlier version of this setup to the current one in the details of whether the `logMessage()` function can access the `QUnit` namespace.

However, they're warnings not errors, and per my testing:
* it doesn't interfere with the fundamental behavior we want (pass on pass, fail on fail): see above for data points
* if i disable the `logMessage()` function entirely, we don't get the log lines about what tests are being run and what their output is.  and we want that!

So i agree something isn't 100% ideal here, but it didn't seem like a blocker (and might not even be new).

cgolubi1:
FWIW, i *think* this is unchanged from before.  Here's the most recent successful run on the primary branch: https://app.circleci.com/pipelines/github/buttonmen-dev/buttonmen/270/workflows/332ff606-58d6-4697-a019-940ae38838f4/jobs/373

It fails out with an exception related to the jscoverage report.  I think cleaning up those jscoverage lines from the test wrapper, is what turns it from that nonfatal warning message, to the current nonfatal warning message about QUnit not being found (because either way it bails out on the first post-test error it finds).

PR #3049, opened 2025-06-28
Standardize logic in apache_setup_certbot so certbot only actually runs once per container startup/deploy

### LINKED ISSUE #3048: dev site standup can fail due to multiple certbot race condition
When `deploy_buttonmen_site` is deploying an ECS site, there are two places certbot might run:
* `startup.sh`: the script that runs at container startup
* `deploy_buttonmen_site` itself, which runs certbot after configuring DNS

If i decide it's important, i'll write a comment and get into the logic of exactly what combination of:
* container is behind a load balancer and thus DNS is already setup on container start, vs container is directly accessible and thus DNS isn't setup until after container start
* this FQDN has been used before and thus there's a cert in the EFS directory, vs it's a new FQDN and there's no cert
* container is being stood up as part of `deploy_buttonmen_site` vs as part of an ECS container replacement
leads to each of those scripts contributing exactly what to the process of running certbot

But the bottom line is:
* If certbot tries to run and discovers another certbot already running, it fails with the error `"Another instance of Certbot is already running."`, and that error causes `apache_setup_certbot` (wrapper) to fail, and thus whatever called it to fail
* If a dev site is stood up for an FQDN which has been used in the past, both of those paths will run `apache_setup_certbot`, and a race condition is possible in which the second one starts while the first one is still running and fails with that error

I can't immediately think of a reason this couldn't cause failures when deploying to staging or prod.  My guess is that i introduced this problem with #3037 because until that PR, the race condition was harmless, but now the race condition causes `deploy_buttonmen_site` to fail.

cgolubi1:
* Also wait for mysqld to startup before running database-related deploy commands

* Fixes #3048

### DIFF
--- a/deploy/docker/deploy_buttonmen_site
+++ b/deploy/docker/deploy_buttonmen_site
@@ -592,8 +592,19 @@ def run_scp_command(local_path, remote_path, public_ipv4):
     raise ValueError(f"SCP-based command failed with exit code: {retcode}")
 
 def configure_container_setup_certbot(public_ipv4):
-  cmdargs = 'sudo /usr/local/bin/apache_setup_certbot'
+  cmdargs = 'sudo /usr/local/bin/apache_setup_certbot new_cert'
   run_ssh_command(cmdargs, public_ipv4)
+
+def configure_container_wait_for_mysqld(public_ipv4):
+  cmdargs = f"sudo /etc/init.d/mysql status"
+  while True:
+    try:
+      run_ssh_command(cmdargs, public_ipv4)
+      print("mysqld is running")
+      return
+    except ValueError:
+      print("...waiting for mysqld")
+      time.sleep(1)
 
 def configure_container_load_database(git_info, public_ipv4):
   load_database_path = git_info['config'].get('load_database_path', None)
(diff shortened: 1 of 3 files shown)


cgolubi1:
Testing:
* Without this change, verified the bug: the first time i setup a dev site using this branch, it succeeded.  The second time, it failed with the "Certbot is already running" error.
* With this change, verified that standing up a dev site which has been stood up before succeeds, and also deleting the letsencrypt directory and rebuilding the dev site succeeds.  In either case, the site is accessible via SSL.

blackshadowshade:
I wonder whether it would be worthwhile having a maximum wait timeout, just so it doesn't hang forever.

cgolubi1:
So, i think it doesn't matter, because it's a script meant to be run manually.  So i only ever run it when i'm trying to deploy something, and if it loops forever saying "still waiting for mysqld to be up...", i'll notice and check it out.

I agree nothing within the container should retry forever, and that nothing should run forever silently, but this isn't that.

blackshadowshade:
All right then, happy to accept these changes as an improvement on existing logic.

PR #3056, opened 2025-07-18
Push changes from master to staging on 2025-07-18

cgolubi1:
This update brings these approved pulls to the staging branch:

* #2131

Database updates:
```
UPDATES=deploy/database/updates/00462_tournament_01.sql
SCRIPTS=
```

Once this pull is merged, i will load the changes on staging.buttonweavers.com as part of #2786.

PR #3057, opened 2025-07-18
Fixed tournament overview page for a tournament that you have cancelled and did not join

blackshadowshade:
Addresses bug mentioned in #3056 where the tournament overview page shows Dismiss instead of Unfollow for a tournament that you have created and cancelled but not joined---different to the tournament page which correctly showed Unfollow.

I have tested this locally with tournaments that I have created and cancelled immediately after (the bugged case), and tournaments that I have created, joined, and then cancelled.

### DIFF
--- a/src/ui/js/TournamentOverview.js
+++ b/src/ui/js/TournamentOverview.js
@@ -318,15 +318,36 @@ TournamentOverview.addDismissCol = function(tournamentRow, tournamentInfo) {
   dismissTd.css('white-space', 'nowrap');
   tournamentRow.append(dismissTd);
 
-  var dismissLink = $('<a>', {
-    'text': 'Dismiss',
-    'href': '#',
-    'data-tournamentId': tournamentInfo.tournamentId,
-  });
-  dismissLink.click(TournamentOverview.formDismissTournament);
-  dismissTd.append('[')
-           .append(dismissLink)
-           .append(']');
+  var apiInfo = $.grep(
+    Api.tournaments.tournaments,
+    function(e){ return e.tournamentId === tournamentInfo.tournamentId; }
+  );
+  var hasJoinedTournament = false;
+  if (apiInfo.length > 0) {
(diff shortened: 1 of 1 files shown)


cgolubi1:
Hmm.  I appreciate you putting this up, and i'll make a dev site for it now, but given that it changes the logic at the UI level and there's no unit tests at the UI level, i don't want to rush it into this weekend's release.

So i think we should try to test it for next week.

cgolubi1:
Indeed, putting up the dev site immediately shows a problem --- when there are no tournaments, the page errors out with a javascript error:

  https://3056-dismiss-tournament-bug.blackshadowshade.dev.buttonweavers.com/ui/tournaments.html

I see that page as blank, and the error is:

```
Uncaught TypeError: can't access property "length", Api.tournaments.tournaments is undefined
```

blackshadowshade:
Okay, this one is theoretically trivial to fix (i.e., test for Api.tournaments.tournaments), but I'm not sure why it's happening.

I've looked at the page http://3056-dismiss-tournament-bug.blackshadowshade.dev.buttonweavers.com/ui/tournaments.html, and I see the error that you have observed. When I type into the console

    Api.tournaments

I get

    {load_status: "failed"}

I've recreated a clean database on my local and created four users. I've gone to the tournament page, but I'm seeing no errors. Api.tournaments.tournaments exists there.

Can you confirm that the database being used has the tournament database table changes applied?

cgolubi1:
Ohh, of course you're right.  I always download a new staging database after deploying to production, which means that during the inter-deployment period, the staging database i use for dev sites is stale.

I'll fix that later today, good catch, and sorry about that!

dwvanstone:
I have verified that the bug is fixed. I couldn't find anything unusual in the rest of my tests.

cgolubi1:
Okay.  I am cranky that we don't have qunit tests here, because it would be easier not to have to rely on manual testing, but given that dwvanstone has done the manual testing, let's go ahead and merge this.

Issue #2915, opened 2023-06-21
Verification and password reset emails not being received


blackshadowshade:
In the last month, we've had a number of users report that they have not received verification and password reset emails.

We need to work out what's going on here and fix it

cgolubi1:
Okay!  I configured SES for buttonweavers.com, stood up a dev site using the new configuration, created a new user, and verified that i received both the new site "hello" e-mail and the new user "verification" email.  So that seems good to go, and i'm going to go ahead and redeploy staging, and tomorrow production, using the new mail configuration.

cgolubi1:
Looks like `deploy_buttonmen_site` doesn't successfully make and deploy a new ECR image version when only the configuration has changed.  Sounds nice to fix someday, but the upshot is, since this isn't time-critical, i'll just wait until the next time i'm doing a deployment for some other reason, e.g. next week with the new year, and the e-mail configuration update will follow too.

cgolubi1:
Okay!  Deployed to staging along with another change, and it looks good.  The "new site" email went via SES, and i created a new account using a gmail address, and i received the "new user" email, it went via SES, and there's a header that says gmail thinks DKIM passes because SES is an authorized sender for our domain.

cgolubi1:
I'm going to claim this is fixed, though i'll also try to keep an eye on things for the next while and make sure e-mail is flowing.  If we see the problem recurring, we can reopen it.

blackshadowshade:
Excellent!

PR #3063, opened 2025-07-28
Updated advice if verification email never arrives

blackshadowshade:
This is a temporary band-aid for #2915.


cgolubi1:
This looks fine to me; i tried it on a dev site briefly just to make sure.  Merging.

PR #3064, opened 2025-07-30
Add replay scripts for tournaments

cgolubi1:
This code should only affect replay testing.  It does:
* Removes a `fulltest_deps` flag which was erroneously left on a test in the previous tournaments PR, which is currently (mildly) breaking replay test of the main branch
* Modifies `random_ai.py` to make replay testing of games involving players other than responder003 and responder004 possible, and to support iterating an in-progress game by a single move, whether or not that move completes the game
* Creates `random_tournament.py` and `test_log_tournament`, which support automated creation and play of random tournaments, including both validation that tournament API calls behave as we've discussed and as documented in the state machine diagram, and validation that tournaments are creatable and tournament games are playable.
* Tournament replay is not currently supported, nor is translation of tournament state pck files into php (which replay would require)
* Adds flags to `replay_loop` to support infinite play of tournaments

I tested this code by:
* Running `replay_loop -t` a lot, and verifying that it doesn't break with any logic errors, and that many tournaments can be played before hitting a DB connection error (i hit these occasionally during routine replay testing, and at some point i'm going to add retries, but not right now.  Anyway, i can play over 1000 games without hitting a connection error much of the time, so that's not so bad)
* Using normal `replay_loop` to verify that games recorded on the primary branch can be replayed using this script, so there should be no regression in normal replay testing

### DIFF
--- a/test/src/api/responderTournamentTest.php
+++ b/test/src/api/responderTournamentTest.php
@@ -10,7 +10,6 @@ class responderTournamentTest extends responderTestFramework {
 
     /**
      * @depends responder00Test::test_request_savePlayerInfo
-     * @group fulltest_deps
      *
      * This test invokes several tournament API methods
      */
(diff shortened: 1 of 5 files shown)


blackshadowshade:
Files only affect replay testing, I'm happy to accept these as-is.

PR #3075, opened 2026-01-04
Support site install on Ubuntu 24.04 + PHP 8.3

### LINKED ISSUE #2936: Upgrade PHP to 8.3
I note that we're going to need to upgrade our infrastructure relatively soon (linking #2784 and #2928), so I'm opening this issue for the PHP part of that.

On #2784, I noted the steps we would need to take to migrate from 7.2-ish to 8.0.

cgolubi1:
Fixes #2936 (when finished).

-------

Changes to get the install working:
* Upgrade base OS version (default PHP follows along)
* Remove MySQL version overrides since MySQL 8.0 is already the default for Ubuntu 24.04
* Update all PHP FS paths to expect 8.3 instead of 7.0
* Fix mandatory puppet syntax changes for puppet 8 (class names can't have dashes, file mode variables are strings, facter variables come from a dict)
* Update some python utilities to new default version python3.12, both in the base install and in CircleCI
* Stop trying to instantiate cloudwatch config (which doesn't work right now for us anyway)
* Work around rsyslog no longer supporting SysVInit syntax - just start the binary on container startup
* Fix a small hook bug in `perform_end_of_turn_die_actions()` - previous versions of PHP would let you | a bool with an array and get something truthy; that doesn't fly in PHP 8.3

Changes to get PHPUnit working:
* Apply puppet version fixes to circleci puppet
* Don't report useless tests, since there are a lot of noop tests and it'll be annoying
* Fix namespace of PHPUnit to match how PHPUnit 9 does it, and also apply required type specification for setUp() and tearDown() functions
* Comment out a couple of die tests which use readAttribute() - that's not allowed any more, so we'll have to figure out what to do instead
* PHPUnit 9 doesn't reset global variables between tests any more; fix a couple of tests which use global variables so the right thing happens (TODO: there may be more of these)

Changes to get UI tests working:
* Install phantomjs via npm on CircleCI nodes
* Work around incompatibility between phantomjs and newer OpenSSL by disabling OpenSSL in the phantomjs invocation

Todo list:
* [x] `/etc/cron.daily/logrotate` not working
* [x] get the SAS e-mail configuration installed and make sure it works

-------
Shadowshade: it'd be great if you could take a look at this:
* Other than the commented-out tests because we no longer have access to `readAttribute()`, do you have any concerns about the approach?
* What are your thoughts about the `readAttribute()` test problem?  Do you think it's a blocker?  I can try to think about other ways to test the internal state you're trying to test, but i kind of think this is something we're going to have to improve over time.  I'm personally not that worried about it, because it's only two tests, and my experience has been that testing the externally-visible state of games works pretty well.  But i'm sure you use the tests for various things before submitting PRs in the first place.
* FYI, without the change i made to `perform_end_of_turn_die_actions()`, some of the responder tests were failing with `TypeError: Unsupported operand types: bool | array` in a case in which a button had `o(V)?` at the end of a round.  I think it's pretty clear what the issue was (as noted above in the commit message), and i don't have concerns about the fix, but LMK if you do.

Obviously i'll need to put up a dev site and run some replay tests and stuff, but i wanted to start out with you taking a look at the change now that CircleCI is passing.

One note in passing: it seems like newer versions of PHP have better support for type hints (that's what i had to add to the `setUp()` and `tearDown()` functions).  I actually think making use of type hints in our PHP code could do us a lot of good in terms of avoiding surprising problems caused by functions that turn out to be more polymorphic than we thought they were.  If you agree, let's cut an issue to start adding them down the line as time permits.


cgolubi1:
Passing CircleCI: https://app.circleci.com/pipelines/github/cgolubi1/buttonmen/347/workflows/0f402533-9668-42dc-9381-482e8d9fa3a7

deploy/circleci/audit_js_unit_test_coverage.py
```diff
@@ -31,8 +31,8 @@ def find_test_files():
   return testpairs
 
 def check_test_file_functions(modname, srcpath, testpath):
-  function_name_re = re.compile('^(%s\.\S+) = function\(' % modname)
-  test_name_re = re.compile('^(test|asyncTest)\("test_([^"]+)"')
+  function_name_re = re.compile('^(%s.[^ \t\n\r\f\v]+) = function[(]' % modname)
+  test_name_re = re.compile('^(test|asyncTest)[(]"test_([^"]+)"')
```
cgolubi1: Drive-by comment: i could not tell you why python 3.12 has decided it's so much cleaner to use `[^ \t\n\r\f\v]` than `\S` (and `[(]` than `\(`).  This seems like a mistake someone is going to fix at some point, but the SyntaxWarnings are pretty distracting, and we don't have to touch this code all that often, so i just decided to go with what python 3.12 told me to do.

blackshadowshade:
This pull request looks good to me. I'm fine with having those tests commented out for a bit, so that we can resolve the issue in a subsequent pull request.

cgolubi1:
I stood up a dev site at https://2936-ubuntu-2404.cgolubi1.dev.buttonweavers.com/ for people to take a look at whether everything looks normal - i'll put out a call for testing at some point.

cgolubi1:
Reminding myself: based on the diff, the only real reasons i marked this `incomplete` were:
* i haven't done replay testing yet and that's very necessary because of the PHP version change (and of course we want dev testing too)
* i left a few FIXMEs about things i did untidily and want to give another look
* i wanted to check whether James thought commenting out the `readAttribute()` tests was a blocker (per above, he says no)

AFAICT there's nothing specifically missing.

cgolubi1:
Starting up replay testing (reminder to self: since we have random tournament play as an option, i should definitely run that too, since it'll exercise a different subset of code).

cgolubi1:
Replay testing has played 13000 new and replayed games without incident, so moving on to random tournament testing.

cgolubi1:
Ah hah, an invalid join action taken while a tournament is in progress throws an exception:

```
PLAY: game_to_advance=8, action=play, is_valid=True
PLAY: username=responder028, api_method=updateTournament, action=join, buttons=None, is_valid=False
could not parse return: <head/><br/>
<b>Fatal error</b>:  Uncaught TypeError: count(): Argument #1 ($value) must be of type Countable|array, null given in /var/www/engine/BMInterfaceTournament.php:665
Stack trace:
#0 /var/www/engine/BMInterfaceTournament.php(638): BMInterfaceTournament-&gt;join_tournament()
#1 /var/www/api/ApiResponder.php(849): BMInterfaceTournament-&gt;act_on_tournament()
#2 /var/www/api/ApiResponder.php(1131): ApiResponder-&gt;get_interface_response_updateTournament()
#3 /var/www/api/responder.php(23): ApiResponder-&gt;process_request()
#4 {main}
  thrown in <b>/var/www/engine/BMInterfaceTournament.php</b> on line <b>665</b><br/>

  File "./test_log_tournament", line 30, in <module>
    c.log_test_tournament(i)
  File "lib/random_tournament.py", line 484, in log_test_tournament
    self.handle_tournament_state_play()
  File "lib/random_tournament.py", line 435, in handle_tournament_state_play
    self.call_tournament_api_method(player_client, api_method, action, buttons, is_valid)
  File "lib/random_tournament.py", line 378, in call_tournament_api_method
    self.bug("%s unexpectedly gave invalid response: %s" % (call_spec, "NULL"))
  File "lib/random_tournament.py", line 82, in bug
    raise ValueError, message
ValueError: update_tournament(3, join, None) called by responder028 unexpectedly gave invalid response: NULL
```

cgolubi1:
Seems like `$buttonNameArray` should default to `array()` rather than to `NULL`.  Trying that change.

PR #3076, opened 2026-01-05
Populate SMTP SASL AUTH options for postfix

cgolubi1:
* Allow us to configure optional SASL AUTH credentials as part of deploy_buttonmen_site
* If a relayhost is defined in the config file, postfix is configured to relay all mail to that host
* If relay credentials are defined in the config file, postfix is configured to load those credentials and use them when relaying mail via the relayhost
* If relayhost and relay credentials are not configured, nothing changes
* Stop configuring postfix at all for CircleCI hosts --- we don't need it there

-------
This PR is on the path to resolving #2915 --- here's the deal:
* I think the correct solution for that issue is to send buttonmen e-mail via SES (AWS Simple Email Service), and also apply some other configuration to the buttonweavers.com domain to make it a better e-mail citizen.  In particular, i've had good luck with SES + DKIM + DMARC making gmail willing to accept e-mail on the first try, on other domains i run.
* Ideally, we should make all of those changes at once --- specifically, we should make changes at the buttonweavers.com level to expect mail to be sent via SES, and start sending mail via SES, at approximately the same time.
* To that end, what this PR does is add the ability to configure all of the pieces that would send authenticated e-mail to SES from a buttonweavers site.  But if the configuration isn't set, nothing will change.
* That way i can deploy this normally and have the code in place turned off, and when i'm ready to pull the lever, i can do a one-off redeploy of staging and production to add SES sending as a config change rather than a code change.

Here's how i tested this:
* deployed to a dev site with the new configuration enabled.  Verified that (A) the "new site is up" e-mail, and (B) the e-mail verification for a new user, attempted to be sent to `email-smtp.us-east-1.amazonaws.com`, and were rejected because the e-mail domains involved in the message were not yet verified parts of my SES account.  I think that means the plumbing works --- the e-mail is talking to the right server and being authenticated correctly with the creds i've setup --- the server-side just isn't ready yet because i haven't migrated buttonweavers.com, which is expected
* deployed to a dev site with the new configuration disabled.  Verified that the "new site is up" e-mail and the e-mail verification for a new user were successfully sent to me in the usual way.  This shows the code is correctly a noop when the configuration isn't in place

I believe this PR is safe to accept.

### DIFF
--- a/deploy/circleci/manifests/init.pp
+++ b/deploy/circleci/manifests/init.pp
@@ -14,7 +14,6 @@ node default {
   # Generic node configuration
   include "apt::client"
   include "ntp::client"
-  include "postfix::base"
   include "fqdn::base"
 
   # Node configuration needed for the circleci buttonmen server
--- a/deploy/docker/deploy_buttonmen_site
+++ b/deploy/docker/deploy_buttonmen_site
@@ -186,6 +186,16 @@ def get_remote_database_password(git_info):
     return git_info['config']['remote_database_admin_pw']
   return None
 
+def get_email_relay_host(git_info):
+  if git_info['config'].get('email_relay_host', None):
+    return git_info['config']['email_relay_host']
+  return 'NORELAY'
+
+def get_email_relay_sasl_creds(git_info):
+  if git_info['config'].get('email_relay_sasl_creds', None):
+    return git_info['config']['email_relay_sasl_creds']
+  return 'NOCREDS'
+
 def find_docker_image_with_tag(tag):
   return get_subprocess_output(['docker', 'images', tag, '--format', '{{.ID}}']).strip()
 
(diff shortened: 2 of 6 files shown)


blackshadowshade:
I'm just going to put here that I have zero expertise in this area. I'm happy to accept it, but I'd like to leave a couple of days for others who might have expertise in this area to take a look.

Anyone know anything about this stuff that could give some feedback? @irilyth? @dwvanstone?

dwvanstone:
> Anyone know anything about this stuff that could give some feedback? @irilyth? @dwvanstone?

I'm afraid this isn't my area of expertise -- I'm happy to test but I can't add to this conversation.

blackshadowshade:
I've accepted this, let's give it a go. If anything goes wrong, it's not too difficult to back out.

PR #3080, opened 2026-01-31
Update current year to 2026

cgolubi1:
It's still January, we're not late!

Obligatory disclaimers:
* It is no longer January for James
* That is not a real rule


blackshadowshade:
Looks good to me, you beat me to it!

